PR #1, opened 2015-08-14
Add sample for using cloud storage api.

jerjou:


package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "nodejs-docs-samples",
```
jonparrott: can you mark this as a private package to prevent accidental publication?

cloud-storage/authSample.js
```diff
@@ -0,0 +1,54 @@
...
+    }, callback);
+  });
+}
+// [END list_buckets]
+
+module.exports = {
+  listBuckets: listBuckets
+};
+
+/**
```
jonparrott: What is this doc string here for?

jerjou: Turns out jsdoc has this odd syntax for specifying callbacks where you declare the callback type separately from the function doc. See here: http://usejsdoc.org/tags-param.html#callback-functions

jonparrott: I see, can you put this above the function instead of at the end of the
file?

jerjou: How were you thinking? In a separate comment, above the function's docstring? Or between the function's docstring and the function?

I feel like it's awkward wherever it is, but at least if it's at the end (like in some of the jsdoc examples) it's out of the way..

jonparrott: Heh. Out of the way is why I don't like it. Above the function as in the jsdoc link you sent over is fine.

jerjou:
Updated. PTAL

jonparrott:
LGTM, merging.

PR #14, opened 2015-10-11
Add sample for using memcached for connect session state

JustinBeckwith:


JustinBeckwith:
@jmdobry do you think we should add this as a stand alone sample, or integrate it into the express one?

PR #16, opened 2015-10-14
add datastore v1beta3 snippets

stephenplusplus:
// @pcostell

#### To Dos

- [x] Entities
- [x] Queries
- [x] Indexes
- [x] Transactions
- [x] Metadata

This is some snippet code in preparation of the datastore v1beta3 release. It's a work in progress, so I wanted to get reviews and guidance on the structure to follow as early as possible.

Some questions:

- Do we have freedom to use `taskKey` (convention) vs `task_key` (snippet spreadsheet) - the linter caught this as an error, and I wasn't sure if either is a hard requirement.
- gcloud-node doesn't have an Entity type; data that goes or comes from an entity are just plain JavaScript objects, without special capabilities
- I assume this JavaScript snippet file is meant to be runnable-- as in, instead of dumping out a bunch of vars and function calls, I should box them inside of functions that can be called individually?

Any insight into how the docs are generated from this file, requirements or tips for conventions to follow would be helpful to make sure I get this right. Thanks!

### DIFF
--- a/datastore/concepts.js
+++ b/datastore/concepts.js
@@ -1,0 +1,1311 @@
+// Copyright 2015, Google, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+var asyncUtil = require('async');
+var gcloud = require('gcloud');
+
+module.exports = {
+  Entity: Entity,
+  Index: Index,
+  Metadata: Metadata,
+  Query: Query,
+  Transaction: Transaction
+};
+
+// This mock is used in the documentation snippets.
+var datastore = {
+  delete: function() {},
+  get: function() {},
+  insert: function() {},
+  key: function() {},
+  update: function() {},
+  upsert: function() {},
+  runQuery: function() {},
+  save: function() {}
+};
+
+function Entity(projectId) {
+  this.datastore = gcloud.datastore({
+    projectId: projectId
+  });
+
+  // To create the keys, we have to use this instance of Datastore.
+  datastore.key = this.datastore.key;
+
+  this.incompleteKey = this.getIncompleteKey();
+  this.namedKey = this.getNamedKey();
+  this.keyWithParent = this.getKeyWithParent();
+  this.keyWithMultiLevelParent = this.getKeyWithMultiLevelParent();
+}
+
+Entity.prototype.getIncompleteKey = function() {
+  // [START incomplete_key]
+  var taskKey = datastore.key('Task');
+  // [END incomplete_key]
+
+  return taskKey;
+};
+
+Entity.prototype.getNamedKey = function() {
+  // [START named_key]
+  var taskKey = datastore.key([
+    'Task',
+    'sampletask'
+  ]);
+  // [END named_key]
+
+  return taskKey;
+};
+
+Entity.prototype.getKeyWithParent = function() {
+  // [START key_with_parent]
+  var taskKey = datastore.key([
+    'TaskList',
+    'default',
+    'Task',
+    'sampleTask'
+  ]);
+  // [END key_with_parent]
+
+  return taskKey;
+};
+
+Entity.prototype.getKeyWithMultiLevelParent = function() {
+  // [START key_with_multilevel_parent]
+  var taskKey = datastore.key([
+    'User',
+    'alice',
+    'TaskList',
+    'default',
+    'Task',
+    'sampleTask'
+  ]);
+  // [END key_with_multilevel_parent]
+
+  return taskKey;
+};
+
+Entity.prototype.getTask = function() {
+  // [START basic_entity]
+  var task = {
+    type: 'Personal',
+    done: false,
+    priority: 4,
+    description: 'Learn Cloud Datastore'
+  };
+  // [END basic_entity]
+
+  return task;
+};
+
+Entity.prototype.testIncompleteKey = function(callback) {
+  this.datastore.save({
+    key: this.incompleteKey,
+    data: {}
+  }, callback);
+};
+
+Entity.prototype.testNamedKey = function(callback) {
+  this.datastore.save({
+    key: this.namedKey,
+    data: {}
+  }, callback);
+};
+
+Entity.prototype.testKeyWithParent = function(callback) {
+  this.datastore.save({
+    key: this.keyWithParent,
+    data: {}
+  }, callback);
+};
+
+Entity.prototype.testKeyWithMultiLevelParent = function(callback) {
+  this.datastore.save({
+    key: this.keyWithMultiLevelParent,
+    data: {}
+  }, callback);
+};
+
+Entity.prototype.testEntityWithParent = function(callback) {
+  var taskKey = this.keyWithParent;
+
+  // [START entity_with_parent]
+  var task = {
+    key: taskKey,
+    data: {
+      type: 'Personal',
+      done: false,
+      priority: 4,
+      description: 'Learn Cloud Datastore'
+    }
+  };
+  // [END entity_with_parent]
+
+  this.datastore.save(task, callback);
+};
+
+Entity.prototype.testProperties = function(callback) {
+  // jshint camelcase:false
+  // [START properties]
+  var task = {
+    type: 'Personal',
+    created: new Date(),
+    done: false,
+    priority: 4,
+    percent_complete: 10.0,
+    description: 'Learn Cloud Datastore'
+  };
+  // [END properties]
+
+  this.datastore.save({
+    key: this.incompleteKey,
+    data: task
+  }, callback);
+};
+
+Entity.prototype.testArrayValue = function(callback) {
+  // [START array_value]
+  var task = {
+    tags: [
+      'fun',
+      'programming'
+    ],
+    collaborators: [
+      'alice',
+      'bob'
+    ]
+  };
+  // [END array_value]
+
+  this.datastore.save({
+    key: this.incompleteKey,
+    data: task
+  }, callback);
+};
+
+Entity.prototype.testBasicEntity = function(callback) {
+  this.datastore.save({
+    key: this.getIncompleteKey(),
+    data: this.getTask()
+  }, callback);
+};
+
+Entity.prototype.testUpsert = function(callback) {
+  var taskKey = this.getIncompleteKey();
+  var task = this.getTask();
+
+  // [START upsert]
+  datastore.upsert({
+    key: taskKey,
+    data: task
+  }, function(err) {
+    if (!err) {
+      // Task inserted successfully.
+    }
+  });
+  // [END upsert]
+
+  this.datastore.upsert({
+    key: taskKey,
+    data: task
+  }, callback);
+};
+
+Entity.prototype.testInsert = function(callback) {
+  var taskKey = this.getIncompleteKey();
+  var task = this.getTask();
+
+  // [START insert]
+  datastore.insert({
+    key: taskKey,
+    data: task
+  }, function(err) {
+    if (!err) {
+      // Task inserted successfully.
+    }
+  });
+  // [END insert]
+
+  this.datastore.insert({
+    key: taskKey,
+    data: task
+  }, callback);
+};
+
+Entity.prototype.testLookup = function(callback) {
+  var self = this;
+  var taskKey = this.getIncompleteKey();
+
+  // jshint unused:false
+  // [START lookup]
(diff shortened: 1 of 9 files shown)


pcostell:
* taskKey is perfectly fine in this context, whatever is language idiomatic.
* For the case of snippets like `basic_entity`, just show creating a javascript object with the appropriate fields.
* Yes, all of our snippets are runnable (and do basic smoke tests to make sure the snippet is correct).

Snippets are handled using different regions. So for the `basic_entity` region this could look in code something like:

    function testBasicEntity() {
      // [START basic_entity]
      var task = { .... };
      // [END basic_entity]
      datastore.put(task); // Just a smoke test to make sure the test doesn't fail.
    }

One thing to note about the snippet is that is removes the appropriate whitespace, so don't have to worry about the snippets being at a top level.

In general, if an operation is not possible either provide a comment with a workaround or a comment that says it's not possible. For example some of our clients don't expose the insert command, so instead we show a transaction that does a get, verifies that it fails, then performs a put.

datastore/concepts.js
```diff
@@ -0,0 +1,141 @@
...
+    type: 'Personal',
+    done: false,
+    priority: 4,
+    description: 'Learn Cloud Datastore'
+  }
+};
+// [END Consistency:entity_with_parent]
+
+// [START Entities:properties]
+var taskData = {
```
pcostell: Maybe just call these task.

datastore/concepts.js
```diff
@@ -0,0 +1,141 @@
+'use strict';
+
+var gcloud = require('gcloud')({
+  projectId: process.env.TEST_PROJECT_ID
+});
+
+var datastore = gcloud.datastore();
+
+// [START Entities:incomplete_key]
```
pcostell: The snippets have to match exactly, so this should be // [START incomplete_key].

stephenplusplus:
Thanks for the review! I pushed some updates, let me know if I'm headed in the right direction.

package.json
```diff
@@ -11,6 +11,7 @@
   "license": "Apache 2",
   "private": true,
   "dependencies": {
+    "gcloud": "stephenplusplus/gcloud-node#spp--datastore-v1beta3",
```
stephenplusplus: (temporary until release: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897)

datastore/entity.js
```diff
@@ -0,0 +1,414 @@
...
+};
+
+Entity.prototype.testProperties = function(callback) {
+  // jshint camelcase:false
+  // [START properties]
+  var task = {
+    type: 'Personal',
+    done: false,
+    priority: 4,
+    percent_complete: 10.0,
```
pcostell: I think it's be ok here to do things in a language idomatic way and call this percentComplete.

jonparrott: +1, other node.js + datastore samples use camelCase.

pcostell:
So far the entity stuff LGTM

datastore/entity.js
stephenplusplus: :+1:

stephenplusplus: Oh, this will change the indexes though, which are currently:

```yaml
indexes:
- kind: Task
  properties:
  - name: done
  - name: priority
    direction: desc
- kind: Task
  properties:
  - name: priority
  - name: percent_complete
- kind: Task
  properties:
  - name: type
  - name: priority
- kind: Task
  properties:
  - name: priority
  - name: created
- kind: Task
  properties:
  - name: done
  - name: created
- kind: Task
  properties:
  - name: type
  - name: priority
    direction: desc
- kind: Task
  properties:
  - name: type
  - name: created
```

I was planning to include that file in this repo, but wherever it goes the `percent_complete` will have to be changed to match if we camelcase it.

stephenplusplus:
The query files are up, please let me know if things look good.

datastore/index.js
```diff
@@ -0,0 +1,60 @@
...
+      collaborators: [
+        'alice',
+        'bob',
+        'charlie'
+      ],
+      created: new Date()
+    }
+  };
+
+  datastore.insert(task, function(err) {
```
pcostell: No need to insert on this snippet.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+
+  return query;
+};
+
+Query.prototype.testRunQuery = function(callback) {
+  var query = this.basicQuery;
+
+  // jshint unused:false
+  // [START run_query]
+  datastore.runQuery(query, function(err, entities) {
```
pcostell: rename entities to tasks for consistency.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+  // [END projection_query]
+
+  return query;
+};
+
+Query.prototype.getAncestorQuery = function() {
+  var datastore = this.datastore;
+
+  // [START anscestor_query]
+  // SELECT * FROM Task WHERE __key__ HAS  ANCESTOR ("TaskList", "default")
```
stephenplusplus: Note to self to remove this.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+
+  this.datastore.runQuery(query, callback);
+};
+
+Query.prototype.testKeyFilter = function(callback) {
+  var datastore = this.datastore;
+
+  // [START key_filter]
+  var query = datastore.createQuery('Task')
+    .filter('__key__ =', datastore.key(['Task', 'someTask']));
```
pcostell: Should be `>`, not `=`.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+  this.datastore.runQuery(query, callback);
+};
+
+Query.prototype.testKindlessQuery = function(callback) {
+  var datastore = this.datastore;
+  var lastSeenKey = this.datastore.key(['Task', Date.now()]);
+
+  // [START kindless_query]
+  var query = datastore.createQuery()
+    .filter('__key__ =', lastSeenKey);
```
pcostell: Should be `>`, not `=`.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+};
+
+Query.prototype.testRunQueryProjection = function(callback) {
+  var query = this.projectionQuery;
+
+  // jshint unused:false
+  // [START run_query_projection]
+  datastore.runQuery(query, function(err, entities) {
+    if (!err) {
+      // Entities found.
```
pcostell: Please build up the priorities and percent_completes list. This is to show how to access the properties in a projection.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+};
+
+Query.prototype.testCursorPaging = function(callback) {
+  var pageSize = 1;
+
+  datastore.createQuery = this.datastore.createQuery;
+
+  // [START cursor_paging]
+  var query = datastore.createQuery('Task')
+    // Pagination is enabled by default in gcloud-node. To switch it off:
```
pcostell: I think this is slightly different. The `autoPaginate` command seems really more of an "autoBatch" feature. Pagination is particularly for users wanting to show pages of results on their site. In this case, they should set the limit to their pageSize and let gcloud get all the results of that page. However, we need to show them how to use cursors and the more_results information to show a "Next Page" link on their site.

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+  delete datastore.createQuery;
+  this.datastore.runQuery(query, callback);
+};
+
+Query.prototype.testEventualConsistentQuery = function(callback) {
+  var query = this.ancestorQuery;
+
+  // jshint unused:false
+  // [START eventual_consistent_query]
+  datastore.runQuery(query, function(err, entities) {
```
pcostell: This snippet is particularly about setting the [read options][1] to be eventually consistent even if they are performing an ancestor query. If this isn't an option, just have this be a comment that indicates that.

[1]: https://github.com/google/googleapis/blob/master/google/datastore/v1beta3/datastore.proto#L259

datastore/query.js
```diff
@@ -0,0 +1,388 @@
...
+    }
+  });
+  // [END eventual_consistent_query]
+
+  this.datastore.runQuery(query, callback);
+};
+
+Query.prototype.testIndex1Query1 = function() {
+  // [START index1_query1]
+  // GQL is not supported in gcloud-node.
```
pcostell: Ah sorry, no need to add these (index*_query*). They are only going to show the GQL format string for them, we decided we don't need per-language snippets for them.

stephenplusplus:
@pcostell thanks! I've made the requested changes in the last few commits.

stephenplusplus:
All the sections are covered now.

datastore/metadata.js
```diff
@@ -0,0 +1,167 @@
...
+
+    entities.forEach(function(entity) {
+      var kind = entity.key.path[1];
+      var propertyName = entity.key.path[3];
+
+      propertiesByKind[kind] = propertiesByKind[kind] || [];
+      propertiesByKind[kind].push(propertyName);
+    });
+
+    // propertiesByKind = {
```
pcostell: We should build (and the other comments) in a language-agnostic way into the docs around the snippets, so no need to include them here.

pcostell:
Pending my small comments, LGTM

datastore/transaction.js
```diff
@@ -0,0 +1,317 @@
...
+
+  // [START transactional_single_entity_group_read_only]
+  function getTaskListEntities(callback) {
+    var error;
+    var taskListEntities;
+
+    datastore.runInTransaction(function(transaction, done) {
+      var taskListKey = datastore.key(['TaskList', 'default']);
+
+      datastore.get(taskListKey, function(err) {
```
stephenplusplus: @pcostell added the lookup here, is this right?

datastore/transaction.js
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    // [START transactional_retry]
+    var async = require('async');
+
+    function attemptTransfer(callback) {
+      transferFunds(fromKey, toKey, 10, callback);
+    }
+
+    async.retry(5, attemptTransfer, callback);
```
stephenplusplus: @pcostell is this too simplified? (formerly, it explicitly checked for a 409)

pcostell: Nope seems ok. I'd rather show the broad concepts here. If there are more odd cases I think they should move into the gcloud-node docs.

datastore/transaction.js
pcostell: looks good.

pcostell:
LGTM

pcostell:
Actually sorry -- Could you combine all the snippets into a single file? I know it's a little more painful, but I think it will make it easier to pull into the docs.

stephenplusplus:
No problem. Let me know if I went too far... https://github.com/GoogleCloudPlatform/nodejs-docs-samples/commit/f48210ee5383f870ce2e3752bf44ec1e82690cb2

Doc snippets can be built with `npm run extract-docs`, which extracts all of the snippets and combines them into a single file by the convention of `{directoryName}/{directoryName}.txt` (so, datastore/datastore.txt and storage/storage.txt). If this doesn't fit with how the existing parser wants to work, or it's just too much, I can pull it back quickly and put all of the source into a single file. LMK!

pcostell:
Ah sorry -- I really just meant for the datastore concepts docs snippets (instead of having entities, indexes, etc). Some of our docs reuse snippets (for example the indexes page actually reuses some snippets from both the entities and the queries page). So something like datastore_concepts.js is probably fine (separate from storage is probably a good thing).

stephenplusplus:
No problem, I think I knew what you meant, I just wanted to offer a solution that kept the source files separate :) It seems like in either case, though, we would end up with the same thing as far as the docs parser is concerned:

  - A single JS file that contains all of the source with all of the snippet blocks
  - A single text file that contains all of the snippet blocks that are compiled from an arbitrary number of JS source files -- [datastore/datastore.txt](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/f48210ee5383f870ce2e3752bf44ec1e82690cb2/datastore/datastore.txt) contains all the snippets (indexes, transactions, etc)

pcostell:
Up to you -- I think I prefer a single js file for the concept snippets (there is going to need to be another one for the getting started snippets). In particular, our doc system can link from a snippet to it's location in the github project, so it'd be better if that linked to js and not txt.

stephenplusplus:
All flattened out. Let me know if this is better!

pcostell:
Looks good, thanks for getting these done!

stephenplusplus:
No problem, thanks for helping me through them!

stephenplusplus:
Can this be merged?

pcostell:
Sounds fine to me (I don't have merge powers) -- can you assign it to someone who does?

jmdobry:
@stephenplusplus Need to resolve merge conflicts. I think you can avoid them if you remove your changes to `package.json` and then create a `datastore/package.json` file, to match how it's now done [here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/storage/package.json). Then squash everything into 1 commit and I'll merge.

stephenplusplus:
Rebased and changed some things to match the others:

- `testSomething.js` -> `something.test.js`
- `process.env.TEST_PROJECT_ID` -> `process.env.TEST_PROJECT_ID || 'nodejs-docs-samples'`

Issue #12, opened 2015-10-02
Add engines field to package.json for samples


JustinBeckwith:
We recently added support for engines in package.json:
http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html

Should we pin down the supported nodejs version in each sample?

jmdobry:
Do we have a list of the available Node versions?

jmdobry:
Found it: https://storage.googleapis.com/gae_node_packages

jmdobry:
Closed by #17

PR #17, opened 2015-10-15
Added "engines" to package.json files. Made package.json and readme f…

jmdobry:
…iles more consistent across samples. Fixed a few samples that were broken once deployed.


jmdobry:
Closes #12

PR #25, opened 2015-11-13
clean up memcache sample

JustinBeckwith:


jmdobry:
I've actually done all this and more in my local memcached cleanup example, just haven't pushed it yet. I'm working on tests. Thanks though.

JustinBeckwith:
np - a lot changed (for the better) after I opened that PR :)

PR #27, opened 2015-11-14
Added MongoDB example

thesandlord:


appengine/mongo/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "appengine-mongo",
+  "description": "An example of using MongoDB with Node.js on Google App Engine.",
+  "version": "0.0.1",
+  "private": true,
+  "license": "Apache Version 2.0",
+  "engines": {
+    "node": "~0.12.7"
```
JustinBeckwith: As of wed, 4.2.1 is now the default version on the image.  Next wed it will be 4.2.2.  Maybe more of a @jmdobry question, but we should do a pass to bring this all up to 4.x.

appengine/mongo/server.js
```diff
@@ -0,0 +1,71 @@
...
+
+mongodb.MongoClient.connect(uri, function(err, db) {
+  
+  if (err) throw err;
+
+  // Create a simple little server. 
+  http.createServer(function(req, res) {
+	  
+	  console.log("test");
+	  console.log(err);
```
JustinBeckwith: probably don't need these :)

appengine/mongo/server.js
```diff
@@ -0,0 +1,71 @@
...
+
+    // Track every IP that has visited this site
+    var collection = db.collection('IPs');
+
+    var ip = {"address": req.connection.remoteAddress};
+
+    collection.insert(ip, function(err, result) {
+
+      if (err) throw err;
```
JustinBeckwith: that'll crash the process, won't it?

appengine/mongo/server.js
```diff
@@ -0,0 +1,71 @@
...
+
+      // push out a range 
+      var iplist = '';
+      collection.find().toArray(function(err, data) {
+        if (err) throw err;
+        console.log('listing data...\n');
+        data.forEach(function(ip) {
+        console.log('IP: ' + ip.address + '\n');
+        iplist += ip.address + '; ';
```
JustinBeckwith: indentation

PR #28, opened 2015-11-16
add task list example application

stephenplusplus:


JustinBeckwith:
LGTM

datastore/package.json
```diff
@@ -4,6 +4,9 @@
   "version": "0.0.1",
   "private": true,
   "license": "Apache Version 2.0",
+  "bin": {
```
jmdobry: All of the samples in the repo should be consistent on how they instruct the user to run the sample. Do you think `npm link` and `"bin"` is better than how the `storage/` samples does it? I'm leaning towards how the `storage/` sample does it because it keeps everything local to the sample without affecting something elsewhere on the user's comp. What are your thoughts?

stephenplusplus: Yeah, probably not worth dealing with issues about the symlink failing due to permissions and such. It does allow a nicer interface, though:

```sh
$ datastore-tasks list
```

vs.

```sh
$ npm run datastore-tasks list
```

I'll switch this over to the second style.

Regarding the consistency note, do you prefer `datastoreTasks` to match `authSample`? Better yet, maybe just `tasks`?:

```sh
$ npm run tasks list
```

jmdobry: Well, to follow how authSamples does it, it will be `npm run getting-started.js -- list`. It's not the prettiest way to execute, but it does help us avoid other potential issues.

jmdobry: authSample follows the convention where the command for executing the sample has the same name as the sample file itself, e.g. `npm run authSample` because the file is named `authSample.js`. So here, the names should be consistent. If you want the command to be `tasks`, then you should rename the file to `tasks.js`

stephenplusplus: Updated

PR #30, opened 2015-11-17
Getting datastore tests running on CI.

jmdobry:


jmdobry:
@stephenplusplus This commit is to get the datastore tests running on Travis. The proper APIs are enabled in the nodejs-docs-samples cloud project, and the datastore samples now initialize gcloud.datastore with the keyfile, but the tests still get a forbidden error. When I switched the datastore samples (locally) to use the latest release of gcloud, it worked. Can you shed any light on this?

Also, these tests need to remove any data they write (and only the data they write) to the datastore at the end of the tests.

stephenplusplus:
The Travis project is whitelisted for v1beta3?

jmdobry:
> The Travis project is whitelisted for v1beta3?

I'm not sure.

I guess that is my other question. These datastore samples, we put them on the master branch, but they're not actually usable by everybody yet? If that's the case, I'm thinking these datastore samples should be pulled back out of master onto a different branch. I think master should only contain samples and docs that pertain to the latest public releases of gcloud.

stephenplusplus:
>  I think master should only contain samples and docs that pertain to the latest public releases of gcloud.

That's fine, whatever you think. All of these Datastore samples are for v1beta3.

PR #31, opened 2015-11-17
Update gcloud-node datastore snippets

pcostell:
- Fixes small typos in the region tags
- Rearranged the build_service snippet to include the correct region.

### DIFF
--- a/datastore/tasks.js
+++ b/datastore/tasks.js
@@ -13,7 +13,6 @@
 
 'use strict';
 
-var gcloud = require('gcloud');
 var input = process.argv.splice(2);
 var command = input.shift();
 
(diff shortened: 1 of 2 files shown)


pcostell:
@stephenplusplus Could you take a look?

datastore/tasks.js
```diff
@@ -22,12 +21,16 @@ if (!projectId) {
   throw new Error('TEST_PROJECT_ID environment variable required.');
 }
 
+// [START build_service
```
stephenplusplus: Missing a trailing ]

datastore/tasks.js
```diff
@@ -22,12 +21,16 @@ if (!projectId) {
...
+var gcloud = require('gcloud');
+
 var datastore = gcloud.datastore({
   projectId: projectId
 });
+// [END build_service]
 
 /*
-// [START build_service]
+Installation and setup instructions.
```
stephenplusplus: Just curious, is this the right way to get these built in the docs?

stephenplusplus:
Sorry for being so sloppy! All looks good to me, though this might clash with #30.

datastore/tasks.js
pcostell: There isn't a great way to do this. However as it was these will all get included as code, rather than markdown. Instead I'll copy your markdown directly into the docs manually.

pcostell:
No worries -- it's easy for me to check because the docs build fails if there are bad regions. We went through several iterations on our snippets before we figured this out.

pcostell:
@jmdobry could you take a look and merge if it looks ok?

jmdobry:
Squash into 1 commit and I'll merge.

pcostell:
Squashed.

PR #32, opened 2015-11-19
Remove datastore v1beta3 samples from master.

jmdobry:
`master` branch of nodejs-docs-samples should only have samples that work with the latest release of gcloud and don't depend on the user having access to beta APIs (or APIs where the user would otherwise need their project to be whitelisted).

This updates (reverts) the `tasks.js` sample to work with gcloud `0.25.0` and removes the concepts.js samples which were made for v1beta3 of Datastore. The `datastore-v1beta3` branch adds back the concepts.js samples, and can be merged into master later.


jonparrott:
LGTM.

stephenplusplus:
I'll get on cleaning up anything that was created with the tests for the concepts samples.

jmdobry:
Cool. I already made it so the tasks.js tests clean up after themselves, and I don't think the concepts.js tests have had a chanceto make a mess in the nodejs-docs-samples datastore, because the nodejs-docs-samples project isn't whitelisted for v1beta32 (and thus the tests for the `datastore-v1beta3` won't pass).

jmdobry:
For reference: #33

PR #40, opened 2015-11-20
Simplify app.yaml files, reduce # of e2e tests, and cleanup for webpa…

jmdobry:
…ck sample.

### DIFF
--- a/appengine/express-memcached-session/app.yaml
+++ b/appengine/express-memcached-session/app.yaml
@@ -14,5 +14,4 @@
 runtime: nodejs
 vm: true
 env_variables:
-  PORT: 8080
   MEMCACHE_URL: localhost:11211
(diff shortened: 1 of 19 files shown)


jmdobry:
In this PR I scaled back from 9 e2e tests (3 per job) to 3 e2e tests (1 per job). That makes one e2e test each for Node.js stable, 0.12, and 0.10. As you can see the first test passed! https://travis-ci.org/GoogleCloudPlatform/nodejs-docs-samples/jobs/92324990

I also maintained variety in the remaining e2e tests. One deploys Express _and_ promotes it, one deploys Grunt which needs the postinstall hook, and one deploys Koa, which needs the --harmony flag.

appengine/express-memcached-session/app.yaml
```diff
@@ -14,5 +14,4 @@
 runtime: nodejs
 vm: true
 env_variables:
-  PORT: 8080
```
JustinBeckwith: Nice!  Is this finally defaulted in the base runtime?

JustinBeckwith:
LGTM

jonparrott:
LGTM.

PR #45, opened 2015-12-07
Added sendgrid example app. Also used for Compute Engine snippet.

jmdobry:


appengine/sendgrid/README.md
```diff
@@ -0,0 +1,20 @@
+## Express.js + Sendgrid on Google App Engine
```
JustinBeckwith: Don't think we really need to specify express for each of the samples that happen to use express.  Would the sendgrid specific code be any different if we were using sails or koa?

appengine/sendgrid/app.js
```diff
@@ -0,0 +1,92 @@
...
+app.post('/hello', function(req, res, next) {
+  Sendgrid.send({
+    from: 'no-reply@appengine-sendgrid-demo.com', // From address
+    to: req.body.email, // To address
+    subject: 'Hello World!', // Subject
+    text: 'Sendgrid on Google App Engine with Node.js' // Content
+  }, function (err, json) {
+    if (err) {
+      return next(err);
+    } else {
```
JustinBeckwith: total nit - do we really need the else in this case?  Can't we just rely on the return on the err case to get us out of scope?

appengine/sendgrid/app.js
```diff
@@ -0,0 +1,92 @@
...
+      // Render the index route on success
+      return res.render('index', {
+        sent: true
+      });
+    }
+  });
+});
+// [END hello]
+
+if (module === require.main) {
```
JustinBeckwith: Is this just for test purposes?

appengine/sendgrid/app.yaml
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# [START app_yaml]
+runtime: nodejs
+vm: true
+env_variables:
+  SENDGRID_API_KEY: <your-mailgun-api-key>
+# [END app_yaml]
+skip_files:
```
JustinBeckwith: @dlorenc when can we start removing these?

appengine/sendgrid/app.js
```diff
@@ -0,0 +1,92 @@
...
+    function () {
+      var address = server.address().address;
+      var port = server.address().port;
+      console.log('App listening at http://%s:%s', address, port);
+      console.log('Press Ctrl+C to quit.');
+    }
+  );
+  // [END server]
+} else {
+  modules.exports = function send() {
```
JustinBeckwith: What calls this?

appengine/sendgrid/README.md
jmdobry: no

appengine/sendgrid/app.js
jmdobry: >  do we really need the else in this case?

nope

appengine/sendgrid/app.js
jmdobry: yep

appengine/sendgrid/app.js
jmdobry: What comes after line 74 is the snippet used on the Compute Engine Sendgrid tutorial page. The `module.exports` is for testing purposes and so the code makes a little more sense.

appengine/sendgrid/app.js
JustinBeckwith: I kind of feel like this is modifying the control flow based on if it's a test or not.  Generally better to avoid if possible, but I get why it's there.

appengine/sendgrid/app.js
JustinBeckwith: Ah - that makes a lot more sense.  Maybe stuff this in another js file in the repo?  With a big comment that explains what it does and why?

appengine/sendgrid/app.js
jmdobry: Conditionally starting the server based on whether the file is being executed directly vs when the file is being `require`'d is not just for testing. It's very useful to be able to export the express app instance so something else can use it. That's exactly what the express appengine sample does, and in that example the server is actually started by the `bin/www`.

JustinBeckwith: Fair point.  Mind changed, carry on.

appengine/sendgrid/app.js
```diff
@@ -0,0 +1,93 @@
...
+// [END setup]
+
+var app = express();
+
+// Setup view engine
+app.set('views', path.join(__dirname, 'views'));
+app.set('view engine', 'jade');
+
+// Parse form data
+app.use(bodyParser.json());
```
jonparrott: Is this needed? I don't see us parsing any json data.

appengine/sendgrid/app.js
```diff
@@ -0,0 +1,93 @@
...
+      var address = server.address().address;
+      var port = server.address().port;
+      console.log('App listening at http://%s:%s', address, port);
+      console.log('Press Ctrl+C to quit.');
+    }
+  );
+  // [END server]
+} else {
+  exports.app = app;
+  exports.send = function () {
```
jonparrott: Why this? You're already showing how to send?

jmdobry: This little snippet is what will be shown on the Google Compute Engine SendGrid tutorial.

appengine/sendgrid/app.js
jmdobry: removed.

appengine/sendgrid/app.js
jonparrott: Why can't it just re-use what's above?

jmdobry: What's above has `req.body` and `res.render`, stuff that's out of context of what's here: https://cloud.google.com/compute/docs/tutorials/sending-mail/using-sendgrid#nodejssendgrid, which talks talks about just creating a file that sends an email, not start a web server and responding to requests.

jonparrott: Why not split `send` into its own parameterized function and just have the request handler call that?

jmdobry: I suppose I could do that, but it would mean an extra code block needs to be shown on the Express + SendGrid tutorial

jonparrott: That's fine, I've done it in other samples. It's better than code duplication like this.

jmdobry:
Okay, I moved the Compute Engine snippet to its own file.

jonparrott:
LGTM.

PR #50, opened 2015-12-14
Added instructions for running the tests. Fixes #49.

### LINKED ISSUE #49: Use a port picker to prevent port reuse
When running `npm test`, at least one test appears to hold on to port 8080 causing later tests to fail.

Rather than using port 8080 for all tests a port picker should be used for each test so that they don't interfere with each other.

jmdobry:


jonparrott:
LGTM.

PR #53, opened 2016-01-14
Add Logging API example.

jmdobry:


logging/list.js
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START list]
+var google = require('googleapis');
+var projectId = process.env.TEST_PROJECT_ID || '<your-project-id>';
+var logging = google.logging('v2beta1').entries;
```
JustinBeckwith: We have veneer support for this now - can we use that instead?
https://github.com/GoogleCloudPlatform/gcloud-node#google-cloud-logging-beta

jmdobry:
@jonparrott ready for review

logging/export.js
```diff
@@ -0,0 +1,92 @@
...
+}
+// [END listSinks]
+
+// [START createSink]
+function createSink(callback) {
+  var gcs = gcloud.storage();
+
+  // create a new sink
+  //
+  // This method only works if you are authenticated as yourself, e.g. using the
```
jonparrott: Why?

logging/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "private": true,
+  "license": "Apache Version 2.0",
+  "scripts": {
+    "list": "node list.js",
+    "write": "node write.js",
+    "export": "node export.js"
+  },
+  "dependencies": {
+    "gcloud": "^0.27.0"
+  }
```
jonparrott: engines?

jonparrott:
LGTM with small nits.

logging/export.js
jmdobry: I don't know, that's what the cloud-node logging docs say.

logging/package.json
jmdobry: why do I need engines?

jonparrott: Shouldn't we at least specify a minimum node version?

logging/export.js
JustinBeckwith: Calling @stephenplusplus

logging/package.json
JustinBeckwith: I'd prefer to provide a minimum version in engines as well.

jmdobry: I think it's unnecessary, but if you guys want it, what version do you want me to specify? It's being tested on v0.10, v0.12, and v5.x

JustinBeckwith: I'm fine either way - just being overly safe :) If we are going to do it, I'd say >= 0.12

logging/export.js
stephenplusplus: https://github.com/GoogleCloudPlatform/gcloud-node/issues/842#issuecomment-140737616

// @munangst

logging/package.json
jmdobry: Lol. I can't move on to anything else until I comment on how much I don't want to add this field.

My stance is that for samples where we actually invite the user to consume and/or deploy (like when there's an `app.yaml` file) then we should specify the `engines` field. e.g. The package.json for `gcloud` should specify `engines`, our MVMs samples should specify `engines`, etc.

Otherwise, for these samples, it's just bloat, something to be maintained and verified (or not). This sample is not going to be installed by NPM or consumed by a deploy or executed by some platform. There isn't anything that's going to programmatically read that field, only eyeballs will see it. I can set `engines` to `0.6.x` and run it with via NPM with Node v5 and NPM doesn't care. If I set it to >=0.12, that implies it only works on >=0.12, which isn't true. I fail to see a benefit that outweighs the cost of maintaining a magic number across dozens of samples.

> It is a strange fate we should suffer so much fear and doubt… over so small a thing.
>
> -- Boromir

jonparrott: I would argue that we do want these samples to be consumable. All of the python samples are runnable applications with their own dependencies.

jmdobry: Sure these samples can be run locally and have their own dependencies, but nothing about that cares about the `engines` field.

JustinBeckwith: I am neutral :)  Whatever y'all decide is fine by me.

jmdobry: I added the `engines` field, but set it to `>= 0.10.x`, since that's the earliest version the samples are being tested on (also tested on 0.12 and 5)

PR #58, opened 2016-01-26
Move back to container-based Travis builds and removed some redundant tests.

jmdobry:


jmdobry:
Tests should be faster and a little bit more stable now. Ready for review.

jonparrott:
LGTM.

PR #62, opened 2016-01-30
Don't store ip addresses.

### LINKED ISSUE #61: Datastore / CloudSQL samples should only store first two octtets of IP
Principle of only using data you need.

jmdobry:
Fixes #61


jonparrott:
LGTM.

PR #70, opened 2016-02-27
Env vars are parsed from app.yaml on local run. Added SERVER_URL to app.yaml

thesandlord:
I am at a event and a lot of people want to run Parse Server on App Engine. The env variables are very confusing when running locally. This patch makes it so the variables set in the app.yaml are applied to the local config as well.

It might make sense to have a "LOCAL_DATABASE_URI" as well in app.yaml, so you can run a local database.

### DIFF
--- a/appengine/parse-server/README.md
+++ b/appengine/parse-server/README.md
@@ -11,19 +11,22 @@ app to [Google App Engine Managed VMs](https://cloud.google.com/appengine).
 1. Setup a MongoDB server. Here are two possible options:
   1. Create a Google Compute Engine virtual machine with [MongoDB pre-installed](https://cloud.google.com/launcher/?q=mongodb).
   1. Use [MongoLab](https://mongolab.com/google/) to create a free MongoDB deployment on Google Cloud Platform.
-
-## Running locally
+  
+## Downloading Files
 
 1. `git clone https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git`
 1. `cd appengine/parse-server`
+
+## Running locally
+
 1. `npm install`
-1. Set the necessary [environment variables](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/appengine/parse-server/server.js#L23).
+1. Set the necessary [environment variables](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/appengine/parse-server/config.json).
 1. `npm start`
 
 ## Deploy
 
-1. Modify `app.yaml` as needed.
-1. `gcloud preview app deploy`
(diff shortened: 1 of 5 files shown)


JustinBeckwith:
Thanks for doing this!  This tripped up a bunch of people on the medium post as well.  The big problem here is that this won't be supported on windows :/  Instead - why don't we just use nconf and put the config vars in a json file?

thesandlord:
How do those get pulled into the app.yaml? I would rather have one place for all the env vars. Cross platform is nice, I can convert that bash script into node.

thesandlord:
Oh you are saying not to use App Engine env vars at all. Got it, I think that's a good idea.

thesandlord:
@JustinBeckwith changed it to use nconf

appengine/parse-server/README.md
```diff
@@ -17,12 +17,12 @@ app to [Google App Engine Managed VMs](https://cloud.google.com/appengine).
 1. `git clone https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git`
 1. `cd appengine/parse-server`
 1. `npm install`
-1. Set the necessary [environment variables](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/appengine/parse-server/server.js#L23).
+1. Set the necessary [environment variables](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/appengine/parse-server/app.yaml#L18).
```
JustinBeckwith: We just need them to update config.json once now, both for local and in prod right?

appengine/parse-server/server.js
```diff
@@ -16,17 +16,21 @@
 
 // [START app]
 var express = require('express');
+var fs = require('fs');
```
JustinBeckwith: Do we need fs here?

thesandlord: fixed (this was in the docs but I don't think it is required)

JustinBeckwith:
lgtm

jmdobry:
@thesandlord Can you squash into 1 commit? Then I'll merge. Thanks.

thesandlord:
Before merging, can we make sure @JustinBeckwith  [blog post](https://medium.com/google-cloud/deploying-parse-server-to-google-app-engine-6bc0b7451d50#.lfzepmx8s) and @jmdobry [docs](https://cloud.google.com/nodejs/resources/frameworks/parse-server) are updated?

JustinBeckwith:
Oh no, my blog post has become a living doc :)  I'll update the post as soon as this is merged.

thesandlord:
@jmdobry can we merge now? I see the docs are still in the old format, do we need to merge first?

jmdobry:
[Docs](https://cloud.google.com/nodejs/resources/frameworks/parse-server) are updated, @JustinBeckwith can you update your blog post?

JustinBeckwith:
Blog post updated!

PR #85, opened 2016-03-21
Add Monitoring Tests

waprin:
@JustinBeckwith @jonparrott 

Unfortunately we have to code freeze these samples until GCP Next is over next week, but feel free to  review at your convenience. 

Had to refactor the main code a bit, since for my custom metric tests, I generate a new custom metric and write one value for it. I made CustomMetrics a class so I could attach the metric name to its state. Not sure if this the most idiomatic way to do it for this repo so let me know.

The tests aren't going to work until the API no longer requires a whitelist. That might be now or very soon (I'll double check), otherwise we can add your test project as an owner of mine.

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -6,47 +6,39 @@ Monitoring API to retrieve API data.
 `create_custom_metric.js` demonstrates how to create a custom metric, write a timeseries value to it,
 and read it back.
 
-
 # Run locally
 
 Create local credentials by running the following command and following the oauth2 flow:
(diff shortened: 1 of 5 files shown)


jmdobry:
Hmm, the test is failing, says `google.monitoring` is not a function. I can see that it's using google-api-nodejs-client 3.1.0, which should support the monitoring API. Clearing the Travis cache and trying again...

jmdobry:
I figured it out.

PR #91, opened 2016-04-01
BigQuery samples.

jmdobry:


bigquery/dataset_size.js
```diff
@@ -0,0 +1,152 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+var async = require('async');
+var google = require('googleapis');
+var tables = google.bigquery('v2').tables;
```
JustinBeckwith: Why aren't you using the veneer?
https://github.com/GoogleCloudPlatform/gcloud-node#google-bigquery

JustinBeckwith: Also - we need to figure out the best place to keep samples that use the api client vs the veneer.  We have a set of samples over in the other repository - when does it make sense to keep samples here vs the api client repo?

jmdobry: For the most part, this repository is for samples that appear in the docs on cloud.google.com.

JustinBeckwith: I'd prefer we change them to use the veneers if possible

jmdobry: Working on it.

jmdobry:
@jsimonweb Can I get a LGTM today if you have time? Thanks.

JustinBeckwith:
lgtm - thanks for updating to gcloud!

jmdobry:
The tests failed on Node v0.10 I think because of a caching issue with the node_modules folder, but for some reason Travis won't let me clear the cache right now.

jsimonweb:
LGTM

PR #94, opened 2016-04-05
Compute Engine samples

jmdobry:


computeengine/vms_api.js
```diff
@@ -0,0 +1,84 @@
...
+    callback(null, authClient);
+  });
+}
+// [END auth]
+
+// [START list]
+/**
+ * @param {Function} callback Callback function.
+ */
+function getVmsExample(callback) {
```
JustinBeckwith: This looks the same as the example above, just using the api client instead of gcloud.  Why?

jonparrott:
LGTM

computeengine/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "Collection of Node.js samples on Google Compute Engine.",
+  "version": "0.0.1",
+  "private": true,
+  "license": "Apache Version 2.0",
+  "author": "Google Inc.",
+  "engines": {
+    "node": ">=0.10.x"
+  },
+  "scripts": {
+    "sendgrid": "node sendgrid.js",
```
JustinBeckwith: nothing in this sample is really compute engine specific - at what point do we just have a single sendgrid sample, and use that for g[cka]e?

jmdobry: Yes

jmdobry: Just realized that wasn't a yes/no question. 😆

PR #99, opened 2016-04-13
Bump gcloud requirement to version based on Datastore v1beta3.

### LINKED ISSUE #98: Datastore tutorial seems to depend on v1beta2
If a project already had v1beta2 enabled, the sample works.

But it fails with a new project even if v1beta3 is enabled. The error is "Service Unavailable - Backend Error". Enabling v1beta2 on the new project will cause the sample to work.

pcostell:
Fixes #98.

### DIFF
--- a/datastore/package.json
+++ b/datastore/package.json
@@ -14,6 +14,6 @@
   },
   "dependencies": {
     "async": "^1.5.2",
-    "gcloud": "^0.28.0"
+    "gcloud": "^0.30.3"
   }
 }


jmdobry:
You changed the version of Node.js not `gcloud`.

jmdobry:
Just changing it to `"gcloud": "^0.30.3"` would be best

pcostell:
Well that was painful :-/ Sorry about that.

PR #100, opened 2016-04-13
Upgraded dependencies

jmdobry:
- No longer testing on Node v0.10
- Switched to Codecov
- Fixed some Datastore stuff related to upgrading to gcloud-node 0.30.3


jsimonweb:
LGTM

PR #102, opened 2016-04-19
Added requested Mailjet samples.

jmdobry:


appengine/mailjet/app.js
```diff
@@ -0,0 +1,86 @@
...
+app.get('/', function(req, res) {
+  res.render('index');
+});
+// [END index]
+
+// [START hello]
+app.post('/hello', function(req, res, next) {
+  var options = {
+    // From
+    FromEmail: 'no-reply@appengine-mailjet-demo.com',
```
jonparrott: Shouldn't this be configurable? Or does mailjet let you send from anything?

jonparrott:
LGTM with one nit.

appengine/mailjet/app.js
jmdobry: As long as they've verified that email address it will send.

jonparrott: Mmmmmkay. SGTM.

PR #110, opened 2016-05-12
Add Cloud Speech sample.

jmdobry:


JustinBeckwith:
I feel a little icky about including the binaries in the git repo.  Opinions on git-lfs for this use?

jmdobry:
My opinion is to leave them in the git repo. That's how the other <lang>-docs-samples repos do it.

jonparrott:
binary files are fine here.

@jerjou to review.

speech/recognize.js
```diff
@@ -0,0 +1,111 @@
...
+
+'use strict';
+
+var google = require('googleapis');
+var async = require('async');
+var fs = require('fs');
+var url = 'https://speech.googleapis.com/$discovery/rest';
+
+function auth(callback) {
+  google.auth.getApplicationDefault(function (err, authClient) {
```
jonparrott: I believe speech requires a service account for the time being.

jonparrott:
LGTM, but please let jerjou take a look before merging.

speech/recognize.js
jmdobry: I got it to work both with an API key and letting it pick credentials.

jmdobry:
This is basically the Node.js version of: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/speech/api/speech_rest.py

speech/recognize.js
```diff
@@ -0,0 +1,127 @@
+// Copyright 2015-2016, Google, Inc.
```
jerjou: I think this can just be 2016, no? That's when you wrote the sample?

test/speech/recognize.test.js
```diff
@@ -0,0 +1,32 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+var test = require('ava');
```
jerjou: 'ava'? Looks like 'java' without the 'j'. I'm assuming this works? :)

jerjou:
`quit.raw` was for the streaming sample (so not necessary here for the batch one).
Also - looks like there's a vision png in this PR?

jerjou:
Otherwise, LGTM

test/speech/recognize.test.js
jmdobry: Yes, this is correct.

jmdobry:
Hopefully I can get a streaming example added soon.

PR #111, opened 2016-05-12
Add Cloud Vision label detection sample.

jmdobry:


jonparrott:
LGTM, my only concern is whether or not we have the rights to distribute these images under Apache 2.

jmdobry:
The images came from here: https://github.com/GoogleCloudPlatform/cloud-vision/tree/master/data/label

jonparrott:
Okay, LGTM.

PR #113, opened 2016-05-16
Add Cloud Vision text detection sample.

jmdobry:


vision/textDetection.js
```diff
@@ -0,0 +1,284 @@
...
+var gcloud = require('gcloud');
+var natural = require('natural');
+var redis = require('redis');
+// [END import_libraries]
+
+// [START authenticate]
+// You must set the GOOGLE_APPLICATION_CREDENTIALS and GCLOUD_PROJECT
+// environment variables to run this sample. See:
+// https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/docs/authentication.md
+var projectId = process.env.GCLOUD_PROJECT;
```
JustinBeckwith: You shouldn't have to do this really.  If you set this env var, gcloud just figures it out.  You should be able to drop this line, and the projectId config below.

jmdobry: Nice!

jonparrott:
LGTM once justin's comment is addressed.

PR #116, opened 2016-05-23
Switched from JSHint to Semistandard and upgraded dependencies.

jmdobry:


jonparrott:
This (cautiously) LGTM.

I really would've preferred the following set of commits in this PR:
1. Switch to semistandard.
2. Fix style issues.
3. Fix nits.

This changeset is difficult to review with accuracy because you made several small, non-style changes.

PR #142, opened 2016-06-29
Switch from "gcloud preview app" to "gcloud app"

jmdobry:


JustinBeckwith:
LGTM

jonparrott:
LGTM

jmdobry:
The Travis tests keep timing out because of Ava I believe, so I'm experimenting with switching back to plain old Mocha.

PR #143, opened 2016-06-30
Switch back to mocha from ava.

jmdobry:


jonparrott:
LGTM, it's a test-only change.

PR #144, opened 2016-06-30
Add Speech API streaming sample.

jmdobry:


speech/recognize_streaming.js
```diff
@@ -0,0 +1,138 @@
...
+
+// [START authenticating]
+function getSpeechService (callback) {
+  var authClient = googleAuth({
+    scopes: [
+      'https://www.googleapis.com/auth/cloud-platform'
+    ]
+  });
+
+  authClient.getAuthClient(function (err, authClient) {
```
jonparrott: doesn't `authClient` alias a variable in the outer scope? Should we give it a different name?

jonparrott:
LGTM with small nit.

jmdobry:
Done

PR #147, opened 2016-07-11
Shows how to exclude property from index (for docs)

WalterHub:


WalterHub:
Actually, it appears that excludeFromIndexes is ignored.

PR #159, opened 2016-07-31
Refactored tests

jmdobry:
Refactored tests into unit tests and system tests, which can be run separately or together. Also, moved tests to be right next to the source files that they're testing.


jmdobry:
@jsimonweb can you review?

Hgergely:
Hey Jason. Can I ask how can we set up language code for streaming? It
works nicely using the langage code with the non streaming api but the
streaming one gives an error unkown field when using the same parameter.
Regards.

jsimonweb:
LGTM

PR #160, opened 2016-08-04
Upgraded storage sample

jmdobry:


storage/buckets.js
```diff
@@ -0,0 +1,132 @@
...
+
+// [START create]
+/**
+ * Creates a new bucket with the given name.
+ *
+ * @param {string} name The name of the new bucket.
+ * @param {function} cb The callback function.
+ */
+function createBucketExample (name, callback) {
+  if (!name) {
```
jonparrott: won't the client library handle this?

storage/buckets.js
```diff
@@ -0,0 +1,132 @@
...
+ * @param {function} cb The callback function.
+ */
+function listBucketsExample (callback) {
+  // See https://googlecloudplatform.github.io/gcloud-node/#/docs/storage?method=getBuckets
+  storage.getBuckets(function (err, buckets, apiResponse) {
+    if (err) {
+      return callback(err);
+    }
+
+    console.log('Found ' + buckets.length + ' buckets!');
```
jonparrott: print the list

storage/buckets.js
jmdobry: I think it does, don't know what came over me when I decided to check the argument in my sample function.

storage/buckets.js
jmdobry: The list will get printed when the user runs the sample. If I print it here it will get printed twice.

jonparrott: Gotcha.

jonparrott:
Looks (mostly) good to me.

jmdobry:
Waiting on Travis

PR #168, opened 2016-08-13
Start storagetransfer samples.

jmdobry:


jmdobry:
@ace-n Can you review?

Note, I can't reliably system test the 3 of the transfer operations functions, but they're covered 100% in the unit tests.

storage/system-test/transfer.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+'use strict';
+
+var uuid = require('node-uuid');
+var program = require('../transfer');
+var Storage = require('@google-cloud/storage');
+
+var storage = Storage();
+var bucketName = 'nodejs-docs-samples-test-' + uuid.v4();
+var bucketName2 = 'nodejs-docs-samples-test-' + uuid.v4();
```
ace-n: Mind calling this something like "firstBucketName" and "secondBucketName"?

storage/system-test/transfer.test.js
```diff
@@ -0,0 +1,125 @@
...
+        });
+      });
+    });
+  });
+
+  describe('createTransferJob', function () {
+    it('should create a storage transfer job', function (done) {
+      program.createTransferJob(bucketName, bucketName2, date, time, description, function (err, transferJob) {
+        assert.ifError(err);
+        jobName = transferJob.name;
```
ace-n: What is the purpose of this?

storage/system-test/transfer.test.js
```diff
@@ -0,0 +1,125 @@
...
+        assert.equal(transferJob.name.indexOf('transferJobs/'), 0);
+        assert.equal(transferJob.description, description);
+        assert.equal(transferJob.status, 'ENABLED');
+        assert(console.log.calledWith('Created transfer job: %s', transferJob.name));
+        setTimeout(done, 2000);
```
ace-n: Is there a cleaner way to implement this delay?

storage/system-test/transfer.test.js
```diff
@@ -0,0 +1,125 @@
...
+      });
+    });
+  });
+
+  describe('updateTransferJob', function () {
+    it('should update a transferJob', function (done) {
+      program.updateTransferJob(jobName, 'status', status, function (err, transferJob) {
+        assert.ifError(err);
+        assert.equal(transferJob.name, jobName);
```
ace-n: Should we create an object and use `deepEqual` or something like that here?

storage/system-test/transfer.test.js
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  describe('listTransferJobs', function () {
+    it('should list transferJobs', function (done) {
+      program.listTransferJobs(function (err, transferJobs) {
+        assert.ifError(err);
+        var matchingTransferJobs = transferJobs.filter(function (transferJob) {
+          return transferJob.name === jobName;
+        });
+        assert.equal(matchingTransferJobs.length, 1);
```
ace-n: Create an object?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+    },
+    transferOperations: {
+      get: sinon.stub().callsArgWith(1, null, transferOperationMock),
+      pause: sinon.stub().callsArgWith(1, null, transferOperationMock),
+      resume: sinon.stub().callsArgWith(1, null, transferOperationMock),
+      list: sinon.stub().callsArgWith(1, null, { operations: [transferOperationMock] })
+    }
+  };
+  var googleapisMock = {
+    storagetransfer: sinon.stub().returns(storagetransferMock),
```
ace-n: storageTransferMock?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+        assert.ifError(callback.firstCall.args[0], 'callback did not receive error');
+        assert.strictEqual(callback.firstCall.args[1], sample.mocks.transferJob, 'callback received transfer job');
+
+        sample.program.createTransferJob(bucketName, bucketName, '2016/08/11', '15:30', description, callback);
+
+        assert(callback.calledTwice, 'callback called twice');
+        assert.equal(callback.secondCall.args.length, 2, 'callback received 2 arguments');
+        assert.ifError(callback.secondCall.args[0], 'callback did not receive error');
+        assert.strictEqual(callback.secondCall.args[1], sample.mocks.transferJob, 'callback received transfer job');
+        assert.equal(sample.mocks.storagetransfer.transferJobs.create.secondCall.args[0].resource.description, description, 'description was set');
```
ace-n: This seems to test description-setting as well as transfer job creation - perhaps split each of these into its own test?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+      it('should require "srcBucketName"', function () {
+        var sample = getSample();
+        var callback = sinon.stub();
+
+        sample.program.createTransferJob(null, null, null, null, null, callback);
+
+        assert(callback.calledOnce, 'callback called once');
+        assert.equal(callback.firstCall.args.length, 1, 'callback received 1 argument');
+        assert(callback.firstCall.args[0], 'callback received error');
+        assert.equal(callback.firstCall.args[0].message, '"srcBucketName" is required!', 'error has correct message');
```
ace-n: Why not combine these?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+      it('should require "destBucketName"', function () {
+        var sample = getSample();
+        var callback = sinon.stub();
+
+        sample.program.createTransferJob(bucketName, null, null, null, null, callback);
+
+        assert(callback.calledOnce, 'callback called once');
+        assert.equal(callback.firstCall.args.length, 1, 'callback received 1 argument');
+        assert(callback.firstCall.args[0], 'callback received error');
+        assert.equal(callback.firstCall.args[0].message, '"destBucketName" is required!', 'error has correct message');
```
ace-n: Combine these?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+        var sample = getSample();
+        var callback = sinon.stub();
+        sample.mocks.googleapis.auth.getApplicationDefault.callsArgWith(0, error);
+
+        sample.program.createTransferJob(bucketName, bucketName, 'time', 'date', null, callback);
+
+        assert(callback.calledOnce, 'callback called once');
+        assert.equal(callback.firstCall.args.length, 1, 'callback received 1 argument');
+        assert(callback.firstCall.args[0], 'callback received error');
+        assert.equal(callback.firstCall.args[0].message, error.message, 'error has correct message');
```
ace-n: Combine these?

storage/test/transfer.test.js
```diff
@@ -0,0 +1,639 @@
...
+        assert.equal(callback.firstCall.args.length, 2, 'callback received 2 arguments');
+        assert.ifError(callback.firstCall.args[0], 'callback did not receive error');
+        assert.deepEqual(callback.firstCall.args[1], [sample.mocks.transferOperation], 'callback received transfer operations');
+
+        sample.program.listTransferOperations(jobName, callback);
+
+        assert(callback.calledTwice, 'callback called twice');
+        assert.equal(callback.secondCall.args.length, 2, 'callback received 2 arguments');
+        assert.ifError(callback.secondCall.args[0], 'callback did not receive error');
+        assert.deepEqual(callback.secondCall.args[1], [sample.mocks.transferOperation], 'callback received transfer operations');
```
ace-n: These seem redundant - the second one includes a `jobName` parameter, but doesn't seem to test anything new.

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+}
+// [END auth]
+
+// [START create_transfer_job]
+/**
+ * Review the transfer operations associated with a transfer job.
+ *
+ * @param {string} srcBucketName The name of the source bucket.
+ * @param {string} destBucketName The name of the destination bucket.
+ * @param {string} [description] Optional description for the new transfer job.
```
ace-n: I assume these brackets are correct?

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+      return callback(err);
+    }
+
+    // The createScopedRequired method returns true when running on GAE or a
+    // local developer machine. In that case, the desired scopes must be passed
+    // in manually. When the code is  running in GCE or a Managed VM, the scopes
+    // are pulled from the GCE metadata server.
+    // See https://cloud.google.com/compute/docs/authentication for more
+    // information.
+    if (authClient.createScopedRequired && authClient.createScopedRequired()) {
```
ace-n: Nit: combine these into one boolean clause.

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+  auth(function (err, authClient) {
+    if (err) {
+      return callback(err);
+    }
+
+    storagetransfer.transferJobs.get({
+      auth: authClient,
+      projectId: process.env.GCLOUD_PROJECT,
+      jobName: jobName
+    }, function (err, transferJob) {
```
ace-n: Nit: put the `function` on the next line?

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+function listTransferJobs (callback) {
+  auth(function (err, authClient) {
+    if (err) {
+      return callback(err);
+    }
+
+    storagetransfer.transferJobs.list({
+      auth: authClient,
+      filter: JSON.stringify({ project_id: process.env.GCLOUD_PROJECT })
+    }, function (err, result) {
```
ace-n: `function` on next line?

storage/system-test/transfer.test.js
jmdobry: `jobName` is re-used in subsequent tests.

storage/system-test/transfer.test.js
jmdobry: That's as short as it gets.

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+
+    if (jobName) {
+      filter.job_names = [jobName];
+    }
+
+    storagetransfer.transferOperations.list({
+      name: 'transferOperations',
+      filter: JSON.stringify(filter),
+      auth: authClient
+    }, function (err, result, apiResponse) {
```
ace-n: Next line?

storage/transfer.js
```diff
@@ -0,0 +1,453 @@
...
+
+  auth(function (err, authClient) {
+    if (err) {
+      return callback(err);
+    }
+
+    storagetransfer.transferOperations.get({
+      name: transferOperationName,
+      auth: authClient
+    }, function (err, transferOperation) {
```
ace-n: Next line?

storage/system-test/transfer.test.js
jmdobry: Nah, I don't want to validate that the Storage Transfer API is responding with all of the correct fields. Ideally they have their own tests for that. This assertion is sufficient to say yes, the sample used the client library to hit a service and we got a response back the _looks_ okay.

Now, if the sample itself were using some other fields, then maybe we'd want to verify them.

storage/test/transfer.test.js
jmdobry: I kept the "t" lowercase because that's how the api itself is named when you're building the client. `var storagetransfer = google.storagetransfer('v1');`

storage/test/transfer.test.js
jmdobry: This tests that the proper argument was pass _to_ the create method, so it's relevant. I could actually added another assertion near the first call that verifies that description was _not_ passed to create, for consistency.

storage/test/transfer.test.js
jmdobry: Because JavaScript, e.g.

```js
var a = new Error('foo');
var b = new Error('foo');
a == b; // false
a.message == b.message; // true
```
Can't assert non-strict equality on instances of `Error`

ace-n:
A couple nits, but otherwise LTGM.

storage/transfer.js
jmdobry: Square brackets mean "optional" in JSDoc notation.

storage/transfer.js
jmdobry: Can't, the first part verifies that the method exists (protect against null pointer exception), the second part executes the method to actually check if the scope needs to be created.

storage/transfer.js
jmdobry: This is standard Node.js format. If I moved the function onto the next line, I'd also have to move the opening curly brace on line 137 to the next line, but I don't do that unless lines are exceeding 80 characters.

storage/transfer.js
jmdobry: See https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/168#discussion_r75028693

storage/transfer.js
jmdobry: See https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/168#discussion_r75028693

storage/transfer.js
jmdobry: See https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/168#discussion_r75028693

storage/system-test/transfer.test.js
jmdobry: See https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/168#discussion_r75027526

storage/system-test/transfer.test.js
jmdobry: Will do.

storage/test/transfer.test.js
ace-n: Does 'assert.deepEqual(a, b)' work though? (Would test, but am on my phone.)

storage/transfer.js
ace-n: Gotcha - I'm familiar with this pattern; I was hoping that the two conditions (null ptr and needing to create the scope) were one and the same.

storage/test/transfer.test.js
jmdobry: Just tried it. `assert.deepEqual(new Error('foo'), new Error('bar'))` does not throw, so can't rely on that...

PR #172, opened 2016-08-18
Storage ACL samples.

jmdobry:
- [x] - Cloud Storage ACL samples
- [x] - Unit tests
- [x] - System tests
- [x] - Sample CLI
- [x] - storage/README.md updated

### DIFF
--- a/storage/README.md
+++ b/storage/README.md
@@ -11,6 +11,7 @@ amount of data at any time.
 
 * [Setup](#setup)
 * [Samples](#samples)
+  * [ACL (Access Control Lists)](#acl-access-control-lists)
   * [Buckets](#buckets)
   * [Encryption](#encryption)
   * [Files](#files)
@@ -27,6 +28,41 @@ amount of data at any time.
 [run]: ../README.md#how-to-run-a-sample
 
 ## Samples
+
+### ACL (Access Control Lists)
+
+View the [documentation][acl_docs] or the [source code][acl_code].
+
+__Usage:__ `node acl --help`
+
+```
+Commands:
+  add <entity> <role>  Add access controls on a bucket or file.
+  get [entity]         Get access controls on a bucket or file.
+  delete <entity>      Delete access controls from a bucket or file.
+
+Options:
+  --bucket, -b   The target storage bucket.                                      [string] [required]
+  --default, -d  Whether to set default access controls. Only valid when setting access controls on
+                 a bucket.                                                                 [boolean]
+  --file, -f     The target file.                                                           [string]
+  --help         Show help                                                                 [boolean]
+
+Examples:
+  node acl add user-bob@domain.com OWNER -b mybucket  Add OWNER access controls for
+                                                      "user-bob@domain.com" to "mybucket".
+  node acl add viewers-2256 WRITER -b mybucket -d     Add default WRITER access controls to
+                                                      "mybucket" for "viewers-2256".
+  node acl get editors-1234 -b mybucket               Get access controls for "editors-1234" in
+                                                      "mybucket".
+  node acl delete -b mybucket -f file.txt             Delete all access controls for all entities
+                                                      from "file.txt" in "mybucket".
+
+For more information, see https://cloud.google.com/storage/docs/access-control/create-manage-lists
+```
+
+[acl_docs]: https://cloud.google.com/storage/docs/access-control/create-manage-lists
+[acl_code]: acl.js
 
 ### Buckets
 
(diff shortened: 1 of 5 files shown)


jmdobry:
@ace-n Ready for review

storage/README.md
```diff
@@ -28,6 +29,41 @@ amount of data at any time.
...
 ## Samples
 
+### ACL (Access Control Lists)
+
+View the [documentation][acl_docs] or the [source code][acl_code].
+
+__Usage:__ `node acl --help`
+
+```
+Commands:
```
ace-n: Wishlist/Pipe dream: auto-generate these.

storage/acl.js
```diff
@@ -0,0 +1,209 @@
...
+  var bucket = storage.bucket(options.bucket);
+  // Reference the bucket's acl resource
+  var acl = bucket.acl;
+
+  if (options.file) {
+    // Optionally target a file's acl resource
+    acl = bucket.file(options.file).acl;
+  } else if (options.default) {
+    // Optionally get "default" access controls for the bucket
+    acl = acl.default;
```
ace-n: I'm assuming `bucket.acl` and `bucket.acl.default` are two different things. (If not, I may have an issue with this.)

storage/README.md
jmdobry: You mean auto-generate `README.md`? The usage information was generated, I just ran `node acl --help` and pasted the output into README.md. I used to have to write all that by hand, now most of README.md _is_ generated.

storage/acl.js
jmdobry: They definitely are two different things.

storage/system-test/acl.test.js
```diff
@@ -0,0 +1,237 @@
...
+      });
+    });
+
+    it('should get an entity\'s access controls for a bucket', function (done) {
+      program.getAccessControl({
+        bucket: bucketName,
+        entity: entity
+      }, function (err, aclObject) {
+        assert.ifError(err);
+        assert.deepEqual(aclObject, expected);
```
ace-n: I'm assuming `deepEqual` is the right kind of equality check to use here.

storage/README.md
ace-n: Generated in the "I don't have to even paste things in, I just run a script" sense.

storage/acl.js
jmdobry: I added links to in the source code relevant gcloud-node methods (with the "latest" version in the URL path).

storage/system-test/acl.test.js
jmdobry: Yeah, they definitely aren't strictly equal. `expected` is an object defined at the top of the file, `aclObject` came from the Storage API. The only other way to verify equality would be to be to recursively loop over properties, aka implement `assert.deepEqual`.

storage/README.md
jmdobry: I think there's negligible work to be saved by scripting the copy paste. I mean, it could be scripted, but we'd need to devise a `manifest.json` file or something that could give the script meta information about the samples in the directory. Maybe it's not such a bad idea...

storage/test/acl.test.js
```diff
@@ -0,0 +1,379 @@
...
+      var sample = getSample();
+      var callback = sinon.stub();
+      var options = {
+        bucket: bucketName
+      };
+      sample.mocks.bucket.acl.delete = sinon.stub().callsArgWith(1, error);
+
+      sample.program.deleteAccessControl(options, callback);
+
+      assert(callback.calledOnce, 'callback called once');
```
ace-n: What are the tradeoffs of this pattern vs. [defining a callback function and putting the tests in the there](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/bigquery/test/query.test.js#L110)? (And which should I be using in my tests?)

storage/test/acl.test.js
```diff
@@ -0,0 +1,379 @@
...
+      assert.equal(callback.firstCall.args[0].message, error.message, 'error has correct message');
+    });
+  });
+
+  describe('main', function () {
+    it('should call addAccessControl', function () {
+      var program = getSample().program;
+
+      sinon.stub(program, 'addAccessControl');
+      program.main(['add', 'owners-1234', 'OWNER', '-b', 'nodejs-docs-samples']);
```
ace-n: Should these values be defined as variables?

ace-n:
LGTM other than a few minor nits.

storage/test/acl.test.js
jmdobry: I would say the two main benefits are:

1) Use of `sinon.stub` as the callback allows for additional assertions about the code execution (it gathers more info about how the callback is used)
2) The test function now actually looks synchronous in addition to being synchronous.

storage/test/acl.test.js
jmdobry: nah

PR #182, opened 2016-08-18
All Storage samples are now up to standard.

jmdobry:
- [x] - All samples use `yargs` for CLI
- [x] - All samples follow sample template pattern
- [x] - All tests follow same conventions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,65 @@ accept your pull requests.
 1. Fork the desired repo, develop and test your code changes.
 1. Ensure that your code adheres to the existing style in the sample to which you are contributing. Refer to the
    [Google Cloud Platform Samples Style Guide](https://github.com/GoogleCloudPlatform/Template/wiki/style.html) for the
-   recommended coding standards for this organization.  You can run `npm run jshint` to match our JavaScript coding standards. 
+   recommended coding standards for this organization.  You can run `npm run jshint` to match our JavaScript coding standards.
 1. Ensure that your code has an appropriate set of unit tests which all pass.
 1. Submit a pull request!
+
+## Sample template
+
+```
+<LICENSE_HEADER>
+
+// [START all]
+// [START setup]
+// By default, the client will authenticate using the service account file
+// specified by the GOOGLE_APPLICATION_CREDENTIALS environment variable and use
+// the project specified by the GCLOUD_PROJECT environment variable. See
+// https://googlecloudplatform.github.io/gcloud-node/#/docs/google-cloud/latest/guides/authentication
+var <API_CLIENT> = require('@google-cloud/<API>');
+
+// Instantiate a <API> client
+var <CLIENT> = <API_CLIENT>();
+// [END setup]
+
+// [START <REGION_TAG_NAME>]
+/**
+ * <DESCRIPTION>
+ *
+ * @param {<TYPE>} <NAME> <DESCRIPTION>.
+ * @param {function} cb The callback function.
+ */
+function <METHOD_NAME> (<ARGUMENTS>, callback) {
+  // <SETUP>
+
+  // <RUN SAMPLE CODE>
+
+  // Inside callback: console.log(<MESSAGE>)
+}
+// [END <REGION_TAG_NAME>]
+
+// The command-line program
+var cli = require('yargs');
+
+var program = module.exports = {
+  <METHOD_NAME>: <METHOD_NAME>,
+  main: function (args) {
+    // Run the command-line program
+    cli.help().strict().parse(args).argv;
+  }
+};
+
+cli
+  .demand(1)
+  .command('<COMMAND> <ARGS>', '<DESCRIPTION>.', {}, function (options) {
+    program.<METHOD_NAME>(options, console.log);
+  })
+  .example('node $0 <COMMAND> <ARGS>', '<DESCRIPTION>.')
+  .wrap(100)
+  .recommendCommands()
+  .epilogue('For more information, see <DOCS_LINK>');
+
+if (module === require.main) {
+  program.main(process.argv.slice(2));
+}
+```
--- a/storage/encryption.js
+++ b/storage/encryption.js
@@ -55,35 +55,26 @@ function generateEncryptionKey () {
  * The file will be encrypted by Google Cloud Storage and only retrievable using
  * the provided encryption key.
  *
- * @param {string} bucketName The bucket where the file will be uploaded.
- * @param {string} srcFileName The local file to be uploaded.
- * @param {string} destFileName The name of the destination file.
- * @param {string} key The encryption key.
+ * @param {object} options Configuration options.
+ * @param {string} options.bucket The bucket where the file will be uploaded.
+ * @param {string} options.srcFile The local file to be uploaded.
+ * @param {string} options.destFile The name of the destination file.
+ * @param {string} options.key The encryption key.
  * @param {function} callback The callback function.
  */
-function uploadEncryptedFile (bucketName, srcFileName, destFileName, key, callback) {
-  if (!bucketName) {
-    return callback(new Error('"bucketName" is required!'));
-  } else if (!srcFileName) {
-    return callback(new Error('"srcFileName" is required!'));
-  } else if (!destFileName) {
-    return callback(new Error('"destFileName" is required!'));
-  } else if (!key) {
-    return callback(new Error('"key" is required!'));
-  }
-
-  var bucket = storage.bucket(bucketName);
-  var options = {
-    destination: destFileName,
-    encryptionKey: new Buffer(key, 'base64')
+function uploadEncryptedFile (options, callback) {
+  var bucket = storage.bucket(options.bucket);
+  var config = {
+    destination: options.destFile,
+    encryptionKey: new Buffer(options.key, 'base64')
   };
 
-  bucket.upload(srcFileName, options, function (err, file) {
+  bucket.upload(options.srcFile, config, function (err, file) {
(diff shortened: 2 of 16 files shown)


storage/encryption.js
```diff
@@ -55,35 +55,26 @@ function generateEncryptionKey () {
...
   };
 
-  bucket.upload(srcFileName, options, function (err, file) {
+  bucket.upload(options.srcFile, config, function (err, file) {
     if (err) {
       return callback(err);
     }
 
-    console.log('Uploaded encrypted file: %s', destFileName);
+    console.log('Uploaded gs://%s/%s', options.bucket, options.destFile);
```
ace-n: Should we make note of the uploaded file being encrypted?

storage/files.js
```diff
@@ -72,326 +68,281 @@ function listFiles (name, callback) {
...
 
-  var bucket = storage.bucket(name);
-  var options = {
-    prefix: prefix
+  var config = {
+    prefix: options.prefix
   };
-  if (delimiter && typeof delimiter === 'string') {
-    options.delimiter = delimiter;
+  if (options.delimiter && typeof options.delimiter === 'string') {
```
ace-n: Should we be checking types here? If `options.delimiter` isn't a string, I'm sure the user would want to be told that in a transparent way (e.g. an error message).

storage/files.js
```diff
@@ -72,326 +68,281 @@ function listFiles (name, callback) {
...
+      description: 'Filter files by a prefix.'
+    },
+    delimiter: {
+      alias: 'd',
+      requiresArg: true,
+      type: 'string',
+      description: 'Specify a delimiter.'
+    }
+  }, function (options) {
+    if (options.prefix !== undefined) {
```
ace-n: Does `if(options.prefix) { ... }` not work here?

storage/files.js
```diff
@@ -72,326 +68,281 @@ function listFiles (name, callback) {
...
+  .command('delete <bucket> <file>', 'Delete a file from a bucket.', {}, function (options) {
+    program.deleteFile(options, console.log);
+  })
+  .command('getMetadata <bucket> <file>', 'Get metadata for a file in a bucket.', {}, function (options) {
+    program.getMetadata(options, console.log);
+  })
+  .command('makePublic <bucket> <file>', 'Make a file public in a bucket.', {}, function (options) {
+    program.makePublic(options, console.log);
+  })
+  .command('move <bucket> <srcFile> <destFile>', 'Rename a file in a bucket.', {}, function (options) {
```
ace-n: "move" is a bit of a nebulous name for this command - could we call it "rename" within this file, or say something like "Move a file __within__ a bucket"?

storage/system-test/encryption.test.js
```diff
@@ -54,24 +54,38 @@ describe('storage:encryption', function () {
...
     it('should upload a file', function (done) {
-      program.uploadEncryptedFile(bucketName, filePath, fileName, key, function (err, file) {
+      var options = {
+        bucket: bucketName,
+        srcFile: filePath,
+        destFile: fileName,
+        key: key
+      };
+
+      program.uploadEncryptedFile(options, function (err, file) {
```
ace-n: What's the reasoning behind passing these in as an object?

storage/encryption.js
jmdobry: I think the method name (`uploadEncryptedFile`) and CLI (`node encryption upload ...`) cover it.

storage/files.js
jmdobry: Actually I should just remove this check.

storage/files.js
jmdobry: I was thinking `prefix` might be an empty string, but I think you're right. I'll change it.

storage/files.js
jmdobry: The Cloud Storage API (and `gsutil` CLI) mirror the S3 API and the also the linux `mv` command somewhat, e.g. `mv old_file.txt new_file.txt`. I think `move` is fine here. I'll clarify in the description though.

storage/system-test/encryption.test.js
jmdobry: All storage samples which before took more than one positional argument (besides the callback) have now collapsed them into a single `options` argument. This cleans up the function signature while also making it easier for the CLI to call the sample methods e.g. [`uploadEncryptedFile(options, callback)`](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/storage/storage/encryption.js#L58)

The JSDoc comment above each sample method very clearly documents the `options` argument and its properties.

CONTRIBUTING.md
```diff
@@ -27,6 +27,65 @@ accept your pull requests.
...
+// Instantiate a <API> client
+var <CLIENT> = <API_CLIENT>();
+// [END setup]
+
+// [START <REGION_TAG_NAME>]
+/**
+ * <DESCRIPTION>
+ *
+ * @param {<TYPE>} <NAME> <DESCRIPTION>.
+ * @param {function} cb The callback function.
```
jmdobry: `@param {<TYPE>} <NAME> <DESCRIPTION>.` can be repeated, which covers that.

storage/files.js
```diff
@@ -72,326 +68,281 @@ function listFiles (name, callback) {
...
-  file.makePublic(function (err, apiResponse) {
+  file.makePublic(function (err) {
     if (err) {
       return callback(err);
     }
 
-    console.log('Made %s public!', fileName);
-    return callback(null, apiResponse);
+    console.log('Made gs://%s/%s public!', options.bucket, options.file);
+    return callback(null);
```
ace-n: Is this null necessary? Can we just do `callback()`, or is that not clear enough?

storage/test/files.test.js
```diff
@@ -56,425 +59,461 @@ function getSample () {
...
+      sample.program.copyFile(options, callback);
+
+      assert(callback.calledOnce, 'callback called once');
+      assert.equal(callback.firstCall.args.length, 1, 'callback received 1 argument');
+      assert(callback.firstCall.args[0], 'callback received error');
+      assert.equal(callback.firstCall.args[0].message, error.message, 'error has correct message');
     });
   });
+
   describe('main', function () {
```
ace-n: I'm assuming we still want to keep this `main()` test around, even with `yargs` - or is that something we should discuss?

storage/test/transfer.test.js
```diff
@@ -191,78 +172,55 @@ describe('storage:transfer', function () {
         var value = 'DISABLED';
```
ace-n: What is the point of these variables if we're manipulating `options` and not them?

storage/test/transfer.test.js
```diff
@@ -271,12 +229,19 @@ describe('storage:transfer', function () {
       });
 
       it('should handle patch error', function () {
+        var field = 'status';
```
ace-n: Can these (field/value) be inserted directly into options?

storage/test/transfer.test.js
```diff
@@ -191,78 +172,55 @@ describe('storage:transfer', function () {
...
-
-        assert(callback.calledOnce, 'callback called once');
-        assert.equal(callback.firstCall.args.length, 1, 'callback received 1 argument');
-        assert(callback.firstCall.args[0], 'callback received error');
-        assert.equal(callback.firstCall.args[0].message, '"value" is required!', 'error has correct message');
+        assert(console.log.calledWith('Updated transfer job: %s', sample.mocks.transferJob.name));
       });
 
       it('should handle auth error', function () {
+        var field = 'status';
```
ace-n: Can these (field/value) be inserted directly into options?

storage/transfer.js
```diff
@@ -51,20 +51,17 @@ function auth (callback) {
 /**
  * Review the transfer operations associated with a transfer job.
  *
```
ace-n: extra line?

storage/system-test/encryption.test.js
ace-n: OK cool - thanks for clarifying!

storage/files.js
jmdobry: Not strictly necessary, but it's consistent with other samples in that `null` is passed as the first argument when there is no error. It's also part of the the [Node.js Errback Standard](http://fredkschott.com/post/2014/03/understanding-error-first-callbacks-in-node-js/#defining-an-error-first-callback:cb6a1cb7301215882ae9130ce938a4f7) to pass `null` as the first argument when there is no error.

storage/test/files.test.js
jmdobry: Yeah, there are a couple lines in each sample that get covered by having these `main` tests.

storage/test/transfer.test.js
jmdobry: You're right, fixed.

storage/test/transfer.test.js
jmdobry: Yes, fixed.

storage/test/transfer.test.js
jmdobry: Yes, fixed.

storage/transfer.js
jmdobry: As a practice, whenever I write JSDoc comments, I leave a line between the description and the tags. I think it helps "de-clutter" the comment block.

ace-n:
...is there a reason test coverage dropped so much, or is `codecov.io` just being flaky?

(If it's just `codecov.io`, LGTM.)

jmdobry:
There is a reason, I accidentally committed `.only` in one of the tests. Removing it. Will merge once tests pass.

ace-n:
LGTM.

PR #183, opened 2016-08-19
Bring Logging samples up to standard.

### LINKED ISSUE #175: Logging API samples
- [ ] refactor files to logs.js and sinks.js
- [ ] - logs.js
  - [ ] - CLI
  - [ ] - Unit tests
- [ ] - sinks.js
  - [ ] - CLI
  - [ ] - Unit tests

jmdobry:
Fixes #175 

- [x] - Sinks
- [x] - Logs


jmdobry:
@ace-n Ready for review, when you have a moment.

logging/logs.js
```diff
@@ -0,0 +1,180 @@
...
+ */
+function listLogEntries (options, callback) {
+  // Configuration to pass to Logging#getEntries
+  var config = {};
+
+  if (options.filter) {
+    config.filter = options.filter;
+  }
+  if (options.limit) {
+    config.pageSize = options.limit;
```
ace-n: Should we rename the `options` values to the names `config` uses? That way we can just pass `options` in as `config`.

logging/logs.js
```diff
@@ -0,0 +1,180 @@
...
+  })
+  .command('write <name> <resource> <entry>', 'Write a log entry.', {}, function (options) {
+    try {
+      options.resource = JSON.parse(options.resource);
+    } catch (err) {
+      console.error('"resource" must be a valid JSON string!');
+    }
+    try {
+      options.entry = JSON.parse(options.entry);
+      program.writeLogEntry(options, console.log);
```
ace-n: Should this statement be moved outside of the `try-catch` block?

logging/sinks.js
```diff
@@ -0,0 +1,217 @@
...
+ * @param {objects} options Configuration options.
+ * @param {string} options.name The name for new sink.
+ * @param {string} options.destination Destination for the new sink.
+ * @param {string} options.type The type of destination. Choices are:
+ *     "bucket", "dataset", or "topic".
+ * @param {function} callback The callback function.
+ */
+function createSink (options, callback) {
+  var sink = logging.sink(options.name);
+
```
ace-n: Nit: remove this line?

logging/sinks.js
```diff
@@ -0,0 +1,217 @@
...
+      options.metadata = JSON.parse(options.metadata);
+      program.updateSink(options, console.log);
+    } catch (err) {
+      console.error('"metadata" must be a valid JSON string!');
+    }
+  })
+  .command('delete <name>', 'Delete the specified sink.', {}, function (options) {
+    program.deleteSink(options.name, console.log);
+  })
+  .example('node $0 create my-sink my-bucket --type bucket', 'Create a new sink named "my-sink" that exports logs to a Cloud Storage bucket.')
```
ace-n: Can we infer `type` from the arguments, *or* do something like `--bucket my-bucket`/`--topic my-topic`?

That would save us a command-line argument...

logging/logs.js
jmdobry: Tried that. gcloud-node throws error if the options contains keys that aren't official options.

logging/logs.js
jmdobry: No, it should only be executed if the preceding line does not throw. Moving it out might cause it to get executed when it shouldn't.

logging/sinks.js
jmdobry: I thought about that, but to do it, I'd have to make three optional arguments. I don't know how to get `yargs` to require that one of the options be set. Ideas?

logging/sinks.js
jmdobry: Done.

logging/logs.js
ace-n: Is there an easy way to not throw the JSON string error if `writeLogEntry()` is the one throwing an exception (and not `JSON.parse()`)?

jmdobry: I made a change, I think it will work.

logging/sinks.js
ace-n: What about `node $0 create bucket my-sink my-bucket`, a la [this snippet](https://github.com/yargs/yargs/issues/225#issuecomment-128532719)?

logging/system-test/sinks.test.js
```diff
@@ -0,0 +1,129 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+var gcloud = require('google-cloud');
```
ace-n: 😢

logging/system-test/sinks.test.js
```diff
@@ -0,0 +1,129 @@
...
+      var options = {
+        name: sinkName,
+        destination: bucketName,
+        type: 'bucket'
+      };
+
+      program.createSink(options, function (err, sink) {
+        assert.ifError(err);
+        assert(sink, 'should have received the new sink');
+        assert.equal(sink.name, sinkName, 'should have received the new sink');
```
ace-n: Isn't this a more restrictive test than line 51?

My vote is for either a) removing line 51 entirely (because line 52 will still fail) or b) using a message along the lines of `sink is not null` for line 51's error.

logging/sinks.js
jmdobry: You mean make the type option a positional argument?

logging/system-test/sinks.test.js
```diff
@@ -0,0 +1,129 @@
...
+      });
+    });
+  });
+
+  describe('updateSink', function () {
+    it('should update metdata for a sink', function (done) {
+      var expected = {
+        name: sinkName,
+        destination: 'storage.googleapis.com/' + bucketName,
+        filter: 'severity > ALERT',
```
ace-n: Should this be a `var`?

logging/system-test/sinks.test.js
jmdobry: Sure, but would have to do this instead:

```js
var Logging = require('@google-cloud/logging');
var Storage = require('@google-cloud/storage');
var BigQuery = require('@google-cloud/bigquery');
var PubSub = require('@google-cloud/pubsub');
```

logging/system-test/sinks.test.js
jmdobry: The problem with a) is if `sink` is undefined then we won't get line 52's assertion error, we'll get `Cannot read property 'name' of undefined`, so we need to have the line 51 assertion. If you want the message changed, that's fine.

logging/sinks.js
ace-n: Yeah

logging/system-test/sinks.test.js
ace-n: Ah, makes sense.

logging/system-test/sinks.test.js
jmdobry: Done.

logging/test/sinks.test.js
```diff
@@ -0,0 +1,345 @@
...
+        type: 'dataset'
+      };
+
+      sample.program.createSink(options, callback);
+
+      assert(sample.mocks.sink.create.calledOnce, 'method called once');
+      assert.equal(sample.mocks.sink.create.firstCall.args.length, 2, 'method received 2 arguments');
+      assert.deepEqual(sample.mocks.sink.create.firstCall.args[0], {
+        destination: sample.mocks.dataset
+      }, 'add received options');
```
ace-n: s/add/create?

logging/sinks.js
jmdobry: I would've done that, but as far as I know I can't define `yargs` validation settings for a positional argument, only for an option. I looked through the docs and couldn't figure it outm so it's an option.

I'd rather not have to manually validate `type`.

logging/test/sinks.test.js
jmdobry: Fixed.

logging/test/sinks.test.js
```diff
@@ -0,0 +1,345 @@
...
+
+    it('should call updateSink', function () {
+      var program = getSample().program;
+
+      sinon.stub(program, 'updateSink');
+      program.main(['update', sinkName, '"{}"']);
+      assert(program.updateSink.calledOnce);
+    });
+
+    it('should call updateSink and validate metadata', function () {
```
ace-n: `should validate metadata before calling updateSink`?

logging/sinks.js
ace-n: Yeah.

ace-n:
LGTM, other than my comments.

logging/test/sinks.test.js
jmdobry: Fixed.

jmdobry:
Cool. Just pushed again, will merge when green.

jmdobry:
Thanks!

PR #185, opened 2016-08-21
Organize samples by product group.

jmdobry:
This updates the main README to group products the same way their grouped here: https://cloud.google.com/products/


jmdobry:
@ace-n What do you think?

ace-n:
LGTM :+1:

PR #192, opened 2016-08-29
Use real Error in tests.

jmdobry:
First of several "mini" PRs

### DIFF
--- a/bigquery/test/queries.test.js
+++ b/bigquery/test/queries.test.js
@@ -85,7 +85,7 @@ describe('bigquery:query', function () {
     });
 
     it('should handle error', function () {
-      var error = Error('syncQueryError');
+      var error = new Error('error');
       var example = getSample();
       example.mocks.bigquery.query = sinon.stub().callsArgWith(1, error);
       example.program.syncQuery(query, function (err, data) {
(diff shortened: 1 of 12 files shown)


bigquery/test/queries.test.js
```diff
@@ -85,7 +85,7 @@ describe('bigquery:query', function () {
     });
 
     it('should handle error', function () {
-      var error = Error('syncQueryError');
```
ace-n: Should this be a global variable? (Or do we have to create a new one within each test?)

logging/test/logs.test.js
```diff
@@ -74,7 +74,7 @@ describe('logging:entries', function () {
     });
 
     it('should handle error', function () {
-      var error = 'error';
```
ace-n: Global variable?

logging/test/sinks.test.js
```diff
@@ -149,7 +149,7 @@ describe('logging:sinks', function () {
     });
 
     it('should handle error', function () {
-      var error = 'error';
```
ace-n: Global variable?

ace-n:
Can some of these be converted to one-per-file global variables?

jmdobry:
How about doing something like that in a separate "Use per-file global variables in tests" PR?

ace-n:
Hmm - that could work, though IMHO the two are related enough to put in the same PR (since they're changing the same things).

jmdobry:
I was thinking there are other variables (not error vars) that could benefit from going global, hence a separate PR.

PR #196, opened 2016-08-30
First draft of streaming insert sample

ace-n:


bigquery/system-test/tables.test.js
```diff
@@ -131,8 +133,28 @@ describe('bigquery:tables', function () {
...
+    it('should insert rows into a table', function (done) {
+      var table = bigquery.dataset(options.dataset).table(options.table);
+      table.getRows({}, function (err, startRows) {
+        assert.equal(err, null);
+
+        program.insertRowsAsStream(options, function (err, insertErrors) {
+          assert.equal(err, null);
+          assert.deepEqual(insertErrors, [], 'no per-row insert errors occurred');
+
+          table.getRows({}, function (err, endRows) {
```
jmdobry: Wrap this in a timeout, say 2000ms

bigquery/tables.js
```diff
@@ -184,18 +184,40 @@ function exportTableToGCS (options, callback) {
...
+ * @param {string} options.table The ID of the table to insert rows into.
+ * @param {function} callback Callback function to receive query status.
+ */
+function insertRowsAsStream (options, callback) {
+  var table = bigquery.dataset(options.dataset).table(options.table);
+  table.insert(options.rows, function (err, insertErrors) {
+    if (err) {
+      return callback(err);
+    }
+    return callback(null, insertErrors);
```
jmdobry: We should have a `console.log` above this line like all the other samples, e.g.:

```js
if (err) {
  return callback(err);
}

console.log('Inserted %d row(s)!', options.rows.length);
return callback(null, insertErrors);
```

bigquery/tables.js
```diff
@@ -243,6 +265,26 @@ cli
   }, function (options) {
     program.exportTableToGCS(utils.pick(options, ['dataset', 'table', 'bucket', 'file', 'format', 'gzip']), utils.makeHandler());
   })
+  .command('insert <json_or_file> <dataset> <table>',
+    'Insert a JSON array (as a string or newline-delimited file) into a BigQuery table.', {},
+    function (options) {
+      var content = fs.readFileSync(options.json_or_file);
```
jmdobry: If the file cannot be found, this will throw an error, so it needs to be wrapped in a try/catch, and in your catch you would just do `content = options.json_or_file;`

ace-n: Will fix; adding a test for this too.

bigquery/tables.js
ace-n: Will fix.

bigquery/system-test/tables.test.js
ace-n: Will fix.

bigquery/tables.js
```diff
@@ -243,6 +266,29 @@ cli
   }, function (options) {
     program.exportTableToGCS(utils.pick(options, ['dataset', 'table', 'bucket', 'file', 'format', 'gzip']), utils.makeHandler());
   })
+  .command('insert <json_or_file> <dataset> <table>',
```
jmdobry: Let's have `dataset` and `table` be the first two args in order to match all the other commands.

bigquery/test/tables.test.js
```diff
@@ -380,5 +449,65 @@ describe('bigquery:tables', function () {
         gzip: false
       }]);
     });
+
+    describe('main:insert', function () {
```
jmdobry: Change `main:insert` to just `insert`. This `describe` clause is already nested under the `main` `describe` clause. Having `main:insert` is like calling this test `main main:insert`

jmdobry:
LGTM

ace-n:
@dpebot merge when green.

PR #201, opened 2016-09-02
Add browsing data sample

ace-n:


jmdobry:
This branch didn't get rebased properly, it overwrites a lot of the copyTable code.

bigquery/tables.js
```diff
@@ -252,7 +266,7 @@ cli
   .command('create <dataset> <table>', 'Create a new table in the specified dataset.', {}, function (options) {
     program.createTable(utils.pick(options, ['dataset', 'table']), utils.makeHandler());
   })
-  .command('list <dataset>', 'List tables in the specified dataset.', {}, function (options) {
+  .command('tables <dataset>', 'List tables in the specified dataset.', {}, function (options) {
```
jmdobry: This should stay `list` to be consistent with all the other sample CLIs.

ace-n: So the problem is that we can `list` both tables and rows - should I use something like `list_tables` and `list_rows` instead?

jmdobry: I was thinking of having `list` for listing tables, since the sample is `tables.js`, and `browse` for the rows, which matches the sample description "Browse a table.".

```
node tables list my_dataset
node tables browse my_dataset my_table
```

jmdobry: Or

```
node tables list my_dataset
node tables rows my_dataset my_table
or
node tables getRows my_dataset my_table
or
node tables get-rows my_dataset my_table
or
node tables listRows my_dataset my_table
or
node tables list-rows my_dataset my_table
```

I don't really mind what the CLI command is for running `listRows`, but I think the command for listing tables should be `list` for consistency.

ace-n: Point taken, will fix.

bigquery/system-test/tables.test.js
```diff
@@ -63,20 +63,17 @@ describe('bigquery:tables', function () {
...
+    bigquery.dataset(options.dataset).delete({ force: true }, function () {
+      // Delete files
+      storage.bucket(options.bucket).deleteFiles({ force: true }, function (err) {
+        if (err) {
+          return done(err);
+        }
+        // Delete bucket
+        setTimeout(function () {
+          storage.bucket(options.bucket).delete(done);
+        }, 2000);
```
jmdobry: Why these changes? Don't we still need to delete both `srcDataset` _and_ `destDataset`?

bigquery/system-test/tables.test.js
```diff
@@ -191,6 +188,20 @@ describe('bigquery:tables', function () {
     });
   });
 
+  describe('browseRows', function () {
+    it('should display rows in a table', function (done) {
+      program.browseRows(options.dataset, options.table, function (err, rows) {
+        assert.equal(err, null);
+        assert.notEqual(rows, null);
```
jmdobry: If we're asserting that `rows` is an array (asserting a type), then I think it's redundant to assert that `rows` is not `null` (asserting a type).

bigquery/tables.js
```diff
@@ -80,6 +80,19 @@ function listTables (options, callback) {
 }
 // [END list_tables]
 
+function browseRows (dataset, table, callback) {
+  var bigquery = BigQuery();
+  var tableObj = bigquery.dataset(dataset).table(table);
+  tableObj.getRows(function (err, rows) {
```
jmdobry: Add a space between `tableObj` and `table.getRows` for readability. Also, add the following above `tableObject.getRows`:

```
// See https://googlecloudplatform.github.io/google-cloud-node/#/docs/bigquery/latest/bigquery/table?method=getRows
```

jmdobry:
LGTM after comments are addressed.

ace-n:
@dpebot merge when green

PR #202, opened 2016-09-02
Simplify PubSub samples.

jmdobry:


pubsub/subscriptions.js
```diff
@@ -13,149 +13,128 @@
...
+  });
+}
+
+function getSubscriptionMetadata (subscriptionName, callback) {
+  var pubsub = PubSub();
+  var subscription = pubsub.subscription(subscriptionName);
+
+  // Get the metadata for the specified subscription
+  // See https://googlecloudplatform.github.io/google-cloud-node/#/docs/pubsub/latest/pubsub/subscription?method=getMetadata
+  subscription.getMetadata(function (err, metadata) {
```
ace-n: Include `apiResponse`?

pubsub/subscriptions.js
```diff
@@ -13,149 +13,128 @@
...
-    // Optionally find subscriptions for a specific topic
-    options.topic = pubsub.topic(topicName);
-  }
-  pubsub.getSubscriptions(options, function (err, subscriptions) {
+  var pubsub = PubSub();
+  var topic = pubsub.topic(topicName);
+
+  // List all subscriptions for the specified topic
+  // See https://googlecloudplatform.github.io/google-cloud-node/#/docs/pubsub/latest/pubsub/topic?method=getSubscriptions
+  topic.getSubscriptions(function (err, subscriptions) {
```
ace-n: `nextQuery` and/or `apiResponse`?

pubsub/subscriptions.js
```diff
@@ -13,149 +13,128 @@
...
+    return callback(null, subscriptions);
+  });
+}
+
+function listAllSubscriptions (callback) {
+  var pubsub = PubSub();
+
+  // List all subscriptions
+  // See https://googlecloudplatform.github.io/google-cloud-node/#/docs/pubsub/latest/pubsub?method=getSubscriptions
+  pubsub.getSubscriptions(function (err, subscriptions) {
```
ace-n: `nextQuery` and/or `apiResponse`?

pubsub/subscriptions.js
```diff
@@ -13,149 +13,128 @@
...
+
+  // List all subscriptions for the specified topic
+  // See https://googlecloudplatform.github.io/google-cloud-node/#/docs/pubsub/latest/pubsub/topic?method=getSubscriptions
+  topic.getSubscriptions(function (err, subscriptions) {
     if (err) {
       return callback(err);
     }
 
-    console.log('Found %d subscriptions!', subscriptions.length);
+    console.log('Found %d subscription(s)!', subscriptions.length);
```
ace-n: Should we differentiate this from the `console.log` in `listAllSubscriptions`? (I'm leaning towards No, but wanted to see what you thought.)

pubsub/subscriptions.js
jmdobry: There isn't actually an `apiResponse`, either the gcloud-node docs are wrong or there's a bug.

pubsub/subscriptions.js
jmdobry: I've been leaving these off because we're taking advantage of the auto-pagination

pubsub/subscriptions.js
jmdobry: There isn't actually an `apiResponse`, either the gcloud-node docs are wrong or there's a bug.

pubsub/subscriptions.js
jmdobry: I think having them the same is fine.

pubsub/system-test/iam.test.js
```diff
@@ -46,7 +47,7 @@ describe('pubsub:iam', function () {
       program.getTopicPolicy(topicName, function (err, policy) {
         assert.ifError(err);
         assert(policy);
-        assert(console.log.calledWith('Got topic policy:', policy));
+        assert(console.log.calledWith('Got policy for topic: %s', topicName));
```
ace-n: IIRC you said you'd clean up the asserts later. (They definitely need to be brought up to our new standard.)

pubsub/system-test/subscriptions.test.js
```diff
@@ -35,39 +36,56 @@ describe('pubsub:subscriptions', function () {
...
+        assert.equal(metadata.name, fullSubscriptionName);
+        assert(console.log.calledWith('Got metadata for subscription: %s', subscriptionName));
+        done();
+      });
+    });
+  });
+
   describe('listSubscriptions', function () {
     it('should list subscriptions', function (done) {
       program.listSubscriptions(topicName, function (err, subscriptions) {
```
ace-n: Can/should we `assert` that this is somehow different than `listAllSubscriptions`? e.g. run through all the results and make sure they have the right topicName?

pubsub/test/iam.test.js
```diff
@@ -20,18 +20,19 @@ var policyMock = 'policy';
 var permissionsMock = 'permissions';
 
 function getSample () {
+  var apiResponseMock = {};
   var subscriptionMock = {
     iam: {
       getPolicy: sinon.stub().callsArgWith(0, null, policyMock),
```
ace-n: Add `apiResponseMock` to this for consistency - or do we care if it's not used in anything that changed?

pubsub/test/iam.test.js
```diff
@@ -130,18 +114,10 @@ describe('pubsub:iam', function () {
 
       assert.ifError(callback.firstCall.args[0]);
       assert.equal(callback.firstCall.args[1], policyMock);
+      assert.strictEqual(callback.firstCall.args[2], sample.mocks.apiResponse);
```
ace-n: I'm assuming you'll clean these up later.

pubsub/test/subscriptions.test.js
```diff
@@ -73,110 +74,126 @@ describe('pubsub:subscriptions', function () {
...
-      assert(callback.firstCall.args[0]);
-      assert(callback.firstCall.args[0].message = '"subscriptionName" is required!');
-      assert.equal(callback.firstCall.args[1], undefined);
-    });
     it('should handle error', function () {
       var sample = getSample();
       var error = new Error('error');
       var callback = sinon.stub();
-      sample.mocks.subscription.delete.callsArgWith(0, error);
+      sample.mocks.subscription.getMetadata.callsArgWith(0, error);
```
ace-n: Go ahead and use `yields` here - if you're changing this anyway, you might as well correct it to the standard.

pubsub/system-test/iam.test.js
jmdobry: Cleaning up asserts in a separate PR.

pubsub/test/iam.test.js
jmdobry: `iam.getPolicy` does not return an `apiResponse`, contrary to what the gcloud-node docs say. Similar issue with `getMetadata`.

pubsub/test/iam.test.js
jmdobry: Yep.

pubsub/system-test/subscriptions.test.js
jmdobry: I think we're testing what needs to be tested with both of the list samples. Both hit the api, and we're verifying that both lists contain at least the subscription we created earlier. Depending on the project that running the system tests, there may or may not be more subscriptions in the list. We _could_ create another topic and subscription, just to verify that the second list contains both subscriptions. What do you think?

pubsub/test/subscriptions.test.js
jmdobry: Also have another in progress PR that fixes this.

I was trying to limit how much stuff you had to review. I fix orthogonal stuff, you point it out. I don't fix orthogonal stuff, you point it out. haha I can't escape 😜

pubsub/topics.js
```diff
@@ -13,115 +13,77 @@
...
- * List all topics in the current project.
- *
- * @param {Function} callback The callback function.
- */
+
 function listTopics (callback) {
+  var pubsub = PubSub();
+
+  // See https://googlecloudplatform.github.io/google-cloud-node/#/docs/pubsub/latest/pubsub?method=getTopics
   pubsub.getTopics(function (err, topics) {
```
ace-n: Include `nextQuery` and `apiResponse`?

ace-n: (Never mind.)

pubsub/test/subscriptions.test.js
ace-n: Ah, gotcha - historically, my way of doing things is "I'm not going out of my way to fix things - but if I'm changing something that needs to be fixed eventually, why not fix it right now (since doing it now isn't *as* orthogonal as fixing something I wouldn't otherwise change)?"

That said, I'm also fine with waiting and doing this in a separate PR. :smile:

pubsub/test/iam.test.js
ace-n: Should I open an issue for this?

pubsub/system-test/subscriptions.test.js
ace-n: I'd say create the second topic/subscription - it's just a few more API calls, and IMHO it's better to go overkill when writing tests then to fall short.

pubsub/test/iam.test.js
jmdobry: Actually, let me. I found like 6 methods so far (not just Pub/Sub ones) that don't return `apiResponse`. I'll make one big issue.

pubsub/subscriptions.js
ace-n: ...the gcloud-node docs are wrong. :stuck_out_tongue:

pubsub/system-test/subscriptions.test.js
jmdobry: Done.

jmdobry:
@ace-n Looks good?

ace-n:
Yep - all of my comments turned out to be non-issues, so LGTM.

(Update your branch, then merge when green.)

pubsub/iam.js
```diff
@@ -13,84 +13,48 @@
...
 
-    console.log('Got topic policy:', policy);
-    return callback(null, policy);
+    console.log('Got policy for topic: %s', topicName);
+    return callback(null, policy, apiResponse);
   });
 }
-// [END get_topic_policy]
 
-// [START get_subscription_policy]
```
jeffmendoza: @jmdobry These are needed for https://cloud.google.com/pubsub/docs/access_control

PR #217, opened 2016-09-26
Upgrade Speech samples.

jmdobry:


speech/recognize.js
```diff
@@ -11,106 +11,140 @@
...
-function getAuthClient (callback) {
-  // Acquire credentials
-  google.auth.getApplicationDefault(function (err, authClient) {
+// [START speech_sync_recognize]
+function syncRecognize (filename, callback) {
+  // Detect speech in the audio file, e.g. "./resources/audio.raw"
+  speech.recognize(filename, {
+    encoding: 'LINEAR16',
+    sampleRate: 16000
+  }, (err, results) => {
```
ace-n: Is there a reason we're using this callback pattern instead of the standard `function (a,b) { ... }` one? (Brevity?)

speech/recognize.js
```diff
@@ -11,106 +11,140 @@
...
+  .command('stream <filename>', 'Detects speech in an audio file by streaming it to the Speech API.', {}, function (options) {
+    program.streamingRecognize(options.filename, utils.makeHandler(false));
+  })
+  .command('listen', 'Detects speech in a microphone input stream.', {}, function () {
+    program.streamingMicRecognize();
+  })
+  .example('node $0 sync ./resources/audio.raw', 'Detects speech in "./resources/audio.raw".')
+  .example('node $0 async ./resources/audio.raw', 'Creates a job to detect speech in "./resources/audio.raw", and waits for the job to complete.')
+  .example('node $0 stream ./resources/audio.raw', 'Detects speech in "./resources/audio.raw" by streaming it to the Speech API.')
+  .example('node $0 listen ./resources/audio.raw', 'Detects speech in a microphone input stream.')
```
ace-n: Remove filepath?

speech/recognize.js
jmdobry: With this PR, I decided to start using the ES2015 features of JavaScript that are supported in the latest 4.5.0 LTS release of Node.js. See https://github.com/nodejs/LTS and http://node.green/

With ES2015, one would use arrow functions (`(a, b) => { ...`) as much as possible. You'll notice in this PR I also started using the `const` keyword. I also added an `engines` field to `package.json`.

speech/recognize.js
jmdobry: Done

PR #216, opened 2016-09-26
Upgrade Pub/Sub samples.

jmdobry:
* Switch to region tags that match other languages' samples
* Simplify tests (cut amount of test code in half)
* Achieved 100% code coverage (as opposed to 95.5% or whatever we've been limited to previously
* Added sample for creating a Push Subscription
* Begin to use ES2015 features available in the Node.js 4.5.0 LTS release

Compare to:

* GoogleCloudPlatform/ruby-docs-samples#63
* GoogleCloudPlatform/python-docs-samples#543

PR #218, opened 2016-09-27
Adds 6 quickstart examples

jmdobry:


jmdobry:
@ace-n Should be an easy review (you wrote the tests after all, and they LGTM).

ace-n:
Will review once #216 is merged. (That removes the flaky Datastore test, which is also failing here.)

PR #222, opened 2016-09-29
Fix pagination

### LINKED ISSUE #115: gcloud-node datastore no longer supports "autoPaginate" option
Need to fix "cursor_paging" sample

ace-n:
As title; fixes #115


jmdobry:
I believe there's a test that needs to be un-commented or un-skipped to make this these changes work?

ace-n:
(Good catch - fixed the test.)

jmdobry:
LGTM

PR #236, opened 2016-10-13
Switch --bucket to --stage-bucket

dlorenc:
gcloud 130.0.0 gives:
`WARNING: --bucket flag is deprecated. Use --stage-bucket instead.` when using --bucket.


jmdobry:
Fixed by #245

PR #256, opened 2016-11-16
Fix multer options

kacepe:
There is no "inMemory" options for multer.

### DIFF
--- a/appengine/storage/app.js
+++ b/appengine/storage/app.js
@@ -18,6 +18,7 @@
 
 const format = require('util').format;
 const express = require('express');
+const Multer = require('multer');
 
 // By default, the client will authenticate using the service account file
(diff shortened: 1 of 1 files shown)


appengine/storage/app.js
```diff
@@ -35,8 +36,8 @@ app.set('view engine', 'pug');
 // [START config]
 // Multer is required to process file uploads and make them available via
 // req.files.
-const multer = require('multer')({
-  inMemory: true,
+const multer = Multer({
+  storage: Multer.MemoryStorage,
```
jmdobry: From multer's readme it seems like this should be `storage: Multer.memoryStorage()` right?

kacepe: Yeah it's. Sorry my bad. Changed

jmdobry:
Thanks!

PR #274, opened 2016-12-14
Add Premium translation sample.

jmdobry:


translate/translate.js
```diff
@@ -107,6 +107,40 @@ function translateText (input, target) {
...
+    model: model
+  };
+
+  // Translates the text into the target language. "input" can be a string for
+  // translating a single piece of text, or an array of strings for translating
+  // multiple texts.
+  return translate.translate(input, options)
+    .then((results) => {
+      let translations = results[0];
+      translations = Array.isArray(translations) ? translations : [translations];
```
ace-n: Nit: not that we should use it, but [there's a library for this](https://www.npmjs.com/package/arrify). :stuck_out_tongue:

jmdobry: Personally, I feel that using a library for something like this is taking the Node.js "module all the things" too far...

PR #289, opened 2017-01-06
Switch from Mocha to Ava for faster tests

jmdobry:


jmdobry:
@ace-n Looks good to you?

jmdobry:
I think we're good, we already discussed my changes. Though I'll experiment with different concurrencies before I merge.

PR #290, opened 2017-01-07
Add yarn.lock files, use Yarn during CI build

jmdobry:


jmdobry:
@ace-n This should be good to go, CI build is much faster now.

PR #294, opened 2017-01-09
Vision snippets.

jmdobry:


vision/detect.js
```diff
@@ -0,0 +1,457 @@
...
+
+const Vision = require('@google-cloud/vision');
+const Storage = require('@google-cloud/storage');
+
+// [START vision_face_detection]
+function detectFaces (fileName) {
+  // Instantiates a client
+  const vision = Vision();
+
+  // The local image file on which to perform face detection
```
gguuss: Any reason to leave this?

vision/detect.js
```diff
@@ -0,0 +1,457 @@
...
+  // fileName = '/path/to/image.jpg';
+
+  // Performs face detection on the local file
+  return vision.detectFaces(fileName)
+    .then((results) => {
+      const faces = results[0];
```
gguuss: Nit. Prefer [four spaces on breaks when wrapping preceding line](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting):

```
  return vision.detectFaces(fileName)
      .then((results) => { 
        const faces = results[0];
```

That said, your aforementioned two-space style persists throughout and is consistent with anonymous function spacing so you're free to ignore as this style is consistent elsewhere and would require more changes to remain so, which is more important.

vision/system-test/detect.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+const bucketName = `nodejs-docs-samples-test-${uuid.v4()}`;
+const cmd = `node detect.js`;
+const cwd = path.join(__dirname, `..`);
+const files = [
+  `face_no_surprise.jpg`,
+  `landmark.jpg`,
+  `logos.png`,
+  `text.jpg`,
+  `wakeupcat.jpg`
```
gguuss: Are you using `water.jpg` for any tests?

vision/detect.js
jmdobry: Removed.

appengine/system-test/all.test.js
```diff
@@ -129,13 +129,13 @@ var sampleTests = [
...
     msg: 'Logged'
   },
-  {
-    dir: 'appengine/loopback',
-    cmd: 'node',
-    args: ['server/server.js'],
-    msg: 'started',
-    code: 304
-  },
+  // {
```
jmdobry: Added a comment.

vision/system-test/detect.test.js
jmdobry: One of the other samples refers to it.

vision/detect.js
jmdobry: We'll stick with 2 spaces.

PR #297, opened 2017-01-12
Update JS quickstart to remove misinformation about global

hjfreyer:


hjfreyer:
Oops. This was already fixed in master. Should we drop this branch?

jmdobry:
Yeah

PR #304, opened 2017-01-27
Add KMS samples.

jmdobry:


kms/keys.js
```diff
@@ -0,0 +1,1467 @@
...
+      if (index === -1) {
+        policy.bindings.push(binding);
+      }
+      if (!binding.members.includes(member)) {
+        binding.members.push(member);
+      }
+
+      request = {
+        // This will be a path parameter in the request URL
+        resource_: `projects/${projectId}/locations/${location}/keyRings/${keyRingName}`,
```
ace-n: Wishful thinking: these parameter names (`resource_` and `resource`) are annoyingly similar. (Not that we can really change that... 🙁 )

kms/keys.js
```diff
@@ -0,0 +1,1467 @@
...
+        console.log(err);
+        return;
+      }
+
+      const cryptoKeyVersions = result.cryptoKeyVersions || [];
+
+      if (cryptoKeyVersions.length) {
+        cryptoKeyVersions.forEach((version) => {
+          console.log(`${version.name}:`);
+          console.log(`  Created: ${new Date(version.createTime)}`);
```
ace-n: Pedantic nit: should we use tab characters (or other special characters) that are easier to count instead of spaces?

kms/keys.js
jmdobry: Yes, I was annoyed too

kms/keys.js
```diff
@@ -0,0 +1,1467 @@
...
+
+    let request = {
+      // This will be a path parameter in the request URL
+      resource_: `projects/${projectId}/locations/${location}/keyRings/${keyRingName}`
+    };
+
+    // Gets the IAM policy of a key ring
+    cloudkms.projects.locations.keyRings.getIamPolicy(request, (err, policy) => {
+      if (err) {
+        console.log(err);
```
jmdobry: The problem with `console.error` is it writes to `stderr`, and the tests had a problem with that.

kms/keys.js
jmdobry: Maybe, but I didn't like how huge the tab characters were when printed to my console. 2 spaces was more reliable.

PR #307, opened 2017-02-01
Update Speech samples.

jmdobry:


speech/recognize.js
```diff
@@ -23,143 +23,275 @@
...
+  // Detects speech in the audio file
+  speech.recognize(filename, request)
     .then((results) => {
       const transcription = results[0];
+
+      console.log(`Transcription: ${transcription}`);
+    });
+  // [END speech_sync_recognize]
+}
+
```
gguuss: FYI. Missing JSDoc for the functions - could be nice to have `@param` blocks with suggested values, e.g. '16000' for samplerate, 'LINEAR16' for encoding. Feel free to ignore.

jmdobry: Yeah we don't really do those. If we were writing library code then we certainly would add them.

PR #313, opened 2017-02-14
Spanner samples

ace-n:
(FYI these were already reviewed by @jmdobry. @jonparrott, these should be good to merge - but feel free to skim them + verify.)


.gitignore
```diff
@@ -1,3 +1,6 @@
+.DS_Store
```
jonparrott: This goes in your global gitignore, not in the repositories.

PR #314, opened 2017-02-15
Remove bogus .value after calling toJSON()

vrana:


ace-n:
Hi @vrana and thank you for the PR.

AFAICT, those `value` properties are necessary - the value of the `json.SingerId` object for instance is `{ [Number: 1] value: '1' }`. Since we only want to print `1`, we have to get the `value` property of that object.

The reason why we can't have `json.SingerId === 1` is because Node doesn't support 64-bit integers by default (i.e. `INT64s`), while Spanner does. As a workaround, the Spanner client libraries use strings to store the value of an `INT64`. This object is necessary to differentiate between `1` (an `INT64`) and `"1"` (a `String`).

ace-n:
@vrana Also, I tried running this on my machine and noticed that the tests failed. In the future, can you make sure to run the tests before submitting a PR?

(You're welcome to make changes to the tests if needed - the main thing for us is that they pass.)

vrana:
This pull-request is bogus, sorry about it.

PR #317, opened 2017-02-22
Updated readWriteTransaction to fix bug

ryanmats:
Documentation at https://cloud.google.com/spanner/docs/transactions#rw_transaction_example says that second budget should be at least $300,000 for the transaction to proceed. I'm fixing this for Python / Node. Addresses bug #35360945.

### DIFF
--- a/spanner/transaction.js
+++ b/spanner/transaction.js
@@ -100,6 +100,7 @@ function readWriteTransaction (instanceId, databaseId) {
   // at a specific point in time
   let transaction, firstBudget, secondBudget;
   const transferAmount = 200000;
+  const minimumAmountToTransfer = 300000;
 
   database.runTransaction()
     .then((results) => {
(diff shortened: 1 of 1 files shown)


spanner/transaction.js
```diff
@@ -126,8 +126,8 @@ function readWriteTransaction (instanceId, databaseId) {
           console.log(`The second album's marketing budget: ${secondBudget}`);
 
           // Makes sure the second album's budget is sufficient
-          if (secondBudget < transferAmount) {
-            throw new Error(`The second album's budget (${secondBudget}) is less than the transfer amount (${transferAmount}).`);
```
ace-n: Nit: can we make the `300000` a constant, and put its declaration next to `transferAmount`?

(I prefer having constants defined in one place if possible rather than repeating literal values.)

ryanmats: done

PR #324, opened 2017-03-14
Add key parameter to translate quickstart

fabulousduck:
added key param to quickstart.js
see commit for more info.

### DIFF
--- a/translate/quickstart.js
+++ b/translate/quickstart.js
@@ -21,10 +21,12 @@ const Translate = require('@google-cloud/translate');
 
 // Your Google Cloud Platform project ID
 const projectId = 'YOUR_PROJECT_ID';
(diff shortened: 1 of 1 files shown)


fabulousduck:
I signed it!

jmdobry:
Thanks for this! However, the Google Cloud Translation API no longer requires an API key, and can authenticate using a Service Account or Application Default Credentials. You can see that none of the quickstart samples on https://cloud.google.com/translate/docs/reference/libraries use an API key, but rely on Application Default Credentials ([more info about auth](https://cloud.google.com/docs/authentication#getting_credentials_for_server-centric_flow)).

We recommend using Service Accounts or Application Default Credentials, so that's what we show in the quickstart sample.

PR #328, opened 2017-03-29
chore: update docker in containerengine example

MylesBorins:
Currently we are pegged to Node:4. This is an older LTS image and
clocks in at 655 MB.

This PR updates us to Node:6-alpine. This will use a more modern
LTS branch, and also switches to the Alpine container which clocks
in at 54 MB.


MylesBorins:
indeed tested locally and everything works A-OK

ace-n:
@dpebot merge when green

PR #331, opened 2017-04-03
Removed commented lines for instanceId and databaseId

vkedia:
These are now being passed in as method arguments.


ace-n:
The samples themselves are composed of the content in between the region tags - users on `cloud.google.com` will not see the method arguments.

(Closing this PR for now; feel free to comment/reopen if I missed something or there is something else to discuss.)

vkedia:
Ok. That makes sense.

PR #335, opened 2017-04-04
Travis: fix failing tests + update dependencies

ace-n:


ace-n:
This doesn't fix everything, and leaves a number of TODOs:
- Monitoring tests are flaky (and need fixing)
- DLP tests should be parallelized
- The `node-mailer` module only works with `v6+` of Node - we need a different module, since we support `v4`.
- It seems like we're currently skipping a bunch of App Engine tests. IMHO, we need to either fix those tests, or remove the associated samples.

However, it does fix some more egregious test failures.

ace-n:
(Fixing a few lint issues, will delete this comment when I'm done.)

monitoring/system-test/quickstart.test.js
```diff
@@ -23,7 +23,8 @@ const client = proxyquire(`@google-cloud/monitoring`, {}).v3().metricServiceClie
 test.beforeEach(stubConsole);
 test.afterEach.always(restoreConsole);
 
-test.cb(`should list time series`, (t) => {
+// TODO(anassri): Fix this test occasionally omitting time series data
```
jonparrott: Don't put TODOs in code, you have *two* world-class bug tracking tools at your disposal.

PR #347, opened 2017-04-12
Update openApi config to use "security" instead of "x-security".

liminw:


ace-n:
Can you provide some context on why this is necessary? (I didn't find anything obvious on the internet - though it's possible I wasn't looking hard enough.)

liminw:
We recently deprecated x-security, and replaced it with security. If users continue using x-security, they will see a warning. openApi config in all endpoints samples need to be updated.

PR #351, opened 2017-04-13
needs to new koa const for koa v 2 to work

emcwilliams:
I was just reading through your tutorial in the Google App Engine kao intro and ran into Koa v2 requiring the "new" statement.  Just wanted to give you the info to help out on the documentation.  This doesn't look like the same code as in the example but maybe the message will reach you and I can sleep tonight knowing I helped fix something today :D

### DIFF
--- a/appengine/koa/app.js
+++ b/appengine/koa/app.js
@@ -17,6 +17,7 @@
 
 // [START server]
 const app = require('koa')();
+const app = new koa();
 
 app.use(function * () {
   this.body = 'Hello World! Koa.js on Google App Engine.';


appengine/koa/app.js
```diff
@@ -17,6 +17,7 @@
 
 // [START server]
 const app = require('koa')();
+const app = new koa();
```
jmdobry: Is this supposed to replace the line above it?

fhinkel:
@emcwilliams Thanks for the PR! Do you want to make the requested changes?

PR #358, opened 2017-04-24
Bring BigQuery samples up to standard.

jmdobry:


bigquery/datasets.js
```diff
@@ -15,123 +15,117 @@
...
 'use strict';
 
-const BigQuery = require('@google-cloud/bigquery');
+function createDataset (datasetId, projectId) {
+  // [START bigquery_create_dataset]
+  // Imports the Google Cloud client library
+  const BigQuery = require('@google-cloud/bigquery');
+
+  // The project ID to use, e.g. "your-project-id"
+  // const projectId = "your-project-id";
```
ace-n: Standards question: should we consistently specify project IDs in samples, or not?

(I don't think we enforce this particular standard yet, but I could be wrong...)

bigquery/datasets.js
```diff
@@ -15,123 +15,117 @@
...
-  const bigquery = BigQuery();
+  const bigquery = BigQuery({
+    projectId: projectId
+  });
+
+  // The ID of the dataset to delete, e.g. "my_new_dataset"
+  // const datasetId = "my_new_dataset";
 
-  // References an existing dataset, e.g. "my_dataset"
+  // Creates a references to the existing dataset
```
ace-n: s/reference**s**/reference/

bigquery/datasets.js
ace-n: If we do specify them, we should also test to make sure they are (properly) configurable - i.e. test the sample without `GCLOUD_PROJECT` being set (or being set to the empty string).

bigquery/tables.js
```diff
@@ -15,308 +15,412 @@
...
+  bigquery
+    .dataset(datasetId)
+    .table(tableId)
+    .insert(rows)
     .then((insertErrors) => {
       console.log('Inserted:');
       rows.forEach((row) => console.log(row));
-      return insertErrors;
+
+      if (insertErrors && insertErrors.length) {
```
ace-n: Why are we checking `insertErrors.length` - if we want this to be `insertErrors.length > 0`, can we be explicit about that?

bigquery/tables.js
```diff
@@ -15,308 +15,412 @@
...
-    cli.help().strict().parse(args).argv; // eslint-disable-line
-  }
-};
-
-cli
+require(`yargs`) // eslint-disable-line
   .demand(1)
   .command(`create <datasetId> <tableId> <schema> [projectId]`, `Creates a new table.`, {}, (opts) => {
-    program.createTable(opts.datasetId, opts.tableId, opts.schema, opts.projectId || process.env.GCLOUD_PROJECT);
+    createTable(opts.datasetId, opts.tableId, opts.schema, opts.projectId || process.env.GCLOUD_PROJECT);
```
ace-n: Nit: is it possible to make a global `projectId` option with yargs that uses `process.env.GCLOUD_PROJECT` as a default? Might be fewer changes/less repetition...

bigquery/datasets.js
jmdobry: Configuring the project ID is generally something we don't do for other products' samples. With BigQuery however, it is more common to be explicit about project ID because of public datasets. Many of the BigQuery samples already allowed configurable project ID via the CLI, I just made all the BigQuery samples consistent.

bigquery/datasets.js
jmdobry: Fixed.

bigquery/tables.js
jmdobry: Checking the truthiness of `length` is a common way to answer the question of "does this array have any items", but I agree about being explicit. Fixed.

bigquery/tables.js
jmdobry: Done.

bigquery/datasets.js
ace-n: 👍

PR #359, opened 2017-04-25
Add IAM samples

ace-n:
@frankyn a few (nitpicky) questions for you:
- Should we be calling our sample methods/region tags [view/add/remove]BucketIam**Role** instead of [view/add/remove]BucketIam**Members**?
- Do we need to explicitly specify project ID? The [canonical Ruby IAM sample](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/storage/iam.rb#L33) specifies it, but only some of the Node.js storage ones do. (Perhaps @jmdobry can chime in too?)


storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+function addBucketIamMember (bucketName, roleName, members) {
+  // [START add_bucket_iam_member]
+  // Your Google Cloud Storage bucket name
+  // const bucketName = "my-bucket";
+
+  // The bucket-level IAM role to grant
+  // const roleName = "roles/storage.objectViewer";
+
+  // The list of IAM members to grant the role to
+    // const members = ['user:jdoe@example.com', 'group:admins@example.com'];
```
jmdobry: there's some extra indentation on this line

storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function viewBucketIamMembers (bucketName) {
+  // [START view_bucket_iam_members]
+
+  // Your Google Cloud Storage bucket name
```
jmdobry: Variable definitions (everything really) should be below any import statements.

storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function viewBucketIamMembers (bucketName) {
+  // [START view_bucket_iam_members]
+
```
jmdobry: there's an extra blank line here

storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+    .catch((err) => {
+      console.error('Error:', err);
+    });
+  // [END view_bucket_iam_members]
+}
+
+function addBucketIamMember (bucketName, roleName, members) {
+  // [START add_bucket_iam_member]
+  // Your Google Cloud Storage bucket name
+  // const bucketName = "my-bucket";
```
jmdobry: Variable definitions (everything really) should be below any import statements

storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+    .catch((err) => {
+      console.error('Error:', err);
+    });
+  // [END add_bucket_iam_member]
+}
+
+function removeBucketIamMember (bucketName, roleName, members) {
+  // [START remove_bucket_iam_member]
+  // Your Google Cloud Storage bucket name
+  // const bucketName = "my-bucket";
```
jmdobry: Variable definitions (everything really) should be below any import statements

storage/iam.js
```diff
@@ -0,0 +1,190 @@
...
+  .example(`node $0 view-members "my-bucket"`)
+  .example(`node $0 add-members "my-bucket" "storage.objectViewer" "user:jdoe@example.com" "group:admins@example.com"`)
+  .example(`node $0 remove-members "my-bucket" "storage.objectViewer" "user:jdoe@example.com" "group:admins@example.com"`)
+  .wrap(120)
+  .recommendCommands()
+  .epilogue(`For more information, see https://cloud.google.com/iam/docs/overview and https://cloud.google.com/storage/docs`)
+  .help();
+
+if (module === require.main) {
+  cli.help().strict().argv; // eslint-disable-line
```
jmdobry: Maybe you're right, perhaps we should make this the standard (a "main" function).

Change to:

```js
  .epilogue(`For more information, see https://cloud.google.com/iam/docs/overview and https://cloud.google.com/storage/docs`)
  .help()
  .strict();

if (module === require.main) {
  cli.parse(process.argv.slice(2));
}
```

frankyn:
> Should we be calling our sample methods/region tags [view/add/remove]BucketIamRole instead of [view/add/remove]BucketIamMembers?

Keep BucketIam**Members** as it makes more sense with the documentation that's being released. I'll send you additional information through chat.

> Do we need to explicitly specify project ID? The canonical Ruby IAM sample specifies it, but only some of the Node.js storage ones do. (Perhaps @jmdobry can chime in too?)

I did it for Ruby because it's used more often. This falls into a per-language bias.

PR #360, opened 2017-04-25
Split analyze.js into two separate samples

ace-n:
Per @sgreenberg's instructions, we want to have separate samples for `v1beta1` and `v1beta2` of the NL API.


ace-n:
Two main ones:
- `v1beta2` supports more languages than `v1beta1` (i.e. German and a few others). This is covered in the tests
- `v1beta2` introduces the `analyzeEntitySentiment` method.

PR #362, opened 2017-04-25
Bring Speech API samples up to standard.

jmdobry:


speech/recognize.js
```diff
@@ -216,7 +216,8 @@ function streamingRecognize (filename, encoding, sampleRateHertz, languageCode)
       encoding: encoding,
       sampleRateHertz: sampleRateHertz,
       languageCode: languageCode
-    }
+    },
+    interimResults: false // Try setting to true to get interim results
```
ace-n: I assume these are meant for the user? (If so, we should clarify that.)

jmdobry: what would you change it to?

ace-n: // If you want interim results, set this to true

jmdobry: Done

PR #371, opened 2017-05-01
Add Spanner + Cloud Functions sample

ace-n:
FYI @WalterHub - this is the sample you wanted me to add.

@jmdobry: the sample code itself has (presumably) already been reviewed (given that it was pasted directly into a docs page). The test code, however, is new and _has **not_** been reviewed yet.

### DIFF
--- a/functions/spanner/index.js
+++ b/functions/spanner/index.js
@@ -39,7 +39,7 @@ const databaseId = 'my-database';
  * @param {Object} req Cloud Function request context.
  * @param {Object} res Cloud Function response context.
  */
-exports.get = function helloGET (req, res) {
+exports.get = (req, res) => {
   // Gets a reference to a Cloud Spanner instance and database
   const instance = spanner.instance(instanceId);
   const database = instance.database(databaseId);
--- a/functions/spanner/index.js
+++ b/functions/spanner/index.js
@@ -19,13 +19,8 @@
 // Imports the Google Cloud client library
 const Spanner = require('@google-cloud/spanner');
 
-// Your Google Cloud Platform project ID
-const projectId = 'YOUR_PROJECT_ID';
-
 // Instantiates a client
-const spanner = Spanner({
-  projectId: projectId
-});
+const spanner = Spanner();
 
 // Your Cloud Spanner instance ID
 const instanceId = 'my-instance';
@@ -53,9 +48,11 @@ exports.get = (req, res) => {
   database.run(query)
     .then((results) => {
       const rows = results[0];
-      var data = [];
-      rows.forEach((row) => data.push(row.toJSON()));
-      res.send(data);
+      res.send(rows.map((row) => row.toJSON()));
+    })
+    .catch((err) => {
+      res.status(500);
+      res.send(`Error querying Spanner: ${err}`);
     });
 };
 // [END spanner_functions_quickstart]
--- a/functions/spanner/index.js
+++ b/functions/spanner/index.js
@@ -58,7 +58,8 @@ exports.get = (req, res) => {
     .catch((err) => {
       res
         .status(500)
-        .send(`Error querying Spanner: ${err}`);
+        .send(`Error querying Spanner: ${err}`)
+        .end();
     });
 };
 // [END spanner_functions_quickstart]
--- a/functions/spanner/index.js
+++ b/functions/spanner/index.js
@@ -47,12 +47,18 @@ exports.get = (req, res) => {
   // Execute the query
   database.run(query)
     .then((results) => {
-      const rows = results[0];
-      res.send(rows.map((row) => row.toJSON()));
+      const rows = results[0].map((row) => row.toJSON());
+      rows.forEach((row) => {
+        res.write(`SingerId: ${row.SingerId.value}, AlbumId: ${row.AlbumId.value}, AlbumTitle: ${row.AlbumTitle}</br>`);
+      });
+      res
+        .status(200)
+        .end();
     })
     .catch((err) => {
-      res.status(500);
-      res.send(`Error querying Spanner: ${err}`);
+      res
+        .status(500)
+        .send(`Error querying Spanner: ${err}`);
     });
 };
 // [END spanner_functions_quickstart]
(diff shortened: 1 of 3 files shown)


functions/spanner/index.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+'use strict';
+
+// [START spanner_functions_quickstart]
+// Imports the Google Cloud client library
+const Spanner = require('@google-cloud/spanner');
+
+// Your Google Cloud Platform project ID
+const projectId = 'YOUR_PROJECT_ID';
```
jmdobry: Users should not be hard-coding `projectId` variables inside Cloud Functions (the environment provides the project ID. Remove this line, and don't pass project ID into the client library constructor.

functions/spanner/index.js
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  // Execute the query
+  database.run(query)
+    .then((results) => {
+      const rows = results[0];
+      var data = [];
+      rows.forEach((row) => data.push(row.toJSON()));
+      res.send(data);
+    });
```
jmdobry: Missing error handling.

functions/spanner/test/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  };
+}
+
+test(`get: Gets albums`, async (t) => {
+  const sample = getSample();
+  const mocks = sample.mocks;
+
+  const err = await sample.program.get(mocks.req, mocks.res);
```
jmdobry: There's no point using async/await in this test function, as `program.get` does not return a promise (and shouldn't).

functions/spanner/test/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+  t.falsy(err, null);
```
jmdobry: Remove this line

functions/spanner/test/index.test.js
jmdobry: so change line to:

```js
sample.program.get(mocks.req, mocks.res);
```

functions/spanner/test/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+      results: resultsMock,
+      res: {
+        status: sinon.stub().returnsThis(),
+        send: sinon.stub().returnsThis()
+      }
+    }
+  };
+}
+
+test(`get: Gets albums`, async (t) => {
```
jmdobry: Remove `async`

functions/spanner/index.js
```diff
@@ -0,0 +1,61 @@
...
+    sql: 'SELECT * FROM Albums'
+  };
+
+  // Execute the query
+  database.run(query)
+    .then((results) => {
+      const rows = results[0];
+      var data = [];
+      rows.forEach((row) => data.push(row.toJSON()));
+      res.send(data);
```
jmdobry: Change to `res.send(rows.map((row) => row.toJSON()));` and delete the redundant lines

functions/spanner/test/index.test.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  };
+}
+
+test(`get: Gets albums`, (t) => {
+  const sample = getSample();
+  const mocks = sample.mocks;
+
+  const err = sample.program.get(mocks.req, mocks.res);
```
jmdobry: Fails lint, plus, `program.get` does not have a return value.

functions/spanner/index.js
```diff
@@ -0,0 +1,58 @@
...
+  // The query to execute
+  const query = {
+    sql: 'SELECT * FROM Albums'
+  };
+
+  // Execute the query
+  database.run(query)
+    .then((results) => {
+      const rows = results[0];
+      res.send(rows.map((row) => row.toJSON()));
```
jmdobry: Let's just be explicit:

```js
res
  .status(200)
  .send(rows.map((row) => row.toJSON()))
  .end();
```

functions/spanner/index.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+    .catch((err) => {
+      res.status(500);
+      res.send(`Error querying Spanner: ${err}`);
```
jmdobry: Same:

```js
res
  .status(500)
  .send(`Error querying Spanner: ${err}`)
  .end();
```

ace-n: Is the `end()` necessary if we're calling `send()`?

jmdobry: It's a convention I follow in all our App Engine and Cloud Functions samples. `end()` is just saying: "we're done formulating the response".

functions/spanner/test/index.test.js
ace-n: FYI - this causes the test to fail. IIUC, the test itself needs to complete before the stubs can be checked.

(It's probably possible to de-promisify the test itself - but unless we do that, we have to keep this `await` here.)

functions/spanner/index.js
ace-n: Gotcha - will update accordingly.

functions/spanner/index.js
```diff
@@ -0,0 +1,65 @@
...
+  const query = {
+    sql: 'SELECT * FROM Albums'
+  };
+
+  // Execute the query
+  database.run(query)
+    .then((results) => {
+      const rows = results[0].map((row) => row.toJSON());
+      rows.forEach((row) => {
+        res.write(`SingerId: ${row.SingerId.value}, AlbumId: ${row.AlbumId.value}, AlbumTitle: ${row.AlbumTitle}</br>`);
```
jmdobry: Do you need to set the `Content-Type` response header? (If you're using `res.write`, I'm wondering if it will correctly detect the content-type as `text/html`)

ace-n: By default, it sets the `Content-Type` to `text/plain` - which IMO should be fine unless we're specifically trying to show HTML.

(If I convert the `</br>`s to `\n`s, I can get the same desired result with `text/plain`.)

functions/spanner/index.js
```diff
@@ -0,0 +1,65 @@
...
+  const instance = spanner.instance(instanceId);
+  const database = instance.database(databaseId);
+
+  // The query to execute
+  const query = {
+    sql: 'SELECT * FROM Albums'
+  };
+
+  // Execute the query
+  database.run(query)
```
jmdobry: To eliminate possible race condition/flakiness in the test:

s/`database.run(query)`/`return database.run(query)`/

PR #372, opened 2017-05-02
updated readme for the renaming of files

merla18:
needed to update the instructions to run the sample analyze since it was rename analyze.v1.js. Also added the instructions for the v1.1 beta features. Instructions were tested.


jmdobry:
Thanks @merla18, but #370 (which will be merged today) fixes this issue, and actually turns this README into an auto-generated file, hopefully eliminating future "stale README" issues.

PR #378, opened 2017-05-08
Update App Engine Websockets sample.

jmdobry:


andrewsg:
I think we need to add the following to app.yaml to ensure that multiple users all hit the same instance (so that their chat messages, which are only handled locally, propagate):

    manual_scaling:
      instances: 1

If we can think of a pithy, straightforward example that uses multiple instances, then that would be fine too, of course.

bebensiganteng:
Hi guys, just realized that the WebSocket sample has been deleted, is there any update for this?

andrewsg:
We're working on some changes internally, and hope to have a public announcement on this soon.

TeamCapermint:
No updates on this yet?

PR #380, opened 2017-05-08
Update Cloud Functions GCS hello world sample.

jmdobry:


functions/helloworld/index.js
```diff
@@ -101,7 +101,13 @@ exports.helloGCS = function (event, callback) {
   if (isDelete) {
     console.log(`File ${file.name} deleted.`);
   } else {
```
ace-n: Nit: is it worth flattening this into a 1-layer `if` statement?

jmdobry: Done

PR #384, opened 2017-05-09
Upgrade repo tools and regenerate READMEs.

jmdobry:


scripts/generate
```diff
@@ -0,0 +1,77 @@
...
+'use strict';
+
+const childProcess = require('child_process');
+const fs = require(`fs`);
+const path = require(`path`);
+
+const PROJECT_ROOT = path.join(__dirname, `..`);
+
+// Dedupe package.json dependencies
+// WARNING: This will fail if two different versions of the same package are required.
```
ace-n: Nit: is it worth adding a check for this?

scripts/generate
```diff
@@ -0,0 +1,77 @@
...
+    .filter((dir) => fs.statSync(path.join(directory, dir)).isDirectory())
+    .forEach((dir) => {
+      findSamples(path.join(directory, dir), depth - 1);
+    });
+};
+
+findSamples(PROJECT_ROOT, 3);
+
+// Update root-level package.json (by shelling to npm)
+// const spawn = require('child_process').spawn;
```
ace-n: I assume this is supposed to be commented out?

jmdobry: Will remove

scripts/generate
jmdobry: Actually will remove, it's leftover because some of this was copied from your unify script.

PR #386, opened 2017-05-10
Bring Storage samples up to standard.

jmdobry:
Also added a script for cleaning up orphaned test Cloud Storage buckets.

### DIFF
--- a/storage/iam.js
+++ b/storage/iam.js
@@ -20,19 +20,19 @@ function viewBucketIamMembers (bucketName) {
   // Imports the Google Cloud client library
   const Storage = require('@google-cloud/storage');
 
+  // The name of the bucket to access, e.g. "my-bucket"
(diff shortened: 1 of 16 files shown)


storage/iam.js
```diff
@@ -47,7 +47,7 @@ function viewBucketIamMembers (bucketName) {
       });
     })
     .catch((err) => {
-      console.error('Error:', err);
+      console.error('ERROR:', err);
```
ace-n: Question: I don't think we've standardized this - should we?

jmdobry: Every time I've added this catch statement, I've done `ERROR`. Let's get together tomorrow and pound out a Node.js sample style guide.

jmdobry: Started one here: #387

PR #400, opened 2017-06-11
Fix uploaded/updated logic for Cloud Storage

maxbbraun:
The metageneration attribute seems to be a string and not a number. The
strict comparison with 1 in the if/else inverted the logic that
separated file uploads from metadata updates.


maxbbraun:
I signed the CLA.

maxbbraun:
Thanks, tests are updated now.

jmdobry:
Thanks

PR #403, opened 2017-06-19
Initial Cloud Functions auth samples.

jmdobry:


jmdobry:
@jasonpolites PTAL

These cover just two use case dealing with OAuth 2 + Google APIs + Cloud Functions. They both involve direct access by end users. One stores the access token in the user's cookie, the other stores the access token in Cloud Datastore. I plan to add more samples for different use cases.

VikramTiwari:
I am working on one related to SMS auth for twilio. Not sure if I should open a PR for that.

It was for https://github.com/GoogleCloudPlatform/community/issues/159
and the code is here: https://github.com/VikramTiwari/nodejs-examples/tree/sms-autentication-twilio-gcf/sms-authentication-twilio-gcf

jmdobry:
@VikramTiwari I would continue with your plan to submit a PR to GoogleCloudPlatform/community to solve GoogleCloudPlatform/community#159, publishing your tutorial on the community site would be awesome

jasonpolites:
@jmdobry It looks like we have a lot of boiler-plate oAuth code in there.  Is it worth separating this?  So we'd have a function which just assumed a valid oAuth access token in scope, and a separate set of functions to do generic oAuth.  I worry that "calling Google APIs" ends up looking very complicated

jmdobry:
This is the same boilerplate oAuth code that would need to be used in any Node.js app to call Google oAuth APIs that have user-owned resources. None of the complexity is due to it being on Cloud Functions. 

We could vastly simplify the samples by relying on say, Firebase Google oAuth, which does all of this boilerplate for you. Then the Cloud Function could just get the access_token from Firebase in order to do its thing.

jasonpolites:
Yeah that's what I'm suggesting.  Not *necessarily* Firebase Auth (although that would be fine) but just separating the boiler-plate oAuth dance from the actual API call.  Perhaps also an example using Auth0.  We can *also* do a Cloud Function that demonstrates how to do oAuth

jmdobry:
Yeah, I was getting to those, just started with these.

fhinkel:
@jmdobry Can we close this?

PR #413, opened 2017-06-30
Move DLP from REST to gRPC + client libraries

ace-n:


dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
   };
 
-  // Construct REST request for creating an inspect job
-  let options = {
-    url: `${API_URL}/inspect/operations`,
-    headers: {
-      'Authorization': `Bearer ${authToken}`,
-      'Content-Type': 'application/json'
+  // Create a GCS File inspection job, and resolve it using promises
+  // TODO: is the client library going to handle pagination if no user settings are specified?
```
jonparrott: TODOs?

dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
   };
 
-  // Run inspect-job creation REST request
-  requestPromise.post(options)
-    .then((createBody) => pollJob(createBody, inspectConfig.initialTimeout, inspectConfig.tries, authToken))
-    .then((jobName) => getJobResults(authToken, jobName))
-    .then((findingsBody) => {
-      const findings = findingsBody.result.findings;
-      console.log(JSON.stringify(findings, null, 2));
+  // Create a GCS File inspection job, and resolve it using event handlers
```
jonparrott: The latter part of this comment seems unhelpful. Would it be better to say:

```
// Create an operation to instance the data in Google Cloud Storage,
// wait for it to complete, then list the findings.
```

dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
+        operation.on('complete', (completeJobResponse) => {
+          return resolve(completeJobResponse);
+        });
+
+        // Handle changes in job metadata (e.g. progress updates)
+        operation.on('progress', (metadata) => {
+          // TODO this doesn't produce meaningful results
+        });
+
+        // Handle job errors
```
jonparrott: This comment is obvious.

dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
+  dlp.createInspectOperation(request)
+    .then((createJobResponse) => {
+      const operation = createJobResponse[0];
+      return new Promise((resolve, reject) => {
+        // Handle job completion
+        operation.on('complete', (completeJobResponse) => {
+          return resolve(completeJobResponse);
+        });
+
+        // Handle changes in job metadata (e.g. progress updates)
```
jonparrott: This comment is also obvious.

dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
+        operation.on('progress', (metadata) => {
+          // TODO this doesn't produce meaningful results
```
jonparrott: If this does nothing, why not just leave it out?

dlp/inspect.js
```diff
@@ -15,167 +15,140 @@
...
+          console.log(`\tInfo type: ${finding.infoType.name}`);
+          console.log(`\tLikelihood: ${finding.likelihood}`);
+        });
+      } else {
+        console.log(`No findings.`);
+      }
     })
     .catch((err) => {
-      console.log('Error in inspectString:', err);
+      console.log('Error in inspectString:', err.message || err);
```
samtstern: nit: Why not: `Error in inspectString: ${err.message || err}` to be consistent with other logging style.

samtstern: Same comment throughout, since this catch block is copy-pasted

dlp/inspect.js
```diff
@@ -192,56 +174,168 @@ function inspectGCSFile (authToken, bucketName, fileName, inspectConfig) {
...
-    .then((createBody) => pollJob(createBody, inspectConfig.initialTimeout, inspectConfig.tries, authToken))
-    .then((jobName) => getJobResults(authToken, jobName))
-    .then((findingsBody) => {
-      const findings = findingsBody.result.findings;
-      console.log(JSON.stringify(findings, null, 2));
+  // Create a GCS File inspection job, and resolve it using event handlers
+  dlp.createInspectOperation(request)
+    .then((createJobResponse) => {
+      const operation = createJobResponse[0];
+      return new Promise((resolve, reject) => {
```
samtstern: What's the difference between this construct and just `operation.promise()` which you used above?

dlp/metadata.js
```diff
@@ -15,102 +15,90 @@
...
-    .then((body) => {
-      console.log(body);
-    })
-    .catch((err) => {
-      console.log('Error in listInfoTypes:', err);
+  dlp.listInfoTypes({
+    category: category,
+    languageCode: languageCode
+  })
+  .then((body) => {
```
samtstern: What is the type of `body`?  In other then blocks you use more descriptive names like `createJobResponse`

dlp/metadata.js
```diff
@@ -15,102 +15,90 @@
...
+    const infoTypes = body[0].infoTypes;
+    console.log(`Info types for category ${category}:`);
```
samtstern: Any possibility of infoTypes being empty?

ace-n: Having a valid category without valid info types is the only way I can see to trigger this. (Invalid categories are detected outside this sample and the resulting errors are handled correctly.)

dlp/inspect.js
ace-n: This uses event handlers - and I wanted to avoid "callback hell".

(If you think the nested callback approach would be better idiomatically vs. mixing promises and event handlers, I'm happy to switch to it.)

dlp/inspect.js
ace-n: Done - will address.

dlp/inspect.js
ace-n: This provides (and prints out) semi-meaningful results now.

dlp/metadata.js
ace-n: Perhaps `apiResponse` would be a better name here - WDYT?

(Something like `listInfoTypesResponse` seems a little long-winded.)

samtstern: Either one SGTM

PR #438, opened 2017-08-01
Add DLP redact image sample

ace-n:


dlp/redact.js
```diff
@@ -64,6 +64,61 @@ function redactString (string, replaceString, minLikelihood, infoTypes) {
...
+    items: fileItems
+  };
+
+  dlp.redactContent(request)
+    .then((response) => {
+      const image = response[0].items[0].data;
+      fs.writeFileSync(outputPath, image);
+    })
+    .then(() => {
+      console.log(`Saved image redaction results to path: ${outputPath}`);
```
jmdobry: This `console.log` doesn't need to be in its own promise

dlp/redact.js
```diff
@@ -64,6 +64,61 @@ function redactString (string, replaceString, minLikelihood, infoTypes) {
...
+      minLikelihood: minLikelihood
+    },
+    imageRedactionConfigs: imageRedactionConfigs,
+    items: fileItems
+  };
+
+  dlp.redactContent(request)
+    .then((response) => {
+      const image = response[0].items[0].data;
+      fs.writeFileSync(outputPath, image);
```
jmdobry: Best practice would not be to use a blocking call to write the file, though it makes the sample simpler. I'll leave it up to you.

ace-n: Agreed all around - I was going with the simpler-sample approach here.

We could use [this module](https://www.npmjs.com/package/mz), if we don't mind adding additional dependencies.

Otherwise, I'd personally vote for using the `sync` functions.

dlp/redact.js
ace-n: True - unless we decide to promisify `writeFileSync()` in some way. (See my comment above.)

jmdobry: Alright, keep the sync function and move this up.

ace-n:
Closing, as #441 was merged.

PR #459, opened 2017-08-22
KMS: Makes samples consistent across languages.

white0ut:
This change, in particular:

* Consistently names variables with the API (locationId, keyRingId,
  cryptoKeyId).
* Displays the function name in the region tag.
* Correctly writes decoded base64 to disk  and encodes file contents
  before calling KMS.
* Labels parameters as plaintextFileName and ciphertextFileName, rather
  than inFile and outFile where applicable.

Tracking bug: http://b/64758639

### DIFF
--- a/kms/README.md
+++ b/kms/README.md
@@ -39,17 +39,19 @@ __Usage:__ `node keys.js --help`
 
 ```
 Commands:
-  create <keyRing> <cryptoKey>                                           Creates a crypto key.
-  decrypt <keyRing> <cryptoKey> <ciphertextFileName> <plaintextFileName> Decrypts a file.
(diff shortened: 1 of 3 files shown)


ace-n:
I agree with most of these changes, save for the region tag modifications. These snippets are meant to be standalone pieces of code that can be run without being wrapped in a function - the sole purpose of the `function` wrapper is to make testing easier.

@jmdobry might have something to add though.

white0ut:
SGTM, thanks for the quick response.

Fixed.

jmdobry:
I moved your branch to #459 so the tests would run. Closing in favor of #459

PR #468, opened 2017-08-29
Update dependencies

ace-n:


appengine/cloudsql/package.json
```diff
@@ -18,36 +18,53 @@
     "lint": "samples lint",
     "pretest": "npm run lint",
     "unit-test": "ava --verbose test/*.test.js",
+    "start-proxy": "cloud_sql_proxy -instances=$INSTANCE_CONNECTION_NAME=tcp:$SQL_PORT &",
```
jmdobry: Do you need `./cloud_sql_proxy`? Or is it expected that `cloud_sql_proxy` will be on the path somewhere?

ace-n: The latter.

jmdobry: Also, shouldn't this script just conditionally start the proxy? What if it's already running?

jmdobry:
LGTM other than last question about conditionally starting the proxy

PR #471, opened 2017-08-31
Fixing non working example with publish

khaosdoctor:
Publish example was not working, so I updated it to the new API (because I needed to read the source code to understand this part), so it's easier for us to use the docs.

Here's my code as an example:

> publisher.js
```js
const PubSub = require('@google-cloud/pubsub')
const chalk = require('chalk')

function run () {
  console.log(chalk.bold.cyan('Publishing Message...'))
  return PubSub().topic('my-topic').publisher().publish(new Buffer.from(process.argv.pop()))
    .then((results) => {
      console.log(`Message ${chalk.bold.green(results.shift())} published`)
      return results
    })
    .catch(console.error)
}

run()
```

Use it like `node publisher.js <mydata>`

### DIFF
--- a/pubsub/topics.js
+++ b/pubsub/topics.js
@@ -83,13 +83,19 @@ function publishMessage (topicName, data) {
 
   // References an existing topic, e.g. "my-topic"
   const topic = pubsub.topic(topicName);
-
+  
+  // Creates a new Topic Publisher
+  const publisher = topic.publisher()
+  
+  // Creates a new Buffer from data
+  let dataBuffer = new Buffer.from(data)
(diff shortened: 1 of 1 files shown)


khaosdoctor:
I signed it!

ace-n:
Thanks for the PR - but I think #464 already addresses these changes, so I'm closing this for now.

Feel free to comment/reopen if that's not the case.

khaosdoctor:
Yeah! That's the stuff lol we even used the same variable names 🤣 thanks @ace-n!

PR #481, opened 2017-09-13
Split Cloud SQL sample into two samples.

jmdobry:


jmdobry:
Since @ace-n is out, can you please review?

PR #482, opened 2017-09-19
Fix for the bug: pod error ImagePullBackOff

lei-tang:
The container registry b.gcr.io has been deprecated. In this bug fix, the deprecated
container registry has been updated to the correct container registry.


lei-tang:
#483

jeffmendoza:
Awesome, thanks for the fix!

PR #503, opened 2017-10-24
Add missing pub-sub samples

ace-n:
As requested by @jabubake

### DIFF
--- a/pubsub/subscriptions.js
+++ b/pubsub/subscriptions.js
@@ -153,7 +153,6 @@ function modifyPushConfig (topicName, subscriptionName, pushEndpoint) {
     pushEndpoint: `https://${pubsub.projectId}.appspot.com/push`
   };
 
-  // Modify the target subscription's push config
   return subscription.modifyPushConfig(options)
     .then((results) => {
       console.log(`Modified push config for subscription ${subscription.name}.`);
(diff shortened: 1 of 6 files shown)


pubsub/subscriptions.js
```diff
@@ -84,6 +84,32 @@ function createSubscription (topicName, subscriptionName) {
 }
 // [END pubsub_create_subscription]
 
+// [START pubsub_subscriber_flow_settings]
+function createFlowControlledSubscription (topicName, subscriptionName, maxInProgress, maxBytes) {
```
jmdobry: I guess the Pub/Sub samples were never updated to hide the method definition?

pubsub/subscriptions.js
```diff
@@ -84,6 +84,32 @@ function createSubscription (topicName, subscriptionName) {
...
+      maxMessages: maxInProgress
+    }
+  })
+    .then((results) => {
+      const subscription = results[0];
+
+      console.log(`Subscription ${subscription.name} created with a maximum of ${maxInProgress} unprocessed messages.`);
+
+      return subscription;
+    });
```
jmdobry: And weren't updated to handle errors.

jmdobry (approved):

pubsub/subscriptions.js
ace-n: Correct - I can address that in a separate PR, if you'd like.

pubsub/subscriptions.js
ace-n: Correct - see above comment.

pubsub/subscriptions.js
jmdobry: Yes please

PR #517, opened 2017-11-02
Adds token refresh to HTTP example.

gguuss:


iot/http_example/cloudiot_http_example_nodejs.js
```diff
@@ -145,6 +151,15 @@ function publishAsync (messageCount, numMessages) {
       // If we have published fewer than numMessage messages, publish payload
       // messageCount + 1.
       setTimeout(function () {
+        let secsFromIssue = parseInt(Date.now() / 1000) - iatTime;
+        if (secsFromIssue > argv.token_exp_mins * 60) {
+          iatTime = parseInt(Date.now() / 1000);
+          console.log(`\tRefreshing token after ${secsFromIssue} seconds.`);
+
+          authToken = createJwt(argv.project_id, argv.private_key_file,
```
ace-n: Nit: this feels like it should be a one-liner.

You can get it under 80 characters using camel case (like the non-IoT samples) and shortening `privateKeyFile` to `keyFile`. (Worst case scenario, it's probably fine to go a little over 80 characters unless the linter says not to.)

gguuss: done

PR #519, opened 2017-11-08
bigquery-stream-data-nodejs docs - use correct json property to check for errors

blundell:
Fix from the documentation here:

https://cloud.google.com/bigquery/streaming-data-into-bigquery#bigquery-stream-data-nodejs

- name json correctly and then use the correct child value to check for errors
- also log in a way that doesn't matter if you read logs top down or bottom up (i.e. firebase function logs vs cmd line)
- use JSON.stringify for more readable output

paired with @ouchadam

### DIFF
--- a/bigquery/tables.js
+++ b/bigquery/tables.js
@@ -415,21 +415,24 @@ function insertRowsAsStream (datasetId, tableId, rows, projectId) {
 
   // Inserts data into a table
   bigquery
-    .dataset(datasetId)
-    .table(tableId)
-    .insert(rows)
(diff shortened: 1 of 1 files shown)


blundell:
screen shot 2017-11-09 at 10 00 19 pm

not sure the failure is something I (a non admin) can fix?

jmdobry:
Thanks for this, I moved the fix to here: https://github.com/googleapis/nodejs-bigquery/pull/29/files#diff-41e74da76efff7f0fbf13acaed97ec9e

PR #521, opened 2017-11-10
MQTT client JWT refresh and set TLS to 1.2

gguuss:


gguuss:
@ace-n Friendly ping :)

PR #528, opened 2017-12-04
Added skip_files to prevent errors on deploy

johnpyp:
In my tutorial run, there was an error during deploying the Hello World app. It said you can't have a yarn.lock and package.lock file when deploying, and to add skip_files to ignore one of them.

### DIFF
--- a/appengine/hello-world/app.yaml
+++ b/appengine/hello-world/app.yaml
@@ -14,4 +14,5 @@
 # [START app_yaml]
 runtime: nodejs
 env: flex
+skip_files: yarn.lock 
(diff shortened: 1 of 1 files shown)


johnpyp:
I signed agreement

jmdobry:
Thanks for pointing this out. However, our sample currently doesn't have either a `package-lock.json` or a `yarn.lock` file, and I don't really want favor one package manager over the other, or mess with Yarn users who may be trying to deploy this sample. My recommendation is to only use one package manager and to delete the lock files from other managers.

PR #529, opened 2017-12-04
System-ify functions tests, part 1 (hello-world + datastore)

ace-n:
~~**Do not merge** yet - this PR removes a sample that's still in the docs.~~


ace-n:
@jmdobry This should be good to review, but we'll either a) need to move the Datastore CI to Kokoro or b) make sure it passes on CircleCI (which is doable, but will require installing the emulator).

jmdobry:
Approved, do what you need to do

PR #538, opened 2017-12-14
Use long options for the ESP config

navinger:
For better readability, let's use the long name options for the ESP config.


kryzthov:
Thanks for doing that, Nancy!
LGTM

PR #544, opened 2018-01-03
Add region tags to JWT refresh.

noerog:


gguuss:
@googlebot I signed it!

Checks should be green, both ben and I have signed, merging.

PR #550, opened 2018-01-10
Add comment about subscribing for config updates.

noerog:


gguuss:
oops, @googlebot I signed it!

PR #556, opened 2018-01-21
Update arg format in readme from snake-case to camelCase

ptone:
@gguuss PTAL - looks like a copy paste from python sample perhaps


gguuss:
@GoogleCloudPlatform/googlebot I signed it!

PR #568, opened 2018-02-13
Update index.js

labtopia:
Per b/73223200, "The "helloGCSGeneric" function name should be "helloGCS" according to deploy command."


ace-n:
There are two `helloGCS` functions in that file. Perhaps this should be fixed in the docs instead? (Change the command-line sample to use `helloGCSGeneric` instead of `helloGCS`.)

PR #569, opened 2018-02-17
Updates to v1 base URI

gguuss:


gguuss:
@jdobry /FYI merging before EOD

PR #573, opened 2018-02-27
Fixes script name for HTTP example

gguuss:


gguuss:
/@jmdobry fyi, crit fix for README and GCF is currently failing.

PR #576, opened 2018-03-08
Change to endpoints managed rollout

simonz130:
We introduce a new feature where we can specify in YAML that service configuration for Endpoints is going to be managed. This means that the service will pull latest deployed service configuration automatically, rather than requiring customer to specify specific version of service configuration.

New flag:
**--rollout_strategy=managed**
Previously the only way to specify a version was using the following flag:
**--version=SERVICE_CONFIG_ID**

### DIFF
--- a/endpoints/getting-started-grpc/README.md
+++ b/endpoints/getting-started-grpc/README.md
@@ -44,15 +44,18 @@ $ protoc \
 
 1. In `api_config.yaml`, replace `MY_PROJECT_ID` and `SERVICE-ACCOUNT-ID` with your Project ID and your service account's email address respectively.
 
-1. Deploy your service's configuration to Endpoints. 
+1. Deploy your service's configuration to Endpoints. Take note of your service's name once the deployment completes.
 ```
 $ gcloud endpoints services deploy api.pb api_config.yaml
 ...
+
 ```
 
 1. Build a Docker image for later use using the following command. Make sure to replace `[YOUR_PROJECT_ID]` with your Project ID.
 ```
 $ gcloud container builds submit --tag gcr.io/[YOUR_PROJECT_ID]/endpoints-example:1.0 .
+...
+Service Configuration [SERVICE_CONFIG_ID] uploaded for service [SERVICE_NAME]
 ```
 
 ### Running your service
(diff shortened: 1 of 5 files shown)


endpoints/getting-started-grpc/README.md
```diff
@@ -44,11 +44,10 @@ $ protoc \
 
 1. In `api_config.yaml`, replace `MY_PROJECT_ID` and `SERVICE-ACCOUNT-ID` with your Project ID and your service account's email address respectively.
 
-1. Deploy your service's configuration to Endpoints. Take note of your service's config ID and name once the deployment completes.
+1. Deploy your service's configuration to Endpoints. 
 ```
 $ gcloud endpoints services deploy api.pb api_config.yaml
 ...
-Service Configuration [SERVICE_CONFIG_ID] uploaded for service [SERVICE_NAME]
```
wmwong: This was the only place where we define what SERVICE_NAME is. Without this, users may be confused about what SERVICE_NAME is.

I think we should leave this here, even if we don't end up using SERVICE_CONFIG_ID.

endpoints/getting-started-grpc/README.md
```diff
@@ -44,11 +44,10 @@ $ protoc \
 
 1. In `api_config.yaml`, replace `MY_PROJECT_ID` and `SERVICE-ACCOUNT-ID` with your Project ID and your service account's email address respectively.
 
-1. Deploy your service's configuration to Endpoints. Take note of your service's config ID and name once the deployment completes.
+1. Deploy your service's configuration to Endpoints.
```
wmwong: We should still say:
Take note of your service's name once the deployment completes.

endpoints/getting-started-grpc/README.md
```diff
@@ -107,7 +106,7 @@ $ gcloud components install kubectl
 $ gcloud container clusters get-credentials [YOUR_CLUSTER_NAME] --zone [YOUR_CLUSTER_ZONE]
 ```
 
-1. Edit the `container_engine.yaml` file, and replace `GCLOUD_PROJECT`, `SERVICE_NAME`, and `SERVICE_CONFIG` with your Project ID and your Endpoints service's name and config ID respectively.
+1. Edit the `container_engine.yaml` file, and replace `GCLOUD_PROJECT` and `SERVICE_NAME` with your Project ID and your Endpoints service's name.
```
wmwong: Change "Project ID" -> "GCloud Project ID".
Change "Endpoints service's name" -> "service's name respectively".

This matches the instructions above.

simonz130:
circleci timeout failure is due to functions emulator and has nothing to do with these changes.

PR #582, opened 2018-03-27
fix the params sent across to the ga endpoint.

oluomoniyi:
spreading the results allows the endpoint to pick up the individual params. Previous method does not work.
see https://developers.google.com/analytics/devguides/collection/protocol/v1/reference for more

### DIFF
--- a/appengine/analytics/app.js
+++ b/appengine/analytics/app.js
@@ -47,9 +47,7 @@ function trackEvent (category, action, label, value, cb) {
     ev: value
   };
 
-  return got.post('http://www.google-analytics.com/collect', {
-   ...data
-  });
+  return got.post('http://www.google-analytics.com/collect', Object.assign({}, data));
 }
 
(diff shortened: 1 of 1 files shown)


oluomoniyi:
I have signed it!

oluomoniyi:
added the semicolon

appengine/analytics/app.js
```diff
@@ -48,7 +48,7 @@ function trackEvent (category, action, label, value, cb) {
   };
 
   return got.post('http://www.google-analytics.com/collect', {
-    form: data
+   ...data
```
jmdobry: This operator only works in Node v8.6+, but we strive to make sure our samples work with the oldest LTS version of Node, which until the end of April is Node v4. Can you update your PR to be Node v4 compatible?

oluomoniyi: Hi @jmdobry,

Makes sense, updated to use Object.assign instead.

Issue #596, opened 2018-04-04
Broken link in App Engine/Cloud Tasks ReadMe


averikitsch:
In nodejs-docs-samples/appengine/cloudtasks/README.md:
"Enable the Cloud Tasks API in the [Google Cloud Console](https://cloud.google.com/apis/library/cloudtasks.googleapis.com)."

The Google Cloud Console link should be https://console.cloud.google.com/apis/api/cloudtasks.googleapis.com.

PR #598, opened 2018-04-05
Fix broken GCP console link

campionfellin:
This fixes https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/596

With the link to the Tasks API.

Though note that there are 2 other broken links on that page, and it's hard to tell where they're supposed to go:

Line 3:
`[Google Cloud Tasks](https://cloud.google.com/cloud-tasks/)` maybe it should link to:
`https://cloud.google.com/sdk/gcloud/reference/alpha/tasks/` instead?

Line 86:
`https://cloud.google.com/cloud-tasks` should be changed similarly.


campionfellin:
Hey, I'm curious about the circleci build and why it fails sometimes. It looks like the 2nd step in dependencies is:

`gcloud auth activate-service-account --key-file /home/ubuntu/nodejs-docs-samples/key.json || true`

and in failing builds, you get this:

`ERROR: (gcloud.auth.activate-service-account) Could not read json file /home/ubuntu/nodejs-docs-samples/key.json: No JSON object could be decoded`

but since the `|| true`, it passes this portion of the build. I think this causes problems later with the first part of the functions test when it tries:

`functions-emulator config set projectId $GCLOUD_PROJECT` but fails to find the env variable $GCLOUD_PROJECT and then times out.

Maybe we should remove the `|| true` from when we activate the service account so it fails there?

Just a thought, though this could be a separate issue or PR

fhinkel:
@campionfellin Thanks so much. Sorry that this was sitting here forever. 
Any chance you did a quick grep if the wrong URL is used anywhere else? 

About the CI: It's not in good shape :cry: All the recent runs didn't even get to that error.

campionfellin:
Hey @fhinkel are you on `googlecloud-community.slack.com` channel? We're having similar problems with our CI over at [clasp](https://github.com/google/clasp) and I'm thinking of a solution, I'd love to run it by you.

campionfellin:
Ah looks like the build passed... @JustinBeckwith did you change the pipeline recently? Looks like it's still getting this error:
```
gcloud auth activate-service-account --key-file /home/ubuntu/nodejs-docs-samples/key.json || true
ERROR: (gcloud.auth.activate-service-account) Could not read json file /home/ubuntu/nodejs-docs-samples/key.json: No JSON object could be decoded
```

Yet somehow passes all the tests later.

JustinBeckwith:
I haven't changed a thing 🤷‍♂️

PR #601, opened 2018-04-07
Remove redundant error listener from Redis client

LewisJEllis:
This is actually a memory leak, attaching an additional listener for each request, but the client already has an error listener.


fhinkel:
@LewisJEllis Thank so much, good find! Sorry that this sat here for so long.

LewisJEllis:
Yay! Thanks :)

PR #606, opened 2018-04-28
Fix readme: app.yaml has been renamed

steren:


steren:
ping?

This is a one line README change and it's been 20 days.

PR #609, opened 2018-05-03
Clarify comments + add field handler to multipart sample

ace-n:
~**Do not merge** until the docs have been updated accordingly.~

cc @marekbiskup

### DIFF
--- a/functions/http/index.js
+++ b/functions/http/index.js
@@ -147,22 +147,22 @@ const Busboy = require('busboy');
 exports.uploadFile = (req, res) => {
   if (req.method === 'POST') {
     const busboy = new Busboy({ headers: req.headers });
+    const tmpdir = os.tmpdir();
 
     // This object will accumulate all the fields, keyed by their name
     const fields = {};
 
     // This object will accumulate all the uploaded files, keyed by their name.
     const uploads = {};
-    const tmpdir = os.tmpdir();
 
-    // This event will be triggered for each non-file field in the form.
+    // This code will process each non-file field in the form.
     busboy.on('field', (fieldname, val) => {
       // TODO(developer): Process submitted field values here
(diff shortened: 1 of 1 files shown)


functions/http/index.js
```diff
@@ -148,20 +148,31 @@ exports.uploadFile = (req, res) => {
...
 
+    // This object will accumulate all the fields, keyed by their name
+    const fields = {};
+
     // This object will accumulate all the uploaded files, keyed by their name.
     const uploads = {};
     const tmpdir = os.tmpdir();
 
-    // This callback will be invoked for each file uploaded.
+    // This event will be triggered for each non-file field in the form.
```
marekbiskup: I'm a bit afraid that users may confuse events described above, with event that trigger a function. For instance, users may think that there are many function executions here.

I think the previous form ("This callback will be invoked") was a bit better

Also, busboy docs say that events are emitted, not triggered.

How about something like this:

- "This code is executed on..."
- "This code handles busboy events emitted on..."
- "This busboy events are emitted on..."

functions/http/index.js
```diff
@@ -148,20 +148,31 @@ exports.uploadFile = (req, res) => {
   if (req.method === 'POST') {
     const busboy = new Busboy({ headers: req.headers });
 
+    // This object will accumulate all the fields, keyed by their name
+    const fields = {};
+
     // This object will accumulate all the uploaded files, keyed by their name.
     const uploads = {};
     const tmpdir = os.tmpdir();
```
marekbiskup: I would add an empty line before:
const tmpdir = os.tmpdir();

Otherwise it looks like the comment "This object will accumulate all the uploaded files, keyed by their name" is for the two lines.

You can group this "const tmpdir" together with  "const busboy"

PR #617, opened 2018-05-10
Update server.js

michaelawyu:
`/computeMetadata/v1/instance/network-interfaces` is not support in standard.


appengine/metadata/standard/server.js
```diff
@@ -22,31 +22,32 @@ const request = require('got');
 const app = express();
 app.enable('trust proxy');
 
-const METADATA_NETWORK_INTERFACE_URL = 'http://metadata/computeMetadata/v1/instance/network-interfaces/0/access-configs/0/external-ip';
+const METADATA_PROJECT_ID_URL = 'http://metadata.google.internal/computeMetadata/' +
```
ace-n: Nit: can this be on one line? (You may have to add an `// eslint-disable` at the end of it if you get linter issues.)

appengine/metadata/standard/server.js
```diff
@@ -22,31 +22,32 @@ const request = require('got');
...
+  'v1/project/project-id';
 
-function getExternalIp () {
+function getProjectId () {
   const options = {
     headers: {
       'Metadata-Flavor': 'Google'
     },
-    json: true
+    json: false
```
ace-n: Nit: do we need to explicitly specify this?

appengine/metadata/standard/server.js
```diff
@@ -22,31 +22,32 @@ const request = require('got');
...
   };
 
-  return request(METADATA_NETWORK_INTERFACE_URL, options)
+  return request(METADATA_PROJECT_ID_URL, options)
     .then((response) => response.body)
     .catch((err) => {
       if (err && err.statusCode !== 200) {
-        console.log('Error while talking to metadata server, assuming localhost');
-        return 'localhost';
+        console.log('Error while talking to metadata server, assuming UnknownProjectID');
```
ace-n: Could we use a more obvious default value? Or perhaps throw an error outright?

michaelawyu:
@ace-n Updated as requested, thanks :)

PR #618, opened 2018-05-12
Add handler section to app.yaml

michaelawyu:


steren:
Do not submit this commit.

Handlers is still optional. No need to add it.
What has changed is that if present, then handlers.script must be "auto".
If our sample do not use handlers.script, then no need to change anything.

michaelawyu:
I seem to have misunderstood the email Steren sent a little bit :( Really sorry about this. I'll close the commit now.

steren:
No worries, thanks for your help

PR #619, opened 2018-05-12
Update memcached sample

michaelawyu:


appengine/memcached/app.js
```diff
@@ -23,11 +23,7 @@ const app = express();
 
 // [START client]
 // Environment variables are defined in app.yaml.
-let MEMCACHE_URL = process.env.MEMCACHE_URL || '127.0.0.1:11211';
-
-if (process.env.USE_GAE_MEMCACHE) {
-  MEMCACHE_URL = `${process.env.GAE_MEMCACHE_HOST}:${process.env.GAE_MEMCACHE_PORT}`;
-}
+let MEMCACHE_URL = process.env.MEMCACHE_URL;
```
ace-n: Two questions:
1) Is `MEMCACHE_URL` guaranteed to be defined? (Both on GAE, and in our tutorials)
2) If not, is there ever a situation where `127.0.0.1:11211` makes sense as a default `MEMCACHE_URL` value?

michaelawyu: Node-on-standard doesn't have built-in memcached service anymore :( 127.0.0.1:11211 is no longer available. In our tutorials there is a clear instruction suggesting users to deploy a free memcached instance provided by RedisLab.

PR #625, opened 2018-05-25
puppeteer sample (Headless Chrome)

steren:


appengine/headless-chrome/test/app.test.js
```diff
@@ -0,0 +1,39 @@
...
+'use strict';
+
+const test = require(`ava`);
+const path = require(`path`);
+const utils = require(`@google-cloud/nodejs-repo-tools`);
+
+const cwd = path.join(__dirname, `../`);
+const requestObj = utils.getRequest({ cwd: cwd });
+
+test.serial.cb(`should return a screenshot`, (t) => {
```
fhinkel: Can you remove the extra `()` when it's just one parameter and `{}` when it's just one statement please.

appengine/headless-chrome/README.md
```diff
@@ -0,0 +1,12 @@
+# Headless Chrome on Google App Engine
+
+This sample application demonstrates how to use Headless Chrome via the [Puppeteer](https://developers.google.com/web/tools/puppeteer/) module to take screenshots of webpages on [Google App Engine](https://cloud.google.com/appengine) Node.js [standard environment](https://cloud.google.com/appengine/docs/standard/nodejs).
+
+## Running locally
+
+* Install dependencies with `npm install` (Headless Chrome will take some time to download)
```
fhinkel: I think we either need a period or extra line breaks here or this will look weird.

appengine/headless-chrome/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "headless-chrome-sample",
+  "engines": {
+    "node": "8.x"
+  },
+  "version": "1.0.0",
+  "description": "An example of taking screenshot with Puppeteer (Headless Chrome) in Node.js on Google App Engine.",
+  "scripts": {
+    "start": "node app.js",
+    "system-test": "repo-tools test app",
```
fhinkel: We only have one test but we're using it both as system-test and as unit-test :wink: It should be run as one or the other.

appengine/headless-chrome/test/app.test.js
```diff
@@ -0,0 +1,39 @@
...
+  requestObj
```
fhinkel: Currently, this test does not test the sample. I'd expect something like `require('./app.js')` in the test. Then the test can start the server from the sample, send a request and do the checks like you have, and shut down the server.

appengine/headless-chrome/README.md
```diff
@@ -0,0 +1,12 @@
+# Headless Chrome on Google App Engine
+
+This sample application demonstrates how to use Headless Chrome via the [Puppeteer](https://developers.google.com/web/tools/puppeteer/) module to take screenshots of webpages on [Google App Engine](https://cloud.google.com/appengine) Node.js [standard environment](https://cloud.google.com/appengine/docs/standard/nodejs).
```
fhinkel: I think the prerequisites are missing here. Do we not link to the `gcloud` "Before you get started" section in these samples?

appengine/headless-chrome/test/app.test.js
steren: done for `()`, I tried to remove the `{}` but tests fail.

appengine/headless-chrome/README.md
steren: I am not sure what you mean. This is just one line in a bullet point. I added a comma at the end.

appengine/headless-chrome/package.json
steren: I am not sure I understand your comment. Can you be more specific about what I should do?

We have 2 different kind of tests runnign here. both of them are run when running `nom run test`

appengine/headless-chrome/test/app.test.js
steren: It does tests and the test passes.

I think the magic happens with `requestObj`.

appengine/headless-chrome/README.md
steren: No we do not.
But I am happy to add it.

appengine/headless-chrome/package.json
fhinkel: But there's only one test file with one test. Where's the second test?

appengine/headless-chrome/test/app.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+const cwd = path.join(__dirname, `../`);
+const requestObj = utils.getRequest({ cwd: cwd });
+
+test.serial.cb(`should return a screenshot`, t => {
+  requestObj
+    .get(`/?url=https://example.com`)
+    .send()
+    .expect(200)
+    .expect((response) => {
```
fhinkel: Please remove the  extra `()` if it's a single parameter.

appengine/headless-chrome/README.md
fhinkel: "Refer to the appengine/README.md file for more instructions on running and deploying this app" is the sentence most examples in the repo use. Can we do the same? Then we also have the section on credentials.

appengine/headless-chrome/package.json
steren: Please compare to the other samples in this repository, I tried to mimic the architecture  as much as possible.
The test run with `system-test` is defined in `package.json` in `cloud-repo-tools`

steren: See for example https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/appengine/datastore

appengine/headless-chrome/test/app.test.js
steren: Done

appengine/headless-chrome/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "author": {
+    "name": "Google LLC"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "express": "^4.16.3",
+    "puppeteer": "^1.2.0"
+  },
+  "devDependencies": {
+    "@google-cloud/nodejs-repo-tools": "2.2.1",
```
fhinkel: Any reason we're not using `2.3.0`?

steren: Updated

appengine/headless-chrome/README.md
steren: Sounds good, I just pushed a change that point to the main README

PR #627, opened 2018-05-30
Fix busboy sample not working on GCF

ace-n:
**Note:** this change breaks the `uploadFile` sample in the Emulator (where `req.pipe(busboy)` is correct). @jmdobry can you investigate?

**Do not merge** until we've resolved this issue in both GCF itself and the Emulator.

### DIFF
--- a/functions/http/index.js
+++ b/functions/http/index.js
@@ -182,7 +182,7 @@ exports.uploadFile = (req, res) => {
       res.send();
     });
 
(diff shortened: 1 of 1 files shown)


ace-n:
The emulator was fixed in [this PR](https://github.com/GoogleCloudPlatform/cloud-functions-emulator/pull/263), so I'm merging this.

PR #636, opened 2018-06-05
Add missing lint dependency

fhinkel:
`npm test`, which runs the linter, needs this dependency explicitly. 

We install `semistandard` manually on CircleCi. But since it's required in a sample's `run test`, it should be in the `package.json`.

### DIFF
--- a/appengine/analytics/package.json
+++ b/appengine/analytics/package.json
@@ -24,7 +24,8 @@
     "got": "8.2.0"
   },
   "devDependencies": {
-    "@google-cloud/nodejs-repo-tools": "2.2.1"
+    "@google-cloud/nodejs-repo-tools": "2.2.1",
+    "semistandard": "^12.0.1"
   },
   "cloud-repo-tools": {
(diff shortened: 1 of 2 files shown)


appengine/analytics/package.json
```diff
@@ -21,7 +21,8 @@
   },
   "dependencies": {
     "express": "4.16.2",
-    "got": "8.2.0"
+    "got": "8.2.0",
+    "semistandard": "^12.0.1"
```
JustinBeckwith: Shouldn't this be a dev dependency?  Here and below!

JustinBeckwith:
Can you also run `semistandard fix`?  Guessing there are changes this would need :)

appengine/analytics/package.json
fhinkel: Oh, yes, you're right. Fixed.

fhinkel:
`semistandard fix` didn't create any changes and `npm run lint` is green.

PR #638, opened 2018-06-05
chore: run repo-tools generate

fhinkel:
Auto-generate the list of contributors. Add the generated CoC.


package.json
```diff
@@ -41,6 +27,101 @@
       "coverage"
     ]
   },
+  "contributors": [
```
JustinBeckwith: Do we need to maintain this list in two places?

PR #648, opened 2018-06-06
Add Kokoro config for standard environment

fhinkel:
Add test config for appengine/datastore/standard.

### DIFF
--- a/.kokoro/appengine-datastore-flex.cfg
+++ b/.kokoro/appengine-datastore-flex.cfg
@@ -1,0 +1,16 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Download secrets from Cloud Storage.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/nodejs-docs-samples"
+
+# Tell the trampoline which build file to use.
+env_vars: {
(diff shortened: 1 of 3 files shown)


.kokoro/appengine-datastore-flex.cfg
```diff
@@ -0,0 +1,16 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Download secrets from Cloud Storage.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/nodejs-docs-samples"
```
tbpg: This could be moved to a `common.cfg` file, then it wouldn't need to be copied into every `.cfg` file.

.kokoro/appengine-datastore-flex.cfg
```diff
@@ -0,0 +1,16 @@
+# Format: //devtools/kokoro/config/proto/build.proto
```
tbpg: Idea (feel free to ignore): maybe create an `appengine` directory to make things a little easier to find?

fhinkel: I was going to add a few tests first to make sure they run as expected. And then worry about duplication and folders.

PR #653, opened 2018-06-12
No need to await for the browser to be closed to send back the screenshot

glaforge:
Awaiting for the headless browser to be closed add a certain amount of time to the duration of the request, although the image is already ready to be sent back to the user.

/cc @steren

### DIFF
--- a/appengine/headless-chrome/app.js
+++ b/appengine/headless-chrome/app.js
@@ -35,7 +35,7 @@ app.use(async (req, res) => {
   const page = await browser.newPage();
   await page.goto(url);
   const imageBuffer = await page.screenshot();
-  await browser.close();
+  browser.close();
 
   res.set('Content-Type', 'image/png');
   res.send(imageBuffer);


steren:
Thanks!

steren:
Thanks!

appengine/headless-chrome/app.js
```diff
@@ -35,7 +35,7 @@ app.use(async (req, res) => {
   const page = await browser.newPage();
   await page.goto(url);
   const imageBuffer = await page.screenshot();
-  await browser.close();
+  browser.close();
```
MylesBorins: would this have made more sense to put the res.set & res.send calls before `await browser.close`? Keeping the await ensured that the browser session was closed without error before the route was finished.

Albeit there is no error handling in this example, so any throw would cause an unhandled exception error, but it does seem odd to me to have this async operation continue doing it's thing after the function has returned

PR #654, opened 2018-06-12
No need for the --disable-setuid-sandbox flag

glaforge:
It seems the --disable-setuid-sandbox flag is becoming deprecating and is being removed from the Chrome codebase.
I tried running the sample without this flag, and the screenshots are properly generated.

/cc @steren

### DIFF
--- a/appengine/headless-chrome/app.js
+++ b/appengine/headless-chrome/app.js
@@ -29,7 +29,7 @@ app.use(async (req, res) => {
 
   // [START browser]
   const browser = await puppeteer.launch({
-    args: ['--no-sandbox', '--disable-setuid-sandbox']
+    args: ['--no-sandbox']
   });
   // [END browser]
   const page = await browser.newPage();


steren:
Thanks!

appengine/headless-chrome/app.js
```diff
@@ -29,7 +29,7 @@ app.use(async (req, res) => {
 
   // [START browser]
   const browser = await puppeteer.launch({
-    args: ['--no-sandbox', '--disable-setuid-sandbox']
+    args: ['--no-sandbox']
```
MylesBorins: has this been confirmed in a deplyoment? If I recall the `setuid-sandbox` was necessary on standard

fhinkel: Tests are green with and without the flag. @steren, how much confidence do you have in your test? :wink:

fhinkel: Ran it manually on standard. Works fine without the flag.

PR #657, opened 2018-06-14
Upgrade repo-tools to fix some failing system tests.

jmdobry:


appengine/metadata/flexible/package.json
```diff
@@ -23,7 +23,7 @@
     "got": "8.2.0"
   },
   "devDependencies": {
-    "@google-cloud/nodejs-repo-tools": "2.2.1"
+    "@google-cloud/nodejs-repo-tools": "2.3.0"
```
fhinkel: Should we use `^2.3.0`? I usually do that so I don't have to update all subfolders manually on the next upgrade.

PR #666, opened 2018-06-18
Simplify build script

fhinkel:
Use a shorter version string. 

Do not check for `worker.yaml`, as no sample has one.


.kokoro/build-with-appengine.sh
```diff
@@ -49,12 +49,10 @@ gcloud config set project $GCLOUD_PROJECT
...
-  gcloud app deploy worker.yaml --version ${GAE_VERSION} --no-promote --quiet
-fi
 
 
-# Test the deployed app
+# Install dependencies and run tests
+npm install
 npm test
 
 exit $?
```
dzlier-gcp: Do the node packages care about ending newlines?

fhinkel: Probably not.

PR #674, opened 2018-06-22
Generate configs

SurferJeffAtGoogle:


.kokoro/error-reporting.cfg
```diff
@@ -1,13 +1,13 @@
...
-env_vars: {
-    key: "TRAMPOLINE_BUILD_FILE"
-    value: "github/nodejs-docs-samples/.kokoro/build.sh"
-}
+﻿# Format: //devtools/kokoro/config/proto/build.proto
+
+# Set the folder in which the tests are run
+env_vars: {
+    key: "PROJECT"
+    value: "error/reporting"
```
fhinkel: The folder is actually called `error-reporting`.

fhinkel:
Thanks! Everything in `appengine` or `functions` needs to end up in the `appengine` or `functions` directory, respectively. For everything else we don't have subfolders because they would mostly only have one entry. But those would shift around later, as we figure out common environment variables and build scripts.

SurferJeffAtGoogle:
Ready for another look.

PR #676, opened 2018-06-25
Update Tasks Samples for App Engine

averikitsch:


appengine/cloudtasks/test/createTask.test.js
```diff
@@ -1,61 +1,41 @@
...
 
-const proxyquire = require(`proxyquire`).noCallThru();
-const sinon = require(`sinon`);
+const path = require(`path`);
 const test = require(`ava`);
 const tools = require(`@google-cloud/nodejs-repo-tools`);
 
-test.before(tools.stubConsole);
-test.after.always(tools.restoreConsole);
+const {
```
fhinkel: Can we fit this in one line or is it too long? If it is too long, I think it should be written as 

```
const {runAsyng} = 
  require(...);
```

appengine/cloudtasks/README.md
```diff
@@ -86,6 +86,51 @@ HTTP POST request and logs it. The log output can be viewed with:
...
+
+```
+export PROJECT_ID=my-project-id
+```
+
+Then the queue ID, as specified at queue creation time. Queue IDs already
+created can be listed with `gcloud alpha tasks queues list`.
+
+```
+export QUEUE_ID=my-appengine-queue
```
fhinkel: Is this correct? Further down we check for `GCP_QUEUE`

PR #687, opened 2018-07-16
update gcloud command

averikitsch:


averikitsch:
\assign @fhinkel 
Note: the failing cloudtasks tests are for samples that will be deprecated

PR #691, opened 2018-07-20
Make logging sample an HTTP function

ace-n:


ericschmidtatwork:
Would a `console.error()` example be helpful to add too?

ace-n:
@ericschmidtatwork added (in #672)

PR #697, opened 2018-07-24
Node 6: add Firebase snippets

ace-n:
**Do not merge** until we figure out the following:
1) Are these needed within our docs? **Yes**
2) Do they work with Node 8? **No**, they depend on a callback. We'll have to add Node 8 samples separately.
3) Does the analytics trigger _actually work_? **No**, so we'll have to revisit it later.


ace-n:
@mogar1980 FYI

functions/firebase/index.js
```diff
@@ -0,0 +1,84 @@
...
+  console.log(JSON.stringify(event));
+
+  // Don't forget to call the callback.
+  callback();
+};
+// [END functions_firebase_firestore]
+
+// [START functions_firebase_auth]
+/**
+ * Triggered by a change to a Firebase Auth user object.
```
samtstern: I think we only have triggers for user creation and deletion right now, not on any change to the user object.

functions/firebase/index.js
```diff
@@ -0,0 +1,84 @@
...
+    console.error(err);
+  }
+  // Don't forget to call the callback.
+  callback();
+};
+// [END functions_firebase_auth]
+
+// [START functions_firebase_analytics]
+/**
+ * Triggered by a Firebase Mobile Analytics log event.
```
samtstern: We call it "Google Analytics for Firebase"

samtstern:
Two small comments but LGTM (assuming they actually work, I did not take the time to deploy them)

PR #703, opened 2018-08-03
Remove tasks samples

ace-n:
Now that the docs use the samples in the `googleapis` repo, we can get rid of these.

~**Do not merge** until the `cloudtasks` test config has been removed internally.~


averikitsch:
Do you need to delete the Kokoro (cloudtasks.cfg) file too?

ace-n:
@averikitsch Not strictly necessary AFAIK, but good catch nonetheless. 😄

PR #708, opened 2018-08-10
[KMS] Generalizing KMS samples for GCS docs

frankyn:
Hi, 

I'm removing global location mention to generalize sample for GCS CMEK documentation. The sample is used to show developers how to set an IAM policy on a KMS key, but global location aren't supported so I' want to generalize that part of the sample.

PTAL, thank you!


frankyn:
Thanks @JustinBeckwith

frankyn:
Merging as failed test is not related to this sample. (Cloud SQL for PostgreSQL failure)

PR #710, opened 2018-08-14
Add gcp-build samples

ace-n:
~~**Do not merge** - this is missing a (TypeScript) sample.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,16 +444,6 @@ View the [tutorial][bookshelf_docs] or the [source code][bookshelf_code].
 [bookshelf_docs]: https://cloud.google.com/nodejs/getting-started/tutorial-app
 [bookshelf_code]: https://github.com/GoogleCloudPlatform/nodejs-getting-started
 
-### LabelCat
-
-[LabelCat][labelcat_docs] is a NodeJS app that uses the Google Cloud Prediction
-API to automatically label GitHub Issues as they are created.
-
-View the [tutorial][labelcat_docs] or the [source code][labelcat_code].
-
-[labelcat_docs]: https://cloud.google.com/nodejs/resources/examples/labelcat
-[labelcat_code]: https://github.com/GoogleCloudPlatform/LabelCat
-
 ### Node.js Codelab
 
 In the [Node.js Web App Google Cloud Platform][codelab_docs] codelab, you learn
--- a/appengine/analytics/app.flexible.yaml
+++ b/appengine/analytics/app.flexible.yaml
@@ -11,12 +11,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# [START app_yaml]
 runtime: nodejs
 env: flex
 
-# [START env]
+# [START gae_flex_analytics_env_variables]
 env_variables:
   GA_TRACKING_ID: YOUR_TRACKING_ID
-# [END env]
-# [END app_yaml]
+# [END gae_flex_analytics_env_variables]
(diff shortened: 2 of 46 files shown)


steren:
Can you add a README.md explaining what the sample does and what is happening when `gcp-build` is run?

appengine/uglifyjs/static/index.html
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+-->
+
+<html>
+  <head>
+    <title>Simple web page</title>
+    <script src="message.js" type="text/javascript"></script>
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
```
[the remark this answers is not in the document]
ace-n: I feel like this should be obvious based on the different page URLs (`/index` vs `/index_min`)

appengine/uglifyjs/static/index_min.html
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+-->
+
+<html>
+  <head>
+    <title>Simple web page</title>
+    <script src="message.min.js" type="text/javascript"></script>
+    <link rel="stylesheet" href="style.min.css">
+  </head>
+  <body>
```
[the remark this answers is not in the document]
ace-n: Ditto to below comment.

appengine/typescript/index.ts
```diff
@@ -0,0 +1,35 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* tslint:disable:no-console no-var-requires */
+
+declare var require: any;
```
steren: I think we can do better as a TS example.
This seems to be the copy paste of my little experiement, but does not feel very TS idiomatic: doe example, the Express types should be loaded from npm and I would expect the function to use typed objects.

appengine/typescript/app.yaml
```diff
@@ -0,0 +1,7 @@
+runtime: nodejs8
+
+handlers:
```
steren: remove handlers, they are optional.

appengine/typescript/index.ts
ace-n: Yes, it is largely a copy-paste as I'm not terribly familiar with Typescript 😛

@JustinBeckwith is our resident TS expert, so I'll see what he suggests.

appengine/typescript/app.yaml
ace-n: Done.

steren: Did you push the commit?

appengine/typescript/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Setup
+
+Install dependencies:
+
+    With `npm`:
+
+        npm install
+
+    or with `yarn`:
```
fhinkel: Personal preference, I would not include `yarn` here. Our Node docs say that you can use `npm` or `yarn`, no need to show everything twice here.

ace-n: Removed `yarn` for brevity.

fhinkel:
Thanks for adding tests, can you add them to kokoro, too?

ace-n:
Done - do we need to merge this first?

PR #713, opened 2018-08-17
Fix busboy + GCF race condition

ace-n:
_Note:_ I wanted this to be Node `6.x` compliant, so I avoided using `async`/`await` in favor of Promises.

### DIFF
--- a/functions/http/index.js
+++ b/functions/http/index.js
@@ -166,6 +166,8 @@ exports.uploadFile = (req, res) => {
       fields[fieldname] = val;
     });
 
+    let fileWrites = [];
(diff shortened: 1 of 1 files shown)


functions/http/index.js
```diff
@@ -166,24 +166,42 @@ exports.uploadFile = (req, res) => {
       fields[fieldname] = val;
     });
 
+    let fileWrites = [];
```
ace-n: Nit: should this be `const`?

fhinkel: I think our linter would complain.

PR #718, opened 2018-08-22
Update Region Tags

averikitsch:
Ongoing region tag updates/removals
*Docs are frozen at last commit*


ace-n:
Are the samples without any region tags not used in the docs at all? Or are you trying to include the _whole file_ instead of a specific region?

(If these samples are used somewhere in the docs, I'd argue they should still have region tags - albeit unique ones.)

averikitsch:
Right, the removed region tags have no associated doc (and we don't want the sample tracker picking them up). I'm either updating or removing just the tag.

appengine/endpoints/app.js
```diff
@@ -15,15 +15,12 @@
...
-// [START setup]
 const express = require('express');
 const bodyParser = require('body-parser');
 const Buffer = require('safe-buffer').Buffer;
 
 const app = express();
 app.use(bodyParser.json());
-// [END setup]
 
 app.post('/echo', (req, res) => {
```
ace-n: Is this sample not used in the docs anymore? If so, can/should we delete it outright?

appengine/grunt/Gruntfile.js
```diff
@@ -15,7 +15,6 @@
 
 module.exports = function (grunt) {
   grunt.initConfig({
-    // [START config]
```
ace-n: Ditto to above - are we still using this? If not, can we delete it?

appengine/errorreporting/app.js
```diff
@@ -13,15 +13,12 @@
  * limitations under the License.
  */
 
-// [START app]
```
ace-n: Ditto to above - are we still using this? If not, can we delete it?

appengine/metadata/standard/server.js
```diff
@@ -15,7 +15,6 @@
 
 'use strict';
 
-// [START appengine_metadata]
```
ace-n: Ditto to above - are we still using this? If not, can we delete it?

appengine/mongodb/server.js
```diff
@@ -31,7 +31,6 @@ const pass = nconf.get('mongoPass');
 const host = nconf.get('mongoHost');
 const port = nconf.get('mongoPort');
 
-// [START client]
```
ace-n: Ditto to above - are we still using this? If not, can we delete it?

appengine/parse-server/server.js
```diff
@@ -15,7 +15,6 @@
 
 'use strict';
 
-// [START app]
```
ace-n: Ditto to above - are we still using this? If not, can we delete it?

PR #722, opened 2018-08-24
Added sample for Using Environment Variables.

michaelawyu:


.gitignore
```diff
@@ -10,3 +10,4 @@ logs/
 .nyc_output
 yarn.lock
 package-lock.json
+.DS_Store
```
ace-n: Nit: this should probably be in your [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead of this file.

PR #723, opened 2018-08-24
Add GCF SQL samples

ace-n:
@stew-r FYI

### DIFF
--- a/functions/sql/package.json
+++ b/functions/sql/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "nodejs-docs-samples-functions-sql",
+  "version": "0.0.1",
+  "private": true,
+  "license": "Apache-2.0",
+  "author": "Google Inc.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git"
+  },
+  "engines": {
+    "node": ">=4.3.2"
+  },
+  "scripts": {
+    "start-proxy-mysql": "cloud_sql_proxy -instances=$INSTANCE_CONNECTION_NAME-mysql=tcp:3306 &",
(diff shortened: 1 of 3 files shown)


functions/sql/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "test": "npm run start-proxy && npm run ava && npm run kill-proxy"
+  },
+  "dependencies": {
+    "@google-cloud/nodejs-repo-tools": "^2.3.3",
+    "mysql": "^2.16.0",
+    "pg": "^7.4.3",
+    "proxyquire": "^2.1.0"
+  },
+  "devDependencies": {
+    "ava": "0.25.0",
```
fhinkel: FYI, I think we're trying to go with mocha. But I don't mind having another ava test.

functions/sql/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "lint": "repo-tools lint",
+    "pretest": "npm run lint",
+    "ava": "ava -T 20s --verbose test/*.test.js",
+    "test": "npm run start-proxy && npm run ava && npm run kill-proxy"
+  },
+  "dependencies": {
+    "@google-cloud/nodejs-repo-tools": "^2.3.3",
+    "mysql": "^2.16.0",
+    "pg": "^7.4.3",
+    "proxyquire": "^2.1.0"
```
fhinkel: I think `proxyquire` should be a dev-dependency?

fhinkel: Same for `repo-tools`

functions/sql/package.json
ace-n: @jmdobry and I switched to AVA from Mocha a few years ago (IIRC, because AVA lets us run tests in parallel). Perhaps we can discuss this elsewhere, but what's the reason for switching back?

functions/sql/package.json
ace-n: Done.

PR #724, opened 2018-08-25
Removing `alpha` tag from `gcloud` command

de314:


de314:
I don't care that much.

PR #732, opened 2018-09-06
Add error handling to composer GCF trigger sample

tswast:
Addressed feedback from bug 111401746.

Please review @TrevorEdwards


functions/composer-storage-trigger/test/index.test.js
```diff
@@ -0,0 +1,56 @@
...
+    }),
+    mocks: {
+      fetch: FetchMock,
+      body: body,
+      bodyJson: bodyJson
+    }
+  };
+}
+
+test.cb.serial(`Does nothing for deleted files`, (t) => {
```
ace-n: Can we remove `serial`?

functions/composer-storage-trigger/test/index.test.js
```diff
@@ -0,0 +1,56 @@
...
+  const event = {
+    data: {
+      file: `some-file`
+    }
+  };
+  const expectedMsg = `Something bad happened.`;
+  const sample = getSample();
+  sample.mocks.bodyJson.error = expectedMsg;
+
+  sample.program.triggerDag(event, (err, message) => {
```
ace-n: Should we switch to promises at some point? (In a separate PR, of course.)

tswast: I didn't realize we could return promises with GCF. Yeah, that'd be ideal.

TrevorEdwards:
Thanks for making this. I think we should consider doing retry on failure in the future as well (maybe someone else can PR that?) We've had issues with execution_date timestamp collisions.

PR #736, opened 2018-09-10
Iot fix tests

gguuss:


gguuss:
@fhinkel This should fix / bootstrap the tests.

PR #739, opened 2018-09-12
Fix helloworld GCF test

ace-n:
**Do not merge** until `functions/helloworld` tests pass.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -23,7 +23,6 @@ export FUNCTIONS_BUCKET=my-bucket
 cd github/nodejs-docs-samples/${PROJECT}
 
 # Install dependencies
-npm install -g @google-cloud/functions-emulator
 npm install
 
 # Configure gcloud
@@ -33,6 +32,7 @@ gcloud config set project $GCLOUD_PROJECT
 
 # Start functions emulator, if appropriate
 if [[ $PROJECT == functions/* ]]; then
+  npm install -g @google-cloud/functions-emulator
   functions-emulator start
 fi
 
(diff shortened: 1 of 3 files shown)


ace-n:
*Note to Googlers:* this PR fixes b/114412909

.kokoro/build.sh
```diff
@@ -32,6 +34,10 @@ gcloud config set project $GCLOUD_PROJECT
 
 # Start functions emulator, if appropriate
 if [[ $PROJECT == functions/* ]]; then
+  export FUNCTIONS_LOG_PATH=$(pwd)/logs/cloud-functions-emulator.log
+  npm install -g @google-cloud/functions-emulator
```
fhinkel: Is there a way we could avoid global installs? Would `npx` solve that for us?

.kokoro/build.sh
```diff
@@ -32,6 +34,10 @@ gcloud config set project $GCLOUD_PROJECT
...
+  touch "$FUNCTIONS_LOG_PATH"
```
fhinkel: Is `touch` right here or do you want to delete the file if it's there already?

ace-n: I want to make sure that the file exists.

.kokoro/build.sh
ace-n: Perhaps - however, these containers are (AFAIK) ephemeral. Is making (and potentially debugging) this change a good use of time?

fhinkel: If anybody wants to reproduce this locally, they might end up with global installs which is not nice. Also, problem caused by globally installed dependencies are hard to debug. 

It should be enough to delete the `npm i -g` line and further down call: 
`npx  functions-emulator start` (instead of `functions-emulator start`). 

Let's land it as is to figure out if it works. And then do another revision to get rid of the global installs.

ace-n:
@dpebot merge when green

PR #740, opened 2018-09-14
GCF: Add Stackdriver trigger sample

ace-n:
@stew-r FYI

This sample is designed for BigQuery, but I can make it more generic if people would like.


ace-n:
cc @mogar1980 as well

PR #744, opened 2018-09-22
Copied config file from endpoints-samples/k8s

navinger:
Copied the k8s config file used in the Endpoints Kubernetes tutorial from the endpoints-samples repo. This way the sample code and the config files are all in the same repo.


fhinkel:
I can't merge this because the branch does not exist anymore 😞

PR #746, opened 2018-09-24
Simplify Headless Chrome sample, fix unit test

steren:


steren:
@fhinkel PTAL

steren:
I have no idea why `functions/helloworld` fails, or why it runs...

PR #748, opened 2018-09-24
Add billing samples (Node 8 ONLY)

ace-n:
@labtopia FYI


functions/billing/index.js
```diff
@@ -0,0 +1,156 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_billing_limit]
+// [START functions_billing_stop]
+const {google} = require('googleapis');
+const {auth} = require('google-auth-library');
+
+const PROJECT_ID = process.env.GCP_PROJECT;
```
fhinkel: Shouldn't the Project id be the id? I think GCP_PROJECT is not a variable we usually set

functions/billing/index.js
```diff
@@ -0,0 +1,156 @@
...
+
+// [START functions_billing_slack]
+const slack = require('slack');
+
+const BOT_ACCESS_TOKEN = 'xxxx-111111111111-abcdefghidklmnopq';
+const CHANNEL = 'general';
+
+exports.notifySlack = async (data, context) => {
+  const pubsubMessage = data;
+  const budgetNotificationText = JSON.stringify(pubsubMessage.attributes) +
```
fhinkel: Would template literals look better here? 
`${JSON.stringify(pubsubMessage.attributes)}, ${Buffer.from(pubsubMessage.data, 'base64').toString(}`

functions/billing/index.js
```diff
@@ -0,0 +1,156 @@
...
+  console.log(res);
+};
+// [END functions_billing_slack]
+
+// [START functions_billing_stop]
+const billing = google.cloudbilling('v1').projects;
+
+exports.stopBilling = async (data, context) => {
+  const pubsubData = JSON.parse(Buffer.from(data.data, 'base64').toString());
+  if (pubsubData.costAmount <= pubsubData.budgetAmount) {
```
fhinkel: Do we need to assure that costAmount is a number and not a string? Want to avoid cases like `'100' < '2' = true`

functions/billing/index.js
ace-n: It is [defined by GCF itself](https://cloud.google.com/functions/docs/env-var#reserved_keys_key_validation).

functions/billing/index.js
ace-n: FWIW this is how the code [is currently structured](https://cloud.google.com/billing/docs/how-to/notify#write_a_cloud_function_1).

ace-n:
CI has been added - PTAL?

PR #751, opened 2018-10-02
Change not working socketPath to host in pg config

tsaarikivi:
Tried multiple times and failed with socketPath, but worked with setting host.
Suggestion found here https://stackoverflow.com/a/48826967.


ace-n:
Thanks for the PR @tsaarikivi! This was already addressed in #749, however - so I'll close this in favor of that one.

Let us know if you have any questions!

PR #752, opened 2018-10-05
Fix import of @google-cloud/storage in auth example

mohsen1:


fhinkel:
@mohsen1 Thanks for the PR! This overlapped with another fix, so I'm closing this.

PR #753, opened 2018-10-06
Add Firestore tests + reactive sample

ace-n:


functions/firebase/test/index.test.js
```diff
@@ -0,0 +1,143 @@
+/**
+ * Copyright 2017, Google, Inc.
```
crwilcox: If this is a new file, copyright should be 2018.

PR #754, opened 2018-10-07
Remove engine.node in Standard Hello World

steren:
No need to pin to a Node version for a Hello World.
This will make the sample forward compatible with Node 10


steren:
I realize they are all over the place. Let me send another PR

PR #761, opened 2018-10-10
Correcting function name to match README.md

apekshithr:
README.md invokes a function named `uuid` bug the function in index.js has getUuid which is wrong.


apekshithr:
I signed it!

PR #769, opened 2018-10-12
Make Stackdriver trigger sample Node-version-agnostic

ace-n:


functions/log/test/index.test.js
```diff
@@ -106,10 +104,32 @@ test(`processLogEntry: should process log entry`, (t) => {
     }
   };
 
-  sample.program.processLogEntry(data, callback);
+  sample.program.processLogEntry(data);
```
crwilcox: I am concerned that this line will pass and t.true will fail as the console log hasn't happened yet. This may need to be 

```
sample.program.processLogEntry(data).then(() => {
t.true(...
```

ace-n: `processLogEntry` is a synchronous operation, so that shouldn't be a concern.

PR #770, opened 2018-10-15
Fix functions/billing on Kokoro

ace-n:
**Do not merge** until the `functions/billing` build passes.

- Upgrade _tests_ to Node 8 (samples still work with Node 6)


crwilcox:
- docker image doesn't have gcloud
- There is an issue with where npm is installing (seems to global which user level isn't permitted to do?)

ace-n:
We'll need to add `gcloud` to the appropriate Docker image to get these passing. I've asked @theacodes' team where those are stored.

fhinkel:
@ace-n would be great if we can get this to work.

PR #776, opened 2018-10-20
Fix promises and surface IAP errors in Composer GCF Dag triggering

### LINKED ISSUE #775: Composer GCF Dag Triggering ignores IAP errors.
If IAP returns an error (e.g. permission denied), the DAG triggering function will ignore the error, and return a 200 (OK) response.

TrevorEdwards:
[Fix #775].

This also removes a useless POST request to the webserver and fully promisifies the function so that a callback does not need to be passed in.


TrevorEdwards:
@tswast  please review. I think the billing test failure is a flake.

PR #812, opened 2018-10-30
Fixed Cloud SQL PostgreSQL Tests

michaelawyu:


michaelawyu:
PR cancelled: There seem to be duplicate tests in the system.

PR #815, opened 2018-10-31
NodeJS 6 code sample for Cloud Function to start/stop GCE instances

djmailhot:
To be used in a tutorial for GCP Cloud Scheduler


djmailhot:
Looks like there are breakages in functions/billing and functions/speech-to-speech.  This CL doesn't touch either of those directories, so I hope those breakages won't jeopardize this Pull Request.

fhinkel:
Those two failures are unrelated, we can ignore them.

djmailhot:
kokoro config added

PR #818, opened 2018-11-01
Updated all App Engine standard samples to use Node.js 10 runtime.

michaelawyu:
Excluding Cloud SQL samples (already sent in another PR).


steren:
ping?

PR #825, opened 2018-11-07
Deprecate Node <8 support

ace-n:
#161 makes this repo dependent on Node 8+ - this PR updates `package.json` to reflect that.


JustinBeckwith:
👋 We should probably make this change in *all* the package.json(s)

ace-n:
This was supposed to be a PR to `nodejs-repo-tools`...whoops.

PR #826, opened 2018-11-08
Use image with Node 8 for kokoro jobs

fhinkel:


fhinkel:
This fixes the linter problems.

fhinkel:
@ace-n Thanks for the heads up, these tests should not require `gcloud`.

fhinkel:
`functions/sendgrid` is failing - but I didn't touch the `functions` image. Any idea?

fhinkel:
Same linter problem. Can I make that test non-blocking for now until you figure out what to do with functions? The problem is that the test is using Node 8 syntax on a Node 6 image.

PR #835, opened 2018-11-08
Add Healthcare API Fhir sample

noerog:


fhinkel:
Do you want to rename the region tag to `heathcare-fhir-...`?

noerog:
I think that it makes sense to do healthcare-fhir for the fhir_resources.js file, but in fhir_stores.js the region tags are already FHIR-specific (create_fhir_store, etc.).

PR #839, opened 2018-11-09
Add linter for healthcare/

fhinkel:


JustinBeckwith:
👋 I don't have a problem with this generally, but wouldn't it be a lot easier to just use prettier and eslint at the top level, instead of adding this to every sample?

healthcare/.eslintrc.yml
```diff
@@ -0,0 +1,15 @@
+---
```
JustinBeckwith: How do you feel about having these `.eslintrc.yml` and `prettierrc` files in the root of every sample?  Should we consider just having a single config at the root of the repo, and then having overrides in the individual dirs if we need em?

fhinkel: I have one file at the root and overrides where needed.

PR #888, opened 2018-11-13
Fix the failed Kokoro test.

michaelawyu:


fhinkel:
We should update the docker image instead of updating gcloud on every test run. LGTM as a quick fix.

PR #892, opened 2018-11-13
Remove the trace samples

JustinBeckwith:
These landed over in https://github.com/googleapis/cloud-trace-nodejs/tree/master/samples, so I figure we can drop the ones in this repo.

Do we need to update region tags in cloud.google.com first?


fhinkel:
There's already a PR open. And yes, Region tag updates, first.

PR #893, opened 2018-11-14
Fix .blur call to add both parameters

wu4f:
Blur needs radius and sigma.  .blur(16) barely blurs the image.  I believe .blur(0,16) was the intended behavior.


fhinkel:
Good catch, thanks!

fhinkel:
The linter in `functions/imagemagick` isn't happy. Can you run `npm run lint -- --fix` and push again. Thanks!

fhinkel:
ping @wu4f

wu4f:
Oh..whoops.  Sorry about that.  I missed the previous email.   Committed
the change with the space added after the ','

Thanks!
Wu-chang

PR #906, opened 2018-11-15
[Asset] Add quickstart code for ExportAssets API

peter-zheng-g:


asset/quickstart.js
```diff
@@ -0,0 +1,82 @@
...
+      uri: dumpFilePath
+    }
+  };
+  var request = {
+    parent: projectResource,
+    outputConfig: outputConfig
+  };
+
+  // Handle the operation using the promise pattern.
+  client.exportAssets(request).then(responses => {
```
fhinkel: Since you have an `async` function, you could use `await` instead of promises, which is much nicer: 

```
let responses = await client.exportAssets(request);
var operation = responses[0];
responses = await operation.promise();
var result = responses[0];
```

Feel free to leave out the `catch` in the sample. If the sample throws, we'll see the error either way.

asset/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@google-cloud/pubsub": "*",
+    "@google-cloud/storage": "^2.3.0",
+    "@google-cloud/asset": "^0.1.0",
+    "uuid": "^3.3.2",
+    "yargs": "^12.0.0"
+  },
+  "devDependencies": {
+    "@google-cloud/nodejs-repo-tools": "^3.0.0",
+    "ava": "^0.25.0",
+    "mocha": "^5.2.0",
```
fhinkel: mocha, proxyquire, and ava are not used in your tests, so we don't need them. I think we can leave out nyc, too.

asset/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@google-cloud/asset-samples",
+  "description": "Samples for the Cloud Asset Client Library for Node.js.",
+  "version": "0.0.1",
+  "license": "Apache-2.0",
+  "author": "Google Inc.",
+  "engines": {
+    "node": ">=8"
+  },
+  "repository": {
```
fhinkel: Let's not include `repository`, as this repo has many samples.

fhinkel:
@peter-zheng-g I wasn't paying attention. Since `asset` is a client library, the sample should live there: https://github.com/googleapis/nodejs-asset/tree/master/samples. 

For Node.js, we have the samples in the client library, if there is a client library. Sorry I didn't realize that right away.

peter-zheng-g:
OK... Not sure why it's structured like that way, but it'd be really hard to find quickstart samples for other cloud services.

PR #914, opened 2018-11-15
Delete outdated region tags for error-reporting

fhinkel:
Delete old tags.


fhinkel:
Error-reporting moved to the client lib some months ago.

PR #916, opened 2018-11-15
Delete outdated region tags for trace

fhinkel:
Do not merge until docs are published.


fhinkel:
Not needed because trace samples are in the client library repository.

PR #930, opened 2018-11-25
Update error logging example

Mytrill:
Add an example of logging an error with both text and an error object.


Mytrill:
CLA signed

fhinkel:
@Mytrill The linter is failing. Could you have a look. Thanks.

PR #998, opened 2018-12-04
merge command beta samples into device manager/mqtt example

hongalex:


iot/mqtt_example/cloudiot_mqtt_example_nodejs.js
```diff
@@ -255,8 +255,13 @@ let iatTime = parseInt(Date.now() / 1000);
 let client = mqtt.connect(connectionArgs);
 
 // Subscribe to the /devices/{device-id}/config topic to receive config updates.
+// Config updates are recommended to use QoS 1 (at least once delivery)
 client.subscribe(`/devices/${argv.deviceId}/config`, {qos: 1});
 
+// Subscribe to the /devices/{device-id}/commands/# topic to receive commands
+// commands is recommended to use QoS 0 (at most once delivery)
+client.subscribe(`/devices/${argv.deviceId}/commands/#`, {qos: 0});
```
gguuss: No action required but I'm using QoS 1 on other samples, it will retry until ack.

iot/mqtt_example/cloudiot_mqtt_example_nodejs.js
```diff
@@ -282,10 +287,15 @@ client.on('error', err => {
...
 
 client.on('message', (topic, message) => {
-  console.log(
-    'message received: ',
-    Buffer.from(message, 'base64').toString('ascii')
-  );
+  let messageStr = 'Message received: ';
+  if (topic === `/devices/${argv.deviceId}/config`) {
+    messageStr = 'Config message received: ';
+  } else if (topic === `/devices/${argv.deviceId}/commands`) {
```
gguuss: It may be worth verifying the topic name fully matches what you expect here as we're subscribing with the wild card above.

iot/mqtt_example/system-test/cloudiot_mqtt_example.test.js
```diff
@@ -146,3 +146,23 @@ test(`should send state message`, async t => {
   );
   await tools.runAsync(`${helper} deleteRegistry ${localRegName}`, cwd);
 });
+
+test(`should receive command message`, async t => {
+  const localDevice = `test-rsa-device`;
+  const localRegName = `${registryName}-rsa256`;
```
gguuss: /FYI I have seen the tests run in parallel which could cause flakiness if the names are not distinct per registry+device.

gguuss:
@hongalex looks like there's something failing in the send command test, you should be good to merge once that's passing

iot/mqtt_example/cloudiot_mqtt_example_nodejs.js
hongalex: I actually tested this and using # doesn't actually match here

PR #999, opened 2018-12-04
You need to inject the redis-ip address

jonasfj:
In other samples you need inject the redis-ip address.

I'm not 100% sure, but when reading:
https://cloud.google.com/memorystore/docs/redis/connect-redis-instance-flex

I'm getting the impression that I have to inject the redis IP address, and there is nothing linking into my appengine flex environment on port `6379`. If there was something linking it, then how could I have multiple redis instances?


fhinkel:
Thanks, yes, `localhost` doesn't make sense here.

PR #1001, opened 2018-12-04
Websockets sample + add tests

ace-n:


appengine/websockets/app.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018, Google LLC.
```
andrewsg: nit: we should omit copyright statement for app.yaml files

PR #1021, opened 2018-12-12
Add Node 8 async sample

ace-n:


functions/node8/index.js
```diff
@@ -231,8 +238,6 @@ exports.helloAnalytics = (data, context) => {
 // [END functions_firebase_analytics_node8]
 
 // [START functions_background_promise_node8]
-const requestPromiseNative = require('request-promise-native');
```
fhinkel: Can you use `node-fetch` instead of `request`? It's much more common and what we use in the client libraries.

ace-n: Done.

PR #1022, opened 2018-12-13
Fix {lint, broken deps} for OCR sample

ace-n:
~~**Do not merge** until @ace-n has properly reviewed this.~~


ace-n:
@fhinkel this uses `semistandard` - should I switch it to `prettier` here, or save that for another PR?

fhinkel:
Linter is failing

PR #1104, opened 2019-01-18
Added Cloud SQL MySQL connectivity samples for mysql library. 

kurtisvg:
Adding some base mysql connectivity samples.


cloud-sql/mysql/mysql/README.md
```diff
@@ -0,0 +1,94 @@
...
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/key.json
+export CLOUD_SQL_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<MY-DATABASE>'
+export DB_USER='my-db-user'
+export DB_PASS='my-db-pass'
+export DB_NAME='my_db'
+```
+Note: Saving credentials in environment variables is convenient, but not secure - consider a more
+secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep secrets safe.
+
+## Running locally
```
fhinkel: Seems to be in here twice

cloud-sql/mysql/mysql/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+To launch your browser and view the app at https://[YOUR_PROJECT_ID].appspot.com, run the following
+command:
+```bash
+gcloud app browse
+```
+
+## Deploy to Google App Engine Flexible
+
+First, update `app.standard.yaml` with the correct values to pass the environment
```
fhinkel: shouldn't this be a different yml file?

cloud-sql/mysql/mysql/README.md
kurtisvg: Removed

kurtisvg:
@fhinkel Ready for another pass.

PR #1109, opened 2019-01-28
Fix lint issues

fhinkel:


fhinkel:
Bug opened and assigned to @kurtisvg to fix mysql tests.

PR #1113, opened 2019-01-29
Makes speech-to-speech function compatible with Firebase.

ricalo:
To make the speech-to-speech function compatible with Cloud Functions for Firebase, the following changes are included in this commit:

* Adds `.firebaserc` file.
* Adds `firebase.json` file.
* Imports `firebase-functions` library.


ricalo:
Thanks, @fhinkel 
All the required tests are passing. Let me know if there's something else I need to do to get this PR merged.

PR #1114, opened 2019-01-29
Fix MySQL tests. 

kurtisvg:


kurtisvg:
@fhinkel This should be good to merge.

PR #1118, opened 2019-01-30
Auto-update minor and patch for some dependencies

fhinkel:
Changing a few hand-picke dependencies where we trust the minor and
patch updates. If dependency versions are fixed, we are drowning in
RenovateBot PRs.

### DIFF
--- a/appengine/analytics/package.json
+++ b/appengine/analytics/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "express": "4.16.4",
-    "got": "9.6.0"
+    "got": "^9.6.0"
   },
   "devDependencies": {
     "@google-cloud/nodejs-repo-tools": "^3.0.0"
--- a/appengine/loopback/package.json
+++ b/appengine/loopback/package.json
@@ -21,7 +21,7 @@
     "loopback-component-explorer": "5.4.0",
     "serve-favicon": "2.5.0",
     "strong-error-handler": "3.2.0",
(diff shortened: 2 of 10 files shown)


appengine/loopback/package.json
```diff
@@ -21,7 +21,7 @@
     "loopback-component-explorer": "5.4.0",
     "serve-favicon": "2.5.0",
     "strong-error-handler": "3.2.0",
-    "loopback-datasource-juggler": "4.5.3",
+    "^loopback-datasource-juggler": "4.5.3",
```
alexander-fenster: is this some special syntax, or a typo? I expected `"loopback-datasource-juggler": "^4.5.3"`

fhinkel: Yikes, thanks for noticing.

fhinkel: How did the tests pass though? 🙈

fhinkel:
@alexander-fenster fixed the issue, please have another look.

appengine/analytics/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "express": "4.16.4",
```
JustinBeckwith: Would it be bad to try brining express up to `^` while you're in here?

fhinkel: Only changing the deps that I recently updated, otherwise I'll have too many breaking changes.

PR #1125, opened 2019-02-02
iot: merge gateway management samples from beta into GA

hongalex:
* updated flaky tests in gateway management samples
* configure kokoro test keys


hongalex:
Seems like the only failing test is the slackbot, not sure if its a blocker.

gguuss:
You're free to merge when you're ready, I doubt this CL broke slackbot. You also can take a look at the Slackbot issue and see if it's something you can fix.

PR #1126, opened 2019-02-04
build:  cleanup artifacts.

shollyman:
slackbot example was removed in https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1081 but not all build references were removed.


fhinkel:
There's already a PR open: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1120

PR #1136, opened 2019-02-08
Fix: use regular strings instead of template strings in require

grant:
Replace useless template literal strings with normal string from samples.

Why?

- This causes our samples to have poor syntax highlighting on CloudSite
  - Example: https://cloud.google.com/functions/docs/bestpractices/testing#unit_tests_2
- For `require` statements, this is not very typical.


grant:
@ace-n @hongalex
Can we get this PR looked at?

If we don't want to make this change for some reason, we should close this PR.

ace-n:
I'll defer to @fhinkel on this.

fhinkel:
Can you please change the commit message to `Fix: use regular strings instead of template strings in require`. Thanks.

grant:
> Can you please change the commit message to `Fix: use regular strings instead of template strings in require`. Thanks.

Done.

grant:
@fhinkel Can we proceed in merging this?
Was the only nit the git commit comment?

I don't like leave PRs pending since they'll gather more conflicts.

fhinkel:
@grant I was waiting for you to resolve the merge conflicts. Can't merge it like this. Please also change the PR title. Thanks.

Overall, I'm on the fence merging PRs like this one. They produce a lot of churn and make it harder to follow history without real benefit. We have a bug open for the correct rendering of backticks, the solution is not to avoid template literals.

fhinkel:
Closing as discussed offline.

PR #1173, opened 2019-02-20
Add OpenCensus sample code

mayurkale22:
Node version of https://github.com/GoogleCloudPlatform/golang-samples/pull/646/files.

### DIFF
--- a/opencensus/metrics-quickstart.js
+++ b/opencensus/metrics-quickstart.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 Google, Inc.
+ * Copyright 2019, Google, LLC.
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


mayurkale22:
> Hi, can you provide tests please. Thanks

Done, can you please take a look again. Thanks

fhinkel:
Thanks, we also need the .kokoro config files so the tests will be run.

opencensus/metrics-quickstart.js
```diff
@@ -0,0 +1,92 @@
...
+
+// GOOGLE_APPLICATION_CREDENTIALS are expected by a dependency of this code
+// Not this code itself. Checking for existence here but not retaining (as not needed)
+if (!projectId || !process.env.GOOGLE_APPLICATION_CREDENTIALS) {
+  throw Error('Unable to proceed without a Project ID');
+}
+
+// The minimum reporting period for Stackdriver is 1 minute.
+const exporter = new StackdriverStatsExporter({
+  projectId: projectId,
```
fhinkel: Is the projectId needed? Our client libraries hat the convention to use the `GOOGLE_APPLICATION_CREDENTIALS` implicitly and users should not be required to provide the project id.

mayurkale22: Yes, it is required. Will open an issue and modify example later. Is that ok?

PR #1192, opened 2019-03-01
iot: fix issues with conflicting topics/registries in parallel tests

hongalex:
Per issue b/126683051, revert to UUID for tests/registries.


hongalex:
Seems like iot/manager tests had a similar issue. Will make similar changes to those tests as well.

PR #1198, opened 2019-03-06
iot: add blank rsa test

hongalex:
test PR to debug issue with rsa keys in iot/manager tests failing.

### DIFF
--- a/appengine/building-an-app/update/package.json
+++ b/appengine/building-an-app/update/package.json
@@ -24,6 +24,6 @@
   },
   "devDependencies": {
     "@google-cloud/nodejs-repo-tools": "3.2.0",
-    "mocha": "^6.0.0"
+    "mocha": "^5.2.0"
   }
 }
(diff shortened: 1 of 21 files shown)


fhinkel:
The IOT tests were green on this PR: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1192
Since then, the only change in `/iot` was the mocha upgrade, but IIRC you already checked without that change it they're still failing. 

I made a ton of changes in the repo recently, not sure how they'd affect these tests if the changes were not in the `/iot` folder.

fhinkel:
Oh, but then the problem is with local vs CI runs. I think we should revert https://github.com/GoogleCloudPlatform/nodejs-docs-samples/commit/dae108888bd70309d8cde4c1cb825b13b685ea4b#diff-97c209ca505f2958db30a42135afab5c in this dummy PR and see if CI is green like it was when we merged #1192

hongalex:
Issue was that the x509 certificates were expired. Updated, and closing PR.

PR #1199, opened 2019-03-07
Update Readme for appengine/mongodb

fhinkel:
Update the link to the community tutorial. The old link redirects to this sample.


fhinkel:
@michaelawyu Should we also delete the `app.standard.yaml` since MongoDB on GAE Standard is not good practice?

PR #1219, opened 2019-03-22
Clean up package.json in root directory

fhinkel:
Tests are run in individual subfolders, but we have a global
lint job.
Follow up PR that deletes the coverage output, as it's at 55% and not tracking correctly 😞

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,31 +5,23 @@
   "license": "Apache-2.0",
   "author": "Google Inc.",
   "engines": {
-    "node": ">=4.0.0"
+    "node": ">=8"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git"
   },
-  "main": "src/index.js",
   "private": true,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,31 +5,22 @@
...
     "url": "https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git"
   },
-  "main": "src/index.js",
   "private": true,
   "scripts": {
-    "lint": "eslint '**/*.js'",
-    "pretest": "npm run lint",
-    "unit-cover": "nyc --cache npm test && nyc report --reporter=html",
-    "system-cover": "nyc --cache npm run system-test && nyc report --reporter=html",
-    "test": "npm run unit-test && npm run system-test",
```
JustinBeckwith: Just an idea - but for contributors, would it be nice if we had `test` do a `echo tests must be run in the sample directory` or something like that?

fhinkel: Good idea! Easier than finding that in the docs. Will update.

PR #1223, opened 2019-03-24
Fix billing tests always passing

jonaseck2:
Tests were passing due to incorrectly named json data. Calling the function with an empty data message will disable billing every time


fhinkel:
@ace-n please take a look

Issue #1080, opened 2019-01-07
Convert tests from ava to mocha


AVaksman:
The following list represents all hat currently use `ava` and need to be updated to use `mocha`:

1. - [ ] functions/http/test/index.test.js
2. - [ ] functions/imagemagick/test/index.test.js
3. - [ ] functions/log/test/index.test.js
4. - [ ] functions/node8/test/sample.unit.pubsub.test.js
5. - [ ] functions/node8/test/sample.unit.storage.test.js
6. - [ ] functions/ocr/app/test/index.test.js
7. - [ ] functions/pubsub/test/index.test.js
8. - [ ] functions/scheduleinstance/test/index.test.js
9. - [ ] functions/sendgrid/test/index.test.js
10. - [ ] functions/slack/test/index.test.js
11. - [ ] functions/spanner/test/index.test.js
12. - [ ] functions/sql/test/index.test.js
13. - [ ] functions/tips/test/index.test.js
14. - [ ] functions/uuid/test/index.test.js
15. - [ ] healthcare/datasets/system-test/datasets.test.js
16. - [ ] healthcare/dicom/system-test/dicomweb.test.js
17. - [ ] healthcare/dicom/system-test/dicom_stores.test.js
18. - [ ] healthcare/fhir/system-test/fhir_resources.test.js
19. - [ ] healthcare/fhir/system-test/fhir_stores.test.js
20. - [ ] healthcare/hl7v2/system-test/hl7v2_messages.test.js
21. - [ ] healthcare/hl7v2/system-test/hl7v2_stores.test.js
22. - [ ] iot/beta-features/gateway/gateway.test.js
23. - [ ] iot/http_example/system-test/cloudiot_http_example.test.js
24. - [ ] iot/manager/system-test/manager.test.js
25. - [ ] iot/mqtt_example/system-test/cloudiot_mqtt_example.test.js
26. - [ ] jobs/v2/system-test/quickstart.test.js
27. - [ ] jobs/v3/system-test/auto-complete-sample.test.js
28. - [ ] jobs/v3/system-test/basic-company-sample.test.js
29. - [ ] jobs/v3/system-test/basic-job-sample.test.js
30. - [ ] jobs/v3/system-test/commute-search-sample.test.js
31. - [ ] jobs/v3/system-test/custom-attribute-sample.test.js
32. - [ ] jobs/v3/system-test/email-alert-search-sample.test.js
[...]

PR #1249, opened 2019-04-12
refactor(endpoints/getting-started-grpc): ava to mocha

AVaksman:
towards #1080 

refactor endpoints/getting-started-grpc tests from ava to mocha.


AVaksman:
Looks like these tests are not triggered.

PR #1254, opened 2019-04-17
functions/tips: add run region tags

grayside:
These two snippets are reused in Cloud Run. Adding dedicated region tags for separate tracking.

x-ref: https://github.com/GoogleCloudPlatform/golang-samples/pull/831


fhinkel:
Rerunning the failed test...

PR #1267, opened 2019-05-07
Update Sendgrid sample + add tests

ace-n:
**Note:** this does not include any additional Kokoro configs. @fhinkel happy to add one if you'd like.

(The other option that I think would be cleaner: we update the samples + add tests now, and fix Kokoro in another PR.)


fhinkel:
Fine with me to add Kokoro later

PR #1268, opened 2019-05-08
Update Twilio sample + add (unit) tests

ace-n:
**Note:** @fhinkel I added ESLint dependencies as `devDependencies` to `package.json` - let me know if I shouldn't do that.


fhinkel:
Instead of having the eslint dependencies and configs copied to every sample, it should only be in the root folder and `npm lint` should be run from the root folder, not for individual samples.

PR #1271, opened 2019-05-09
Update storage trigger samples

ace-n:
**Note:** I haven't tested these on GCP, and the local tests are all unit tests.

@fhinkel should I do something more comprehensive?


functions/composer-storage-trigger/index.js
```diff
@@ -50,23 +50,28 @@ exports.triggerDag = function triggerDag(event) {
...
+
     return makeIapPostRequest(
       WEBSERVER_URL,
       BODY,
       iap.idToken,
       USER_AGENT,
       iap.jwt
     );
-  });
+  } catch (err) {
```
fhinkel: It's cleaner if we throw instead of returning a rejected promise:

```
} catch (err) {
  throw new Error(err);
}
```

functions/composer-storage-trigger/index.js
```diff
@@ -76,89 +81,87 @@ function authorizeIap(clientId, projectId, userAgent) {
...
     body: JSON.stringify(body),
-  }).then(res => {
-    if (!res.ok) {
-      return res.text().then(body => Promise.reject(body));
-    }
   });
-}
+
+  if (!res.ok) {
+    return res.text().then(body => Promise.reject(body));
```
fhinkel: Could we write this with `async/await`? In its current form it's hard to read.

PR #1291, opened 2019-05-16
fix: require node.js 8.13.0 and up (http/2)

JustinBeckwith:


ace-n:
@stew-r (GCF PM) has said we should keep the Node 6 `function/` samples around. I asked @fhinkel to sync with him and see what next steps should be.

fhinkel:
considering this is going from `8` to `8.13`, I think we can land this. Removing Node 6 samples is a different discussion.

PR #1310, opened 2019-05-20
Fix error in documentation

samuelli:
This matches the code in `googleapis`. However, it is entirely plausible that this documentation is correct and the `googleapis` function needs to be updated.

See https://github.com/googleapis/google-api-nodejs-client/blob/6817c39728b199b82af5478de5aef2cda51e969c/src/apis/healthcare/v1beta1.ts#L1290-L1364 for implementation function.


fhinkel:
@noerog Could you have a look at this healthcare API question? Thank you!

fhinkel:
Thanks @samuelli . I'll merge this as soon as CI finishes.

fhinkel:
Could you fix the lint problems please :)

PR #1313, opened 2019-05-21
Add Node 8 standards to linter

ace-n:
**Do not merge** until samples have been updated to conform to these standards.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -20,8 +20,8 @@ rules:
   prefer-const: error
   no-var: error
   prefer-arrow-callback: error
-  no-throw-literal: error
-  require-await: error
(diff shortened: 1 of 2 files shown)


fhinkel:
Looks good to me. But we need to update a lot of samples first. I'd prefer if we'd use the same linting rules as the client libraries. 

Also, how does this work with the functions samples that are currently tested on Node 6?

ace-n:
- Copy linting rules from the client libraries - those are less restrictive than these (e.g. `prefer-await-to-then` isn't there, but it's part of our Node 8 upgrade process).
- `functions` samples - I'm following up more on this internally. I might add an exception to these (either via in-file comments, or a lower-level `eslintrc.yml`) until we agree on a path forward.

PR #1321, opened 2019-05-28
fix: add kokoro config for composer-storage-trigger

leahecole:


leahecole:
Test does pass locally so 🤞 that it would work in CI. Just did the edits to the internal CL

ace-n:
Sounds good. To clarify - you'll want to merge the internal CL **before** merging this.

PR #1329, opened 2019-06-03
Update Cloud SQL sample (Node 8 + system tests)

ace-n:
- [x] Node 8
- [x] Unit-only tests --> system tests

(@kurtisvg as FYI. Please **do not review** before Monday morning.)

### DIFF
--- a/cloud-sql/mysql/mysql/server.js
+++ b/cloud-sql/mysql/mysql/server.js
@@ -99,34 +99,25 @@ app.on('listening', async () => {
 // Serve the index page, showing vote tallies.
 app.get('/', async (req, res) => {
   // Get the 5 most recent votes.
-  const recentResultPromise = pool
-    .query(
-      'SELECT candidate, time_cast FROM votes ORDER BY time_cast DESC LIMIT 5'
-    )
-    .then(rows => {
-      return rows;
-    });
+  const recentVotes = await pool.query(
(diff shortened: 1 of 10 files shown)


cloud-sql/mysql/mysql/server.js
```diff
@@ -99,34 +99,25 @@ app.on('listening', async () => {
...
 app.get('/', async (req, res) => {
   // Get the 5 most recent votes.
-  const recentResultPromise = pool
-    .query(
-      'SELECT candidate, time_cast FROM votes ORDER BY time_cast DESC LIMIT 5'
-    )
-    .then(rows => {
-      return rows;
-    });
+  const recentVotes = await pool.query(
```
kurtisvg: These needed to remain awaited below for all 3 to remain concurrent.

ace-n: Aren't they already?

(If you're referring to the `.count` property - I don't believe non-`function` properties can return promises.)

kurtisvg: Currently the sample does `Create, Create, Create -> Await, Await, Await` which means each statement is executing concurrently. This change moves it to a `Create -> Await, Create -> Await, Create -> Await` for each execution, which means each statement is performed serially. (I might be wrong since I'm not super familiar with Node, but that's my understanding of Async/Await)

ace-n: Ah, I see. Will fix. 🙂

PR #1341, opened 2019-06-05
Update IoT samples to Node 8

ace-n:


iot/manager/system-test/manager.test.js
```diff
@@ -335,7 +303,7 @@ it('should list gateways', async () => {
 
 it('should bind existing device to gateway', async () => {
   const gatewayId = `nodejs-test-gateway-iot-${uuid.v4()}`;
-  await tools.runAsync(
+  await tools.runAsyncWithIO(
```
hongalex: At some point we're trying to migrate everything to use child_process instead and away from `nodejs-repo-tools`. This works for now though.

ace-n: The key thing is that the samples should be consistent. (If you want to move away from that package, definitely loop in @fhinkel RE enforcing that throughout the repo - but I'll save that for another thread. 🙂)

PR #1350, opened 2019-06-09
Enhancement in getOauthToken API

santhoshvaddi:
Removed the hard coded values which are projectID, scope and TokenCollection name in database. These values will be read from the getAuthToken API's input params.


functions/TokenGenerator/functions/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "functions",
+  "description": "Cloud Functions for Firebase",
+  "scripts": {
+    "lint": "eslint .",
```
fhinkel: We keep only one lint script in the top level directory.

functions/TokenGenerator/functions/index.js
```diff
@@ -0,0 +1,179 @@
...
+  // Checking that the user is authenticated.
+  if (!context.auth) {
+    // Throwing an HttpsError so that the client gets the error details.
+    throw new functions.https.HttpsError(
+      'failed-precondition',
+      'The function must be called ' + 'while authenticated.'
+    );
+  }
+  const projectID = data["projectID"];
+  scope = data["scope"];
```
nnegrey: Shouldn't be passed to the function, this should be hard-coded. We don't want people to be able to request any `oauth` scope even if our Service Account is limited.

functions/TokenGenerator/functions/index.js
```diff
@@ -0,0 +1,179 @@
...
+
+const admin = require('firebase-admin');
+const functions = require('firebase-functions');
+const http = require('http');
+const https = require('https');
+admin.initializeApp(functions.config().firebase);
+const db = admin.firestore();
+let tokenCollection = 'Tokens';
+const token = 'OauthToken';
+let scope = 'cloud-platform';
```
nnegrey: Should make a comment with like:
```
let scope = 'TODO';  // DEVELOPER TODO: change this to a specific value. 
```

functions/TokenGenerator/functions/index.js
```diff
@@ -0,0 +1,179 @@
...
+exports.getOAuthToken = functions.https.onCall((data, context) => {
+  // Checking that the user is authenticated.
+  if (!context.auth) {
+    // Throwing an HttpsError so that the client gets the error details.
+    throw new functions.https.HttpsError(
+      'failed-precondition',
+      'The function must be called ' + 'while authenticated.'
+    );
+  }
+  const projectID = data["projectID"];
```
nnegrey: I think this might be fine, but should consider the implications if a user can change this variable and what abuse it opens the function to.

functions/TokenGenerator/functions/index.js
```diff
@@ -0,0 +1,179 @@
...
+  scope = data["scope"];
+  tokenCollection = data["tokenCollection"];
```
nnegrey: I'm not sure we want the user to know / set the db token collection either, that shouldn't be client side.

santhoshvaddi: The main intention of enhancing the index.js is to make it more generic. It can be used not only dialogflow but also any project which requires auth token. So the client(iOS or Android) really don’t need to touch or make changes to index.js. By getting rid of these hardcoded values and passing them from the client end we are removing one layer of complexity.  For client passing these values from the iOS/Android side is easier than changing it in the index.js.

nnegrey: In all our samples we expect users to have to configure some stuff before the samples will work. 

I'm fine with making them configure a few settings if it maintains the security of the sample. The less information shared between the client and the server is better.

santhoshvaddi:
@nnegrey Closing this PR as per our conversation.

Issue #834, opened 2018-11-08
Improve version tags in headless chrome sample


fhinkel:
@steren can you fix this? 

Per our guidelines: 

> Region tag should use a product prefix, e.g. "vision_".

Where does the code snippet show up in our docs? Our tool can't find it. 

> Update region tag and link to a new or existing tracked sample.
> Import into Devsite or delete.

fhinkel:
ping @steren

steren:
I will not have time for it, and this needs to be synced with docs changes. Please follow up with @llatif

fhinkel:
@llatif Can you please update the region tags for the headless chrome sample to follow our guidelines. Thanks.

fhinkel:
ping @llatif

grant:
See above PR.

PR #1356, opened 2019-06-11
fix: Fixes GAE headless chrome region tags. See #834

grant:
Fixes https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/834.
See cl/252516607


grant:
See sample 1712.

PR #1361, opened 2019-06-12
functions/billing: Node 8 + system tests

ace-n:


.kokoro/functions/functions-billing.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+      
+export GCLOUD_PROJECT=cdpe-functions-billing-test
+export GCP_PROJECT=$GCLOUD_PROJECT
+export GOOGLE_CLOUD_PROJECT=$GCLOUD_PROJECT
```
fhinkel: Is this env variable being used?

ace-n: I believe it's used elsewhere - I was mainly adding it to save us the trouble if future samples start relying on it.

Issue #1362, opened 2019-06-12
Consider console error instead of error for GCF Composer sample


leahecole:
Related to another internal bug - will link them on internal bug tool

PR #1364, opened 2019-06-12
fix: Add console error messages for better logging

leahecole:
#1362

### DIFF
--- a/functions/composer-storage-trigger/index.js
+++ b/functions/composer-storage-trigger/index.js
@@ -61,7 +61,7 @@ exports.triggerDag = async data => {
       iap.jwt
     );
   } catch (err) {
-    console.error('Error authorizing IAP: ' + err.message);
+    console.error('Error authorizing IAP: ', err.message);
     throw new Error(err);
   }
 };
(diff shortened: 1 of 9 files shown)


leahecole:
Kokoro seems to be stuck so attempting a re-run

functions/composer-storage-trigger/index.js
```diff
@@ -88,6 +88,7 @@ const authorizeIap = async (clientId, projectId, userAgent) => {
   );
   const tokenResponse = await res.json();
   if (tokenResponse.error) {
+    console.error(JSON.stringify(tokenResponse.error));
```
ace-n: Is it worth adding something more explanatory to these, and printing out specific error details?

e.g.
```javascript
console.error('Error in authorizeIap:', tokenResponse.error.message);
```

leahecole: Definitely can do that

functions/composer-storage-trigger/index.js
```diff
@@ -88,6 +89,7 @@ const authorizeIap = async (clientId, projectId, userAgent) => {
   );
   const tokenResponse = await res.json();
   if (tokenResponse.error) {
+    console.error('Error in token reponse:' + tokenResponse.error.message);
```
ace-n: Tiny nit: use a comma instead of `+`? (`console.error` can take two arguments, and this is more idiomatic vs. our other samples.)

leahecole:
Yup - I verified this in GCF @ace-n

PR #1366, opened 2019-06-14
Force mocha to exit after running all tests

michaelawyu:
mocha v4 no longer ends itself when all the tests are finished; with HTTP samples continuing to listen at 8080 the tests will run forever.


ace-n:
Is this the only sample with this problem? (If there are others, please update them as well. 🙂)

PR #1367, opened 2019-06-14
Lint: add 'prefer-template' rule

ace-n:
(Encourages the use of ES6 template strings.)


fhinkel:
Will this change regular strings to templates, too? E.g., in ```require(`dep`)```?

ace-n:
@fhinkel no - that would be the [`quote` rule](https://eslint.org/docs/rules/quotes)

PR #1368, opened 2019-06-14
Firebase: print params

ace-n:
Note: I didn't update the `assert` styles here, since I'm planning to do that in a later PR. (And ideally, I'll find a lint rule for it. 🙂)


ace-n:
@grant it's in the branch name (Googlers: b/122877408)

PR #1371, opened 2019-06-17
Add Healthcare API dataset get/set IAM policy

noerog:


healthcare/datasets/setDatasetIamPolicy.js
```diff
@@ -0,0 +1,70 @@
...
+
+/* eslint-disable no-warning-comments */
+
+'use strict';
+
+function main(
+  projectId = process.env.GCLOUD_PROJECT,
+  cloudRegion = 'us-central1',
+  datasetId,
+  member,
```
fhinkel: Can we add default values or comments for `id`, `member`, and `role` as well please?

PR #1372, opened 2019-06-19
Update Healthcare API Hl7v2, FHIR, and DICOM samples to v1beta1

noerog:
Also split DICOM/FHIR/HL7v2 store functions into their own files. Same with HL7v2 messages. Clean up tests/etc.


fhinkel:
Great! Can you run the linter (in the parent directory, `npm run lint -- --fix`). Thanks

PR #1373, opened 2019-06-19
Add Cloud Run to Nodejs Samples

grayside:
* READMEs
* Blind effort at CI plumbing
* Initial sample to be hosted here: "manual" (client-free) structured logging to Stackdriver


.kokoro/build-with-run.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Register post-test cleanup
+function cleanup {
+  gcloud beta --quiet run services delete "${CLOUD_RUN_SERVICE_NAME}"
+  gcloud --quiet container images delete "${CONTAINER_IMAGE}"
+}
+trap cleanup EXIT
+
+# Run Dockerfile linting
+docker run --rm -i hadolint/hadolint < Dockerfile
```
grayside: Assumes we can run docker commands as part of the build. This is the best linter I've found for Dockerfiles and is best run via Docker.

.kokoro/build-with-run.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Build the service
+gcloud builds submit --tag="${CONTAINER_IMAGE}"
+
+# Deploy the service
+gcloud beta run deploy "${CLOUD_RUN_SERVICE_NAME}" --image="${CONTAINER_IMAGE}"
+
+# Capture the URL
+export CLOUD_RUN_SERVICE_URL=$(gcloud beta run services describe "${CLOUD_RUN_SERVICE_NAME}" --format='value(status.domain)')
```
grayside: FYI: Cloud Run URLs are not deterministic so we need to retrieve the URL from the deployed service.

run/README.md
```diff
@@ -0,0 +1,131 @@
+<img src="https://avatars2.githubusercontent.com/u/2810941?v=3&s=96" alt="Google Cloud Platform logo" title="Google Cloud Platform" align="right" height="96" width="96"/>
+
+# Google Cloud Run Node.js Samples
+
+[Cloud Run][run_docs] stateless containers on a fully managed environment or in your own GKE cluster.
```
ace-n: Nit: Cloud Run _runs_?

Also: worth linking to Docker here, and/or an explanation of what containers are? (e.g. [this page](https://cloud.google.com/containers/))

run/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Samples
+
+|           Sample                |        Description       |     Deploy    |
+| ------------------------------- | ------------------------ | ------------- |
+|[Hello World][helloworld]&nbsp;&#10149; | Quickstart | [<img src="https://storage.googleapis.com/cloudrun/button.svg" alt="Run on Google Cloud" height="30">][run_button_helloworld] |
+|[Manual Logging][manual_logging] | Structured loggin without client library | [<img src="https://storage.googleapis.com/cloudrun/button.svg" alt="Run on Google Cloud" height="30">][run_button_manual_logging] |
```
ace-n: Typo: add g (`structured loggin*g* without client library`)

run/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+For more Cloud Run samples beyond Nodejs, see the main list in the [Cloud Run Samples repository](https://github.com/GoogleCloudPlatform/cloud-run-samples).
```
ace-n: Nit: `Node.js` (add `.`)

.kokoro/build-with-run.sh
```diff
@@ -0,0 +1,63 @@
...
+trap cleanup EXIT
+
+# Run Dockerfile linting
+docker run --rm -i hadolint/hadolint < Dockerfile
+
+# Build the service
+gcloud builds submit --tag="${CONTAINER_IMAGE}"
+
+# Deploy the service
+gcloud beta --quiet run deploy "${CLOUD_RUN_SERVICE_NAME}" --image="${CONTAINER_IMAGE}" --region=us-central1 --allow-unauthenticated
```
grayside: `--allow-unauthenticated` makes the service public, meaning we can defer adding ID Token handling to the test. This is something I mean to add in a follow-up but I can address in this PR if necessary.

.kokoro/run/common.cfg
```diff
@@ -0,0 +1,22 @@
...
+# Download secrets from Cloud Storage.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/nodejs-docs-samples"
+
+# All builds use the trampoline script to run in docker.
+build_file: "nodejs-docs-samples/.kokoro/trampoline.sh"
+
+# Tell the trampoline which build file to use.
+env_vars: {
+    key: "TRAMPOLINE_BUILD_FILE"
+    value: "github/nodejs-docs-samples/.kokoro/build-with-run.sh"
```
grayside: Created this file on the assumption I need it to override the build script to `build-with-run.sh`

run/README.md
```diff
@@ -0,0 +1,131 @@
...
+    ```
+    rm package-lock.json
+    npm install
+    ```
+
+    Because we're using a read/write volume mount, the revised file will be
+    written to the host's local filesystem. Once you exit the container you can
+    add the file to version control.
+
+    4. Exit the container: ^D
```
ace-n: Nit: use `Ctrl-D` here instead of `^D`? (Not everyone may realize that `^` == `Ctrl`)

run/logging-manual/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Google LLC. All rights reserved.
+# Use of this source code is governed by the Apache 2.0
+# license that can be found in the LICENSE file.
+
+# Use the official Node.js 10 image.
+# https://hub.docker.com/_/node
+FROM node:10
```
ace-n: @fhinkel is Node 10 fine here, or should we use Node 8?

run/logging-manual/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Google LLC. All rights reserved.
```
grayside: This file is copied verbatim from the quickstart. Longer term I'm hoping to create a template mechanism so default Dockerfiles in the samples can be maintained in one place.

run/logging-manual/app.js
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019 Google LLC. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+const express = require('express');
+const app = express();
+
+app.get('/', (req, res) => {
+  let project = process.env.GOOGLE_CLOUD_PROJECT;
```
ace-n: Can this be a `const` value?

run/logging-manual/app.js
```diff
@@ -0,0 +1,43 @@
...
+  // Uncomment and populate this variable in your code:
+  // $project = 'The project ID of your Cloud Run service';
+
+  // Build structured log messages as an object.
+  let globalLogFields = {};
+
+  // Add log correlation to nest all log messages beneath request log in Log Viewer.
+  const traceHeader = req.header('X-Cloud-Trace-Context');
+  if (traceHeader && project) {
+    let trace = traceHeader.split('/')[0];
```
ace-n: Use [array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)?

run/logging-manual/index.js
```diff
@@ -0,0 +1,26 @@
...
+const PORT = process.env.PORT || 8080;
+const SERVICE = require('./package').name;
+
+const startServer = () => {
+  app.listen(PORT, () => console.log(`${SERVICE} listening on port ${PORT}`));
+}
+
+if (!process.env.GOOGLE_CLOUD_PROJECT) {
+  metadata.getProjectId()
+    .then(project => {
```
ace-n: Use [`await` instead of `then`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/prefer-await-to-then.md).

(If you run `eslint` from the repo root, this should show up as a `WARNING`. **Please fix those `WARNING`s where possible**, as they will eventually be converted to `ERROR`s.)

run/logging-manual/metadata.js
```diff
@@ -0,0 +1,12 @@
+const request = require('got');
+
+// Load the project ID from GCP metadata server.
+// For more use of GCP Metadata see https://www.npmjs.com/package/gcp-metadata.
+exports.getProjectId = async () => {
+  const METADATA_PROJECT_ID_URL = 'http://metadata.google.internal/computeMetadata/v1/project/project-id';
+  const options = {
+    headers: { 'Metdata-Flavor': 'Google' }
```
ace-n: Typo: `Met*a*data`?

run/logging-manual/test/system.test.js
```diff
@@ -0,0 +1,114 @@
...
+  projectId: process.env.GOOGLE_CLOUD_PROJECT
+});
+
+const getLogEntriesPolling = async (filter, max_attempts) => {
+  const WRITE_CONSISTENCY_DELAY_MS = 10000;
+  const MAX_ATTEMPTS = max_attempts || 8;
+  let entries;
+
+  for (i = 0; i < MAX_ATTEMPTS; i++) {
+    await setTimeoutPromise(WRITE_CONSISTENCY_DELAY_MS);
```
ace-n: Can we use a dedicated retry library here? Maybe [`promise-retry`](https://github.com/IndigoUnited/node-promise-retry) perhaps?

run/logging-manual/test/system.test.js
```diff
@@ -0,0 +1,114 @@
...
+  let entries = await logging.getEntries({
+    filter: preparedFilter,
+    autoPaginate: false,
+    pageSize: 2
+  });
+
+  return entries;
+}
+
+const dateMinutesAgo = (date, min_ago) => {
```
ace-n: Nit: this is probably unnecessary if you use the [`moment`](https://momentjs.com) library

run/logging-manual/test/system.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+    it ('generates Stackdriver logs', async() => {
+      // The latest two log entries for our service in the last 5 minutes
+      // are treated as the result of the previous test.
+      // Concurrency is supporting by distinctly named service deployment per test run.
+      let filter = `resource.labels.service_name="${service_name}" timestamp>="${dateMinutesAgo((new Date), 5)}"`;
+      const entries = await getLogEntriesPolling(filter);
+      entries.forEach(entry => {
+        if (entry.metadata.httpRequest) {
+          requestLog = entry;
```
ace-n: Nit: are these values used for anything within this `describe` block?

(I suspect they _may_ be used in the _subsequent_ one, but personally I feel like that's bad practice. @fhinkel WDYT?)

run/logging-manual/test/system.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+    it('uses "trace" property as valid entry trace ID', () => {
+      assert(sampleLog.metadata.trace, "'trace' lifted to official property");
+      assert(sampleLog.metadata.trace == requestLog.metadata.trace, "`trace` property matches request log trace ID");
+    });
+  });
+
+  describe('Cloud Run Log Metadata', () => {
+    it('has expected label properties', () => {
+      let labels = sampleLog.metadata.labels;
```
ace-n: Nit: use object destructuring (unless these need to use `let` instead of `const`.)
```javascript
const {labels, resource} = sampleLog.metadata;
```

run/logging-manual/test/system.test.js
```diff
@@ -0,0 +1,114 @@
...
+      let resource = sampleLog.metadata.resource;
+
+      assert(labels.instanceId, "'labels.instanceId' found in the log entry");
+      assert(resource.labels.configuration_name, "'resource.labels.configuration_name' found in the log entry");
```
ace-n: Is it worth shortening the assertion descriptions + doing this in a `for` loop instead?

e.g:
```javascript
const desiredProps = ['configuration_name', 'location', 'revision_name', ...]
desiredProps.each((prop) => {
  assert(resource.labels[prop], `${prop} found in log entry`
});
```

.kokoro/build-with-run.sh
ace-n: Thought: maybe we want to have a dedicated `Dockerfile` linting job? I created an issue (#1374) to log this.

(Not something that needs to be addressed in _this_ PR, though. 🙂)

.kokoro/run/common.cfg
```diff
@@ -0,0 +1,22 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Download trampoline resources. These will be in ${KOKORO_GFILE_DIR}
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
```
ace-n: @fhinkel do you know if we need these other properties (besides the custom `env-vars` at the bottom?)

.kokoro/run/common.cfg
ace-n: Yep - you are correct. 🙂

.kokoro/build-with-run.sh
grayside: sgtm, trying to keep my scope to Run for now but happy to see it expanded.

run/logging-manual/Dockerfile
grayside: All the Node samples for Cloud Run were built in Node 10. I see it as advantageous to have them be at the forefront of what we're supporting generally. It's also easy to port just one sample to a forward version and see what happens.

run/logging-manual/test/system.test.js
grayside: If necessary. I don't see a library saving many lines of code for the single time I need to do retries.

grayside: On the other hand, if we wanted to turn this into a logging test helper, whatever makes it generally usable :)

run/logging-manual/test/system.test.js
grayside: Agreed, but this is a fairly simple manipulation and a one-off bit of timestamp math, not worth [2.66 MB](https://packagephobia.now.sh/result?p=moment)

run/logging-manual/test/system.test.js
grayside: Correct, the variables they are assigning to are in the top-level describe but for use in subsequent describes. It is a questionable practice.

These started out by being in before() statements, but I decided that checking the service is live and Stackdriver has logs at all should be test cases in their own right.

run/logging-manual/test/system.test.js
grayside: Nice!

run/logging-manual/test/system.test.js
ace-n: I feel like Node, idiomatically-speaking, favors the use of libraries. @fhinkel thoughts?

run/README.md
grayside: Whoa, I didn't know about that page! Thanks.

run/logging-manual/test/system.test.js
grayside: Here's what that would look like in practice, dynamic is nice but I'm not sure it's all that much clearer:

```
      const {labels, resource} = sampleLog.metadata;
      const labelProps = ['instanceId']

      labelProps.forEach(prop => {
        assert(labels[prop], `${prop} found in log entry`)
      })

      const resourceProps= [
        'configuration_name',
        'location',
        'project_id',
        'revision_name',
        'service_name'
      ];
      resourceProps.forEach(prop => {
        assert(resource.labels[prop], `${prop} found in log entry`)
      });
```

run/logging-manual/test/system.test.js
grayside: I agree that Node tends to favor libraries, but still the question is:

* Is maintaining the updated dependency or maintaining the lines of code more burden?
* Is this use of the library going to drive meaningful community collaboration?
* Is adding the dependency going to appreciably affect build or deploy times? (`promise-retry` seems quite light. `moment` and `gcp-metadata` less so.)

run/logging-manual/test/system.test.js
fhinkel: I've learned that anything that has to do with dates, is never simple 🤣

fhinkel:
LGTM, no opinion on whether to use libraries or a few lines of code for the `retry` and date manipulation. If we wait a little, the next Kokoro run should pick up the jobs for this.

.kokoro/run/common.cfg
fhinkel: We use them in the other config files, I think they're needed. If not, at least we're consistent and can remove them all at once.

run/logging-manual/Dockerfile
fhinkel: Node 10 is fine

run/logging-manual/test/system.test.js
fhinkel: Personally, I prefer fewer libraries for samples. They can be black boxes unless they're well established libraries.

grayside:
Looks like the run/logging-manual job was visible while in progress, but is missing from the list after completion.

grayside:
This test failure indicates gcloud is not available in the container image `gcr.io/cloud-devrel-kokoro-resources/node:10-user`. 

My understanding is that the source of this image is
https://github.com/googleapis/testing-infra-docker/blob/master/node/10-user/Dockerfile, which does not show gcloud. However, the main `build.sh` and `build-with-appengine.sh` scripts both use gcloud.

Am I missing something?

grayside:
The container image used for Kokoro builds has now been updated to include gcloud.

grayside:
Enabled the Cloud Run API for the nodejs-docs-samples-tests project.

grayside:
Two passes in a row. Currently awaiting another round of review from Ace, I have no further changes in progress.

PR #1375, opened 2019-06-21
functions/datastore Node 8 upgrade

ace-n:
As title.

@fhinkel (and @grant) please review _no sooner than Friday morning_. 🙂

### DIFF
--- a/functions/datastore/test/index.test.js
+++ b/functions/datastore/test/index.test.js
@@ -25,7 +25,6 @@ const execPromise = require('child-process-promise').exec;
 const sinon = require('sinon');
 
 const FF_TIMEOUT = 3000;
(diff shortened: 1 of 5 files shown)


functions/datastore/test/index.test.js
```diff
@@ -33,178 +44,270 @@ const VALUE = {
...
+
+    after(async () => {
+      await handleLinuxFailures(ffProc);
+    });
+
+    it('del: Fails without a key', async () => {
+      const req = {
+        body: {},
+      };
+
```
grant: Nit: Remove unnecessary \n. We can put `const` declarations together. Even putting the `const` declarations next to the `assert` statements looks fine like `del: Deletes an entity`.

PR #1403, opened 2019-07-10
chore: update save-buffer dependency

fhinkel:


functions/pubsub/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "@google-cloud/pubsub": "^0.28.1",
-    "safe-buffer": "5.2.0"
+    "safe-buffer": "^5.2.0"
```
JustinBeckwith: so we don't have to do this here, but I think we've dropped our usage of safe buffer elsewhere.  Do you think this is something we should keep using?  (I honestly don't know).  Also adding @bcoe for thoughts

bcoe: As `Buffer.from` and its like has been available since Node 6, I'd argue it's probably worth dropping.

fhinkel: Oh, good call. :sweat_smile:

Issue #1247, opened 2019-04-07
App engine hello world example too many dependencies


randomekek:
Hello, the hello app engine hello world seems to depend on express. Please remove the dependency, as hello world should always be as succinct as possible for pedagogical purposes!

Is google-cloud/nodejs-repo-tools even strictly necessary?

```
const http = require('http');
const PORT = process.env.PORT || 8080;

const server = http.createServer((request, response) => {
  console.log(request.url);
  response.end('Hello Node.js Server!');
});

server.listen(PORT, err => {
  console.log(`App listening on port ${PORT}`);
  console.log('Press Ctrl+C to quit.');
});
```

fhinkel:
I agree, `npm install` takes a long time considering this is just hello-world. 

We use `@google-cloud/nodejs-repo-tools` for the tests. We're in the process of simplifying this.

QuantumInformation:
Also points to deprecated  npm install -g @google-cloud/functions-emulator
 and doesn't inform you how to set the project: ie

```
➜  helloworld git:(master) gcloud functions deploy helloGET --runtime nodejs8 --trigger-http

ERROR: (gcloud.functions.deploy) Error parsing [name].
The [function] resource is not properly specified.
Failed to find attribute [project]. The attribute can be set in the following ways:
- provide the argument [--project] on the command line
- set the property [core/project]
➜  helloworld git:(master)
```

QuantumInformation:
But thanks for these examples, glad I don't have to depend on Serverless framework now, just another layer than can go wrong.

grant:
@ace-n, can you look at the `@google-cloud/functions-emulator` dependency? Is that effort tracked somewhere?

ace-n:
The closest thing to a tracking issue is (Google internal) b/134942934, but given the fragmentation of this work (lots of minor changes in many places) fully tracking each piece of it would add (IMHO) significant overhead. (Once we've fully updated the `functions` tests [which I'm working on doing this week], it would probably only take 1-2 reasonable-sized PRs.)

This is also _vaguely_ on my radar. My task queue is currently **100% full** however, so it'll likely be at least a few weeks before I have time to look at this. (You're welcome to beat me to it, of course.)

fhinkel:
I'm working on removing `nodejs-repo-tools` for testing `hello-world`

PR #1416, opened 2019-07-16
Use supertest instead of repo-tools test app

fhinkel:
Trying to remove `repo-tools`. Will send a follow up PR that removes `repo-tools test deploy`. 

Ref: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/1247

### DIFF
--- a/appengine/hello-world/standard/package.json
+++ b/appengine/hello-world/standard/package.json
@@ -15,7 +15,7 @@
   "scripts": {
     "deploy": "gcloud app deploy",
     "start": "node app.js",
-    "system-test": "repo-tools test app",
+    "system-test": "mocha --exit test/*.test.js",
     "test": "npm run system-test",
     "e2e-test": "repo-tools test deploy"
   },
@@ -23,7 +23,9 @@
     "express": "^4.16.3"
   },
   "devDependencies": {
-    "@google-cloud/nodejs-repo-tools": "^3.3.0"
+    "@google-cloud/nodejs-repo-tools": "^3.3.0",
+    "mocha": "^6.1.4",
+    "supertest": "^4.0.2"
   },
   "cloud-repo-tools": {
     "test": {
--- a/appengine/hello-world/standard/app.js
+++ b/appengine/hello-world/standard/app.js
@@ -34,3 +34,5 @@ app.listen(PORT, () => {
   console.log('Press Ctrl+C to quit.');
 });
 // [END gae_node_request_example]
+
+module.exports = app;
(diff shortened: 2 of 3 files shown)


fhinkel:
@bcoe Can you review this?

appengine/hello-world/standard/package.json
```diff
@@ -15,15 +15,17 @@
   "scripts": {
     "deploy": "gcloud app deploy",
     "start": "node app.js",
-    "system-test": "repo-tools test app",
+    "system-test": "mocha --exit test/*.test.js",
```
bcoe: fun idea which you can take or leave, if you add `c8` as a dependency, it turns on Node's V8 coverage collection, so the user gets:

coverage

also understand if you think this is too much information for the user all at once.

appengine/hello-world/standard/package.json
```diff
@@ -15,15 +15,17 @@
...
     "test": "npm run system-test",
     "e2e-test": "repo-tools test deploy"
   },
   "dependencies": {
     "express": "^4.16.3"
   },
   "devDependencies": {
-    "@google-cloud/nodejs-repo-tools": "^3.3.0"
+    "@google-cloud/nodejs-repo-tools": "^3.3.0",
```
bcoe: I think we can delete this dependency now?

fhinkel: No :) It's used in the e2e-test. But turns out, we're not calling 2e2-test from CI, instead we have special test files that run a deployment. But I'll prefer deleting it in a separate PR. There's a lot of magic happening in `repo-tools` based on env variables.

appengine/hello-world/standard/package.json
fhinkel: Trying to keep dependencies as light as possible.

fhinkel: But lol, that coverage looks good!

JustinBeckwith: The `--exit` always makes me nervous, but it's not really blocking

fhinkel: These tests don't finish without `--exit`

bcoe: The default for `--exit` was changed relatively recently in mocha:

https://github.com/mochajs/mocha/pull/3034

I think it's reasonable to go back to the old default ... alternatively you'd need to capture the `server` object returned by `app.listen` and then call `.close()` on it when tests are complete... sounds like a hassle :stuck_out_tongue:

fhinkel: So by using `--exit` I'm just using Mocha the way it used to be, right?

PR #1419, opened 2019-07-17
Migrate Sample: Cloud Run as Pub/Sub Push Handler

grayside:
This sample has been previously reviewed in previous source control. Changes include updating dependencies and switching from ava to mocha.

This sample is currently in use in http://cloud.google.com/run/docs/tutorials/pubsub


grayside:
The test output from this sample is currently impacted by https://github.com/GoogleCloudPlatform/nodejs-repo-tools/issues/214

run/pubsub/app.js
```diff
@@ -0,0 +1,44 @@
+// Copyright 2019 Google LLC. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// [START run_pubsub_server_setup]
+
```
ace-n: Nit: extra space in region tag area?

run/pubsub/app.js
```diff
@@ -0,0 +1,44 @@
...
+app.use(bodyParser.json());
+
+// [END run_pubsub_server_setup]
+
+// [START run_pubsub_handler]
+
+app.post('/', (req, res) => {
+  if (!req.body) {
+    const msg = 'no Pub/Sub message received';
+    console.error(`error: ${msg}`);
```
ace-n: Nit: in an ideal world, lines 19-20 wouldn't be repeated in the clause below (lines 26-27) - but I don't think there's a good way to avoid this repetition here. 🙁

run/pubsub/test/app.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+'use strict';
+
+const assert = require('assert');
+const path = require('path');
+const tools = require('@google-cloud/nodejs-repo-tools');
+const uuid = require('uuid');
+
+const cwd = path.join(__dirname, '../');
+const requestObj = tools.getRequest({cwd});
```
ace-n: @fhinkel this uses `nodejs-repo-tools` - I assume that's OK for this sample, since it's not a simple hello-world?

run/pubsub/app.js
grayside: Ah, go standards transference. Fixing.

run/pubsub/test/app.test.js
fhinkel: We're trying to remove the usage of `repo-tools`. Can you use `request` or `fetch` instead of `tools.getRequest`?

fhinkel: You can have a look at the appengine Hello World example, we use `supertest` for this kind of testing. Since we're not running e2e tests there's no need to install `repo-tools`. But not blocking, I'll work on removing `repo-tools` from the samples anyways.

grayside: Switched to sinon and supertest.

grayside:
@fhinkel why not commit package-lock.json? Without that the builds are not deterministic and I need to switch the Dockerfile back to npm install instead of following the guidance on https://nodejs.org/de/docs/guides/nodejs-docker-webapp/ to steer a production-centric build to use npm ci.

grayside:
Not sure why the test is failing. The [build log](https://source.cloud.google.com/results/invocations/39ede3eb-9fc5-49e6-9ff7-0060fbd163ab/targets/cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Frun%2Fpubsub/log) indicates it should be passing

fhinkel:
In production you need `package-lock.json`. But for samples, I prefer no lock file. If there are breaking changes in the dependencies, our tests should catch that and we can fix it. If we rely on specific versions, we end up with very old dependencies. 

Lock files other cause merge conflicts and I see to benefit checking them in for samples. 

If we decide to change this, we should do it consistently for all samples and not just this one.

fhinkel:
Failing tests see https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/1435

grayside:
Tests passed

grayside:
@fhinkel looks like I've addressed all current concerns, could you approve or clarify remaining issues?

fhinkel:
@grayside `internal CI (run/logging-manual)` is still failing. It passed on a [docs change](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1433) I landed a minute ago. Did this PR break that test again?

fhinkel:
Landing and making `run/pubsub` a mandatory test 🎉

PR #1431, opened 2019-07-25
Cloud Functions: ImageMagick: Explain object var

grayside:
The reassignment of the event parameter to const object seems superfluous without realizing its meant to help the developer read the following lines as an interaction with a Cloud Storage object. Adding a comment to establish this.


grant:
I don't see the need for this PR as reassigning identifiers is common. Is there a reported issue that warrants this change?

Should we add comments to all param reassignments rather than just this one?

PR #1444, opened 2019-08-09
functions/imagemagick: fix "callback must be a function" error

grayside:
When I run the index.js code from the Functions imagemagick sample in Cloud Run on a Node 10 container, I get the following error:

```
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function 
     at makeCallback (fs.js:136:11) 
     at Object.unlink (fs.js:943:14) 
     at blurImage (/usr/src/app/image.js:106:23) 
     at process._tickCallback (internal/process/next_tick.js:68:7) 
```

Looking at the current promisify implementation around fs.unlink, it differs from what I've read in trying to return the result of the promisify. Taking a note from https://dev.to/mrm8488/from-callbacks-to-fspromises-to-handle-the-file-system-in-nodejs-56p2, with this change I do not see any errors.

### DIFF
--- a/functions/imagemagick/index.js
+++ b/functions/imagemagick/index.js
@@ -103,6 +103,7 @@ const blurImage = async (file, blurredBucketName) => {
   }
 
   // Delete the temporary file.
-  return promisify(fs.unlink(tempLocalPath));
+  const unlink = promisify(fs.unlink);
+  return unlink(tempLocalPath);
 };
 // [END functions_imagemagick_blur]


functions/imagemagick/index.js
```diff
@@ -103,6 +103,7 @@ const blurImage = async (file, blurredBucketName) => {
   }
 
   // Delete the temporary file.
-  return promisify(fs.unlink(tempLocalPath));
+  const unlink = promisify(fs.unlink);
+  return unlink(tempLocalPath);
```
grant: Please don't use the npm package`util's `promisify` since you can use the native Node `fsPromises.unlink`.

Remove
```js
const {promisify} = require('util');
```

Add
```js
const fsPromises = require('fs').promises;
```

https://nodejs.org/api/fs.html#fs_fspromises_unlink_path

ace-n: The docs say `fs.promises` is only available in Node 10+ - AFAIK these samples (still) need to support Node 8.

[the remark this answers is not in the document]
grayside: I didn't test that, but it seems likely to work.
My reasoning is that this is more explicit, happy to change if you prefer.

grayside: Yes, this is why I didn't use it.

grant: Got it. Disregard the suggestion then.

fhinkel: Yes, `fs.promises` is not supported in Node 8, and it's experimental in Node 10.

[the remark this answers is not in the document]
fhinkel: I like explicit samples 👍

PR #1445, opened 2019-08-09
run/image-processing: add sample

grayside:
This is a new sample demonstrating how Cloud Run can be triggered by Pub/Sub to perform work with GCS for storage, Cloud Vision integration, and command-line utilities.

A few considerations:
* This sample explicitly copies in the functions/imagemagick/index.js. It is oriented this way in the notion of directly sharing the code, but the current implementation does this using manual copy operations.
* I've applied #1444 to the copied code in this PR.
* Because of the copying behavior, I've not yet committed to any kind of test coverage for the logic covered in the GCF tests.

### Reviewer Actions

In addition to the normal review of this sample, I would like feedback on whether a more automated approach to code sharing would be reasonable at this time. I anticipate more of this kind of sharing and cross-product reuse to come up, especially as Function Frameworks grow.

fhinkel: Code sharing general feedback
grant: Cloud Run sample review
ace: imagemagick sharing/usage

Node.js practices by anyone.

### DIFF
--- a/run/image-processing/.env
+++ b/run/image-processing/.env
@@ -1,0 +1,2 @@
+export INPUT_BUCKET_NAME=adamross-svls-kibble
+export BLURRED_BUCKET_NAME=adamross-svls-kibble-output
--- a/run/image-processing/Dockerfile
+++ b/run/image-processing/Dockerfile
@@ -6,7 +6,7 @@
 # https://hub.docker.com/_/node
 FROM node:10
 
-# [START run_imageproc_imagemagick_ubuntu]
+# [START run_imageproc_dockerfile_imagemagick]
 
 # Install Imagemagick into the container image.
 # For more on system packages review the system packages tutorial.
@@ -16,7 +16,7 @@ RUN set -ex; \
   apt-get -y install imagemagick; \
   rm -rf /var/lib/apt/lists/*
 
-# [END run_imageproc_imagemagick_ubuntu]
+# [END run_imageproc_dockerfile_imagemagick]
 
 # Create and change to the app directory.
 WORKDIR /usr/src/app
--- a/run/image-processing/Dockerfile
+++ b/run/image-processing/Dockerfile
@@ -2,11 +2,11 @@
 # Use of this source code is governed by the Apache 2.0
 # license that can be found in the LICENSE file.
 
-# [START run_pubsub_dockerfile]
-
 # Use the official Node.js 10 image.
 # https://hub.docker.com/_/node
 FROM node:10
+
+# [START run_imageproc_imagemagick_ubuntu]
 
 # Install Imagemagick into the container image.
 # For more on system packages review the system packages tutorial.
@@ -15,6 +15,8 @@ RUN set -ex; \
   apt-get -y update; \
   apt-get -y install imagemagick; \
   rm -rf /var/lib/apt/lists/*
+
+# [END run_imageproc_imagemagick_ubuntu]
 
 # Create and change to the app directory.
 WORKDIR /usr/src/app
@@ -34,5 +36,3 @@ COPY . .
 
 # Run the web service on container startup.
 CMD [ "npm", "start" ]
-
-# [END run_pubsub_dockerfile]
--- a/run/image-processing/Dockerfile
+++ b/run/image-processing/Dockerfile
@@ -1,0 +1,38 @@
+# Copyright 2019 Google LLC. All rights reserved.
+# Use of this source code is governed by the Apache 2.0
+# license that can be found in the LICENSE file.
+
+# [START run_pubsub_dockerfile]
+
+# Use the official Node.js 10 image.
+# https://hub.docker.com/_/node
+FROM node:10
+
+# Install Imagemagick into the container image.
+# For more on system packages review the system packages tutorial.
+# https://cloud.google.com/run/docs/tutorials/system-packages#dockerfile
+RUN set -ex; \
+  apt-get -y update; \
+  apt-get -y install imagemagick; \
+  rm -rf /var/lib/apt/lists/*
+
+# Create and change to the app directory.
+WORKDIR /usr/src/app
+
+# Copy application dependency manifests to the container image.
+# A wildcard is used to ensure both package.json AND package-lock.json are copied.
+# Copying this separately prevents re-running npm install on every code change.
+COPY package*.json ./
+
+# Install dependencies.
+RUN npm install --production
+# If you add a package-lock.json, speed your build by switching to 'npm ci'.
+# RUN npm ci --only=production
+
+# Copy local code to the container image.
+COPY . .
+
+# Run the web service on container startup.
+CMD [ "npm", "start" ]
+
+# [END run_pubsub_dockerfile]
--- a/run/image-processing/README.md
+++ b/run/image-processing/README.md
@@ -1,0 +1,29 @@
+# Cloud Run Image Processing Sample
+
+This sample service applies [Cloud Storage](https://cloud.google.com/storage/docs)-triggered image processing with [Cloud Vision API](https://cloud.google.com/vision/docs) analysis and ImageMagick transformation.
+
+Use it with the [Image Processing with Cloud Run tutorial](http://cloud.google.com/run/docs/tutorials/image-processing).
+
+For more details on how to work with this sample read the [Google Cloud Run Node.js Samples README](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/run).
+
+## Dependencies
+
+* **express**: Web server framework
+* **body-parser**: express middleware for request payload processing
+
+## Environment Variables
+
+* `INPUT_BUCKET_NAME`: The Cloud Run service will be notified of images uploaded to this Cloud Storage bucket. The service will then retreive and process the image.
+* `BLURRED_BUCKET_NAME`: The Cloud Run service will write blurred images to this Cloud Storage bucket.
+
+## Maintenance Note
+
+* The `image.js` file is copied from the [Cloud Functions ImageMagick sample `index.js`](../../functions/imagemagick/index.js). Region tags are changed.
+* The package.json dependencies used in the copied code should track the [Cloud Functions ImageMagick `package.json`](../../functions/imagemagick/package.json)
+
+```sh
+cp ../../functions/imagemagick/index.js image.js
+sed -i '' 's/functions_imagemagick_setup/run_imageproc_handler_setup/' image.js
+sed -i '' 's/functions_imagemagick_analyze/run_imageproc_handler_analyze/' image.js
+sed -i '' 's/functions_imagemagick_blur/run_imageproc_handler_blur/' image.js
+```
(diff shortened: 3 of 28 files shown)


run/image-processing/.env
```diff
@@ -0,0 +1,2 @@
+export INPUT_BUCKET_NAME=adamross-svls-kibble
```
fhinkel: I'm assuming this will change.

run/image-processing/README.md
```diff
@@ -0,0 +1,29 @@
...
+Use it with the [Image Processing with Cloud Run tutorial](http://cloud.google.com/run/docs/tutorials/image-processing).
+
+For more details on how to work with this sample read the [Google Cloud Run Node.js Samples README](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/run).
+
+## Dependencies
+
+* **express**: Web server framework
+* **body-parser**: express middleware for request payload processing
+
+## Environment Variables
```
fhinkel: Would it makes sense to explain how to set environment variables?

run/image-processing/app.js
```diff
@@ -0,0 +1,62 @@
+// Copyright 2019 Google LLC. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+const express = require('express');
+const bodyParser = require('body-parser');
+const app = express();
+
+app.use(bodyParser.json());
```
fhinkel: I would include the `require`s inside the region tag.

fhinkel: If you only want to show `post('/')` in the sample, that's fine. But since you do full error handling in that function, it seems that you can also show the `require`s.

run/image-processing/app.js
```diff
@@ -0,0 +1,62 @@
...
+  if (!req.body.message || !req.body.message.data) {
+    const msg = 'invalid Pub/Sub message format';
+    console.error(`error: ${msg}`);
+    res.status(400).send(`Bad Request: ${msg}`);
+    return;
+  }
+
+  // Decode the Pub/Sub message.
+  const pubSubMessage = req.body.message;
+  pubSubMessage.data = Buffer.from(pubSubMessage.data, 'base64')
```
fhinkel: I wouldn't overwrite `pubSubMessage.data`. I'd probably assign the value to `data`, and then re-assign in the `json()` call. But maybe I'm missing something and it makes sense to change the `pubSubMessage`

run/image-processing/app.js
```diff
@@ -0,0 +1,62 @@
...
+    console.error(`error: ${msg}`);
+    res.status(400).send(`Bad Request: ${msg}`);
+    return;
+  }
+
+  try {
+    await image.blurOffensiveImages(data);
+    res.status(204).send();
+  } catch (err) {
+    console.error('error: Blurring image:', err);
```
fhinkel: Do you want to use backticks with template strings like in the error messages above?

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+'use strict';
+
+// [START run_imageproc_handler_setup]
+const gm = require('gm').subClass({imageMagick: true});
+const fs = require('fs');
+const {promisify} = require('util');
+const path = require('path');
+const {Storage} = require('@google-cloud/storage');
+const storage = new Storage();
+const vision = require('@google-cloud/vision').v1p1beta1;
```
fhinkel: Any reason we can't use the default version?

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+
+'use strict';
+
+// [START run_imageproc_handler_setup]
+const gm = require('gm').subClass({imageMagick: true});
+const fs = require('fs');
+const {promisify} = require('util');
+const path = require('path');
+const {Storage} = require('@google-cloud/storage');
+const storage = new Storage();
```
fhinkel: I'd list all `requires` before other variables.

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+
+const client = new vision.ImageAnnotatorClient();
+
+const {BLURRED_BUCKET_NAME} = process.env;
+// [END run_imageproc_handler_setup]
+
+// [START run_imageproc_handler_analyze]
+// Blurs uploaded images that are flagged as Adult or Violence.
+exports.blurOffensiveImages = async event => {
+  // This event represents the triggering Cloud Storage object.
```
fhinkel: Why not use `object` in the function signature instead of reassigning it? Looking at the call site, we use `data`. Maybe change the parameter to `data` or `imageData`?

fhinkel: Maybe `data` isn't the greatest name. But I'd not re-assign immediately.

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+  const filePath = `gs://${object.bucket}/${object.name}`;
+
+  console.log(`Analyzing ${file.name}.`);
+
+  try {
+    const [result] = await client.safeSearchDetection(filePath);
+    const detections = result.safeSearchAnnotation || {};
+
+    if (
+      detections.adult === 'VERY_LIKELY' ||
```
fhinkel: Do you want to add a comment here what the other levels are?

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+      detections.violence === 'VERY_LIKELY'
+    ) {
+      console.log(`Detected ${file.name} as inappropriate.`);
+      return blurImage(file, BLURRED_BUCKET_NAME);
+    } else {
+      console.log(`Detected ${file.name} as OK.`);
+    }
+  } catch (err) {
+    console.error(`Failed to analyze ${file.name}.`, err);
+    return Promise.reject(err);
```
fhinkel: Usually you don't want to manually create `promises` when you have `async/await`. It's more idiomatic with `async/await` to throw instead. 
```
} catch (err) {
    console.error(`Failed to analyze ${file.name}.`, err);
    throw err; // Or throw new Error(`Failed to analyze ${file.name}.: ${err}`);
}
```

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+// [END run_imageproc_handler_analyze]
+
+// [START run_imageproc_handler_blur]
+// Blurs the given file using ImageMagick, and uploads it to another bucket.
+const blurImage = async (file, blurredBucketName) => {
```
fhinkel: Don't you have `Blurred_bucket_name` as a global variable? Sorry, might be missing it in the GitHub view.

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+  const tempLocalPath = `/tmp/${path.parse(file.name).base}`;
+
+  // Download file from bucket.
+  try {
+    await file.download({destination: tempLocalPath});
+
+    console.log(`Downloaded ${file.name} to ${tempLocalPath}.`);
+  } catch (err) {
+    console.error('File download failed.', err);
+    return Promise.reject(err);
```
fhinkel: throw instead of promise

run/image-processing/image.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  await new Promise((resolve, reject) => {
+    gm(tempLocalPath)
```
fhinkel: can you use `promisify` on `gm`?

run/image-processing/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "nodejs-image-processing",
+  "version": "1.0.0",
```
fhinkel: I think if the package is private, you don't need a version. Private means it can't be published.

fhinkel:
**Code sharing general feedback**

I'd say for samples, copy and paste duplication is better than a special setup for the sake of DRY.

Pros:
- For automated syntax updates, it doesn't matter if the tool updates one or two files.
- If something breaks, CI will alert us in both places and it's not that big a deal to fix it twice. 
- When it's time to archive a sample, we can easily do so without figuring out if other places rely on the code.
- Clear ownership, self contained sample.

Cons:
- I just reviewed code that's already checked in. That was unnecessary - or somebody should make the same improvements to the other sample. But then the whole sample might need changes to fit the style. 🤔 

cc @bcoe if they have thoughts

run/image-processing/app.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  // Decode the Pub/Sub message.
+  const pubSubMessage = req.body.message;
+  pubSubMessage.data = Buffer.from(pubSubMessage.data, 'base64')
+    .toString()
+    .trim();
+  let data;
+  try {
+    data = JSON.parse(pubSubMessage.data);
```
bcoe: I agree with @fhinkel I would just decode and parse in this same try, with the error:

```
Invalid Pub/Sub message: data is not valid base64 encoded JSON
``

File: run/image-processing/.env (hunk 1 above)
Comment (grayside) [reply run/image-processing/.env]: Hm, something happened to my global gitignore that this file got in. Will remove.

File: run/image-processing/README.md (hunk 2 above)
Comment (grayside) [reply run/image-processing/README.md]: How standalone should the README be from the tutorial?

File: run/image-processing/image.js (hunk 7 above)
Comment (grayside) [reply run/image-processing/image.js]: Per my PR notes, the image.js file is literally functions/imagemagick/index.js, copied over and with region tags replaced. My goal is not to fork it, hence my previous proposal of uniting them around a code templating tool. Can we move this feedback to a separate issue to the GCF code?

(And of course, feedback welcome on how to approach this idea of shared code across samples)

File: run/image-processing/package.json (hunk 14 above)
Comment (grayside) [reply run/image-processing/package.json]: I think it's important for developers to think about version numbers for all projects even if not publishing, but I can't say I thought about this one. As a practice what do we prefer to do in samples on version and private?

File: run/image-processing/Dockerfile [hunk 16]
```diff
@@ -0,0 +1,38 @@
...
+# [START run_pubsub_dockerfile]
+
+# Use the official Node.js 10 image.
+# https://hub.docker.com/_/node
+FROM node:10
+
+# Install Imagemagick into the container image.
+# For more on system packages review the system packages tutorial.
+# https://cloud.google.com/run/docs/tutorials/system-packages#dockerfile
+RUN set -ex; \
```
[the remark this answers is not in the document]
grayside: We also use it in functions/imagemagick, from which the image.js file in this sample is copied. It is also included in App Engine and Cloud Functions runtimes.

grayside:
I am now looking at applying the feedback here for image.js to functions/imagemagick/index.js to decouple the improvements.

run/image-processing/package.json
fhinkel: I agree that version numbers are important. In our samples though we never paid much attention to them and I'd prefer not having them than having meaningless numbers.

fhinkel:
Should I go ahead and merge this? I suggest we duplicate the code for now and merge if we have a discussion and decide that we don't like the code duplication. Or would it be easier to start with shared resources and copy if we prefer simplicity over DRY?

PR #1456, opened 2019-08-22
Functions: ImageMagick tutorial: Idiomatic & Robustness Improvements

grayside:
This PR is intended as a spot for some broad improvements to the ImageMagick sample after feedback in the duplicated code in #1445. Once this PR is in, the code in the other can be updated. We can knock some stuff around but let's focus on getting something ready with a ~1 week max turnaround.

### test/index.test.js
* Only clean up blurred image if it exists
* Only run tests if source files exist in Cloud Storage.

### index.js
* Promise.reject() => throw new Error
* Re-order require statements
* Use stable/default version of the client library

## Other Feedback Not Yet Incorporated

* **Promisify gm**. Unclear how to do this, https://github.com/aheckmann/gm/issues/320 leads to the current implementation.
* data/object naming: data is the standard function signature in the Function Framework, object reassignment is to align with the idea of a Cloud Storage object. I defer to Ace on this.

### Other feedback for discussion

* **BLURRED_BUCKET_NAME is a global variable, why pass it in as a function parameter?** Yes it is, do we prefer to access the global variable generally or try to limit functions to their parameter signature where possible?

### DIFF
--- a/functions/imagemagick/index.js
+++ b/functions/imagemagick/index.js
@@ -21,9 +21,9 @@ const fs = require('fs');
 const {promisify} = require('util');
 const path = require('path');
 const {Storage} = require('@google-cloud/storage');
-const storage = new Storage();
-const vision = require('@google-cloud/vision').v1p1beta1;
+const vision = require('@google-cloud/vision');
 
+const storage = new Storage();
 const client = new vision.ImageAnnotatorClient();
 
 const {BLURRED_BUCKET_NAME} = process.env;
@@ -45,6 +45,7 @@ exports.blurOffensiveImages = async event => {
     const detections = result.safeSearchAnnotation || {};
 
     if (
+      // Levels are defined in https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.Likelihood
       detections.adult === 'VERY_LIKELY' ||
       detections.violence === 'VERY_LIKELY'
     ) {
@@ -55,7 +56,7 @@ exports.blurOffensiveImages = async event => {
     }
   } catch (err) {
     console.error(`Failed to analyze ${file.name}.`, err);
-    return Promise.reject(err);
+    throw err;
   }
 };
 // [END functions_imagemagick_analyze]
@@ -71,8 +72,7 @@ const blurImage = async (file, blurredBucketName) => {
 
     console.log(`Downloaded ${file.name} to ${tempLocalPath}.`);
   } catch (err) {
-    console.error('File download failed.', err);
-    return Promise.reject(err);
+    throw new Error(`File download failed: ${err}`);
   }
 
   await new Promise((resolve, reject) => {
@@ -98,8 +98,7 @@ const blurImage = async (file, blurredBucketName) => {
     await blurredBucket.upload(tempLocalPath, {destination: file.name});
     console.log(`Uploaded blurred image to: ${gcsPath}`);
   } catch (err) {
-    console.error(`Unable to upload blurred image to ${gcsPath}:`, err);
-    return Promise.reject(err);
+    throw new Error(`Unable to upload blurred image to ${gcsPath}: ${err}`);
   }
 
   // Delete the temporary file.
--- a/functions/imagemagick/test/index.test.js
+++ b/functions/imagemagick/test/index.test.js
@@ -31,40 +31,64 @@ requestRetry = requestRetry.defaults({
   retryDelay: 1000,
 });
 
-const BUCKET_NAME = process.env.FUNCTIONS_BUCKET;
-const {BLURRED_BUCKET_NAME} = process.env;
 
 const safeFileName = 'bicycle.jpg';
 const offensiveFileName = 'zombie.jpg';
-
+const BUCKET_NAME = process.env.FUNCTIONS_BUCKET;
+const {BLURRED_BUCKET_NAME} = process.env;
+const blurredBucket = storage.bucket(BLURRED_BUCKET_NAME);
 const cwd = path.join(__dirname, '..');
 
-const blurredBucket = storage.bucket(BLURRED_BUCKET_NAME);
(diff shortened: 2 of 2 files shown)


functions/imagemagick/index.js
```diff
@@ -21,9 +21,9 @@ const fs = require('fs');
 const {promisify} = require('util');
 const path = require('path');
 const {Storage} = require('@google-cloud/storage');
-const storage = new Storage();
-const vision = require('@google-cloud/vision').v1p1beta1;
+const vision = require('@google-cloud/vision');
 
+const storage = new Storage();
```
grant: Not sure why separated defining and instantiating `Storage`. I'd undo change L24 to L26.

functions/imagemagick/index.js
```diff
@@ -45,6 +45,7 @@ exports.blurOffensiveImages = async event => {
     const detections = result.safeSearchAnnotation || {};
 
     if (
+      // Levels are defined in https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.Likelihood
```
grant: Please link REST, not RPC reference, as we are using a REST client.

functions/imagemagick/index.js
```diff
@@ -71,8 +72,7 @@ const blurImage = async (file, blurredBucketName) => {
 
     console.log(`Downloaded ${file.name} to ${tempLocalPath}.`);
   } catch (err) {
-    console.error('File download failed.', err);
-    return Promise.reject(err);
+    throw new Error(`File download failed: ${err}`);
```
grant: Are we sure this will stringify correctly? Do we not mean to provide the `${err.message}`?

functions/imagemagick/test/index.test.js
```diff
@@ -31,40 +31,64 @@ requestRetry = requestRetry.defaults({
...
+    if (!exists[0]) {
+      throw Error(`Missing required file: gs://${BUCKET_NAME}/${safeFileName}`);
     }
-  };
+  });
+
 
-  beforeEach(tools.stubConsole);
-  afterEach(tools.restoreConsole);
+//  beforeEach(tools.stubConsole);
```
grant: This block needs a comment as to why these two lines are disabled rather than deleted.

functions/imagemagick/test/index.test.js
```diff
@@ -31,40 +31,64 @@ requestRetry = requestRetry.defaults({
...
       }
+    };
+  });
+
+  before(async () => {
+    let exists = await storage
+      .bucket(BUCKET_NAME)
+      .file(offensiveFileName)
+      .exists();
+    if (!exists[0]) {
```
grant: I'd rather see `if (!exists) {`
by setting
`let exists = ... .exists()[0];`

See example where we declare `exists` to be the first element of the data array:
https://cloud.google.com/nodejs/docs/reference/storage/1.6.x/Bucket#exists

functions/imagemagick/test/index.test.js
```diff
@@ -31,40 +31,64 @@ requestRetry = requestRetry.defaults({
...
 const offensiveFileName = 'zombie.jpg';
-
+const BUCKET_NAME = process.env.FUNCTIONS_BUCKET;
+const {BLURRED_BUCKET_NAME} = process.env;
+const blurredBucket = storage.bucket(BLURRED_BUCKET_NAME);
 const cwd = path.join(__dirname, '..');
 
-const blurredBucket = storage.bucket(BLURRED_BUCKET_NAME);
+// Successfully generated images require cleanup.
+let cleanupRequired = false;
```
grant: The identifier `cleanupRequired` should be scoped within the `describe` closure rather than file-scoped.
Could we put the boolean next to `let startFF, stopFF;`?

functions/imagemagick/test/index.test.js
```diff
@@ -31,40 +31,64 @@ requestRetry = requestRetry.defaults({
   retryDelay: 1000,
 });
 
-const BUCKET_NAME = process.env.FUNCTIONS_BUCKET;
-const {BLURRED_BUCKET_NAME} = process.env;
 
 const safeFileName = 'bicycle.jpg';
 const offensiveFileName = 'zombie.jpg';
-
+const BUCKET_NAME = process.env.FUNCTIONS_BUCKET;
```
grant: Nit: It would be nice if we just used `FUNCTIONS_BUCKET` so that we could destruct `process.env` in the same way:

- `FUNCTIONS_BUCKET`
- `BLURRED_FUNCTIONS_BUCKET`

functions/imagemagick/index.js
```diff
@@ -98,8 +98,7 @@ const blurImage = async (file, blurredBucketName) => {
     await blurredBucket.upload(tempLocalPath, {destination: file.name});
     console.log(`Uploaded blurred image to: ${gcsPath}`);
   } catch (err) {
-    console.error(`Unable to upload blurred image to ${gcsPath}:`, err);
-    return Promise.reject(err);
+    throw new Error(`Unable to upload blurred image to ${gcsPath}: ${err}`);
```
grant: Same comment about `err.message`.

functions/imagemagick/test/index.test.js
grayside: Agreed. From what I can see, BUCKET_NAME is used in the same because it makes sense as simply the bucket causing the function trigger. FUNCTIONS_BUCKET is defined in `.kokoro/build.sh` and has that name so the product testing config is disambiguated.

functions/imagemagick/test/index.test.js
grayside: Sure, I ended up on the current implementation as a result of my path to realizing exists() returns an array.

Should I use `let [exists] = ... .exists()` instead?

grant: Sounds fine.

PR #1480, opened 2019-09-13
run: add hello-broken sample

grayside:
This adds a "broken" / difficult to troubleshoot sample for use in a troubleshooting tutorial.

This sample app is unusual in that it's is deliberately trying to show something **not good** as part of an exercise leading the developer through the full process of root cause analysis, fixing production, then improving the code.

Go variant: https://github.com/GoogleCloudPlatform/golang-samples/pull/981


run/hello-broken/test/runner.sh
```diff
@@ -0,0 +1,54 @@
...
+
+echo
+echo '---'
+echo
+
+# Register post-test cleanup.
+# Only needed if deploy completed.
+function cleanup {
+  set -x
+  gcloud --quiet beta run services delete ${SERVICE_NAME} \
```
grant: Ditto about putting --quiet between subcommand words.

run/hello-broken/test/system.test.js
```diff
@@ -0,0 +1,69 @@
+// Copyright 2019 Google LLC. All rights reserved.
```
grant: Incomplete Apache license header.

run/hello-broken/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Copying this first prevents re-running npm install on every code change.
+COPY package*.json ./
+
+# Install production dependencies.
+# If you add a package-lock.json, speed your build by switching to 'npm ci'.
+# RUN npm ci --only=production
+RUN npm install --only=production
+
+# Copy local code to the container image.
+COPY . ./
```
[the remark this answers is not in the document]
grayside: While the previous practice was to use `COPY . .`, this is a newer practice I think we should use in COPY statements. The issue is COPY supports having multiple source arguments and one destination argument, but if the source arguments are:

* A directory
* More than one file

Then the destination needs to have the trailing slash otherwise it will error.

grayside:
Note the README here has a good amount of wording copied from other Run samples, I will revise all of them in follow-up to keep it in sync.

run/hello-broken/README.md
```diff
@@ -0,0 +1,69 @@
...
+export CONTAINER_IMAGE=gcr.io/${GOOGLE_CLOUD_PROJECT}/broken
+export REGION=us-central1
+test/deploy.sh
+```
+
+### runner.sh
+
+The `runner.sh` script is a helper that facilitates:
+
+* Deploy the service to Cloud Run based on the `deploy.sh` script.
```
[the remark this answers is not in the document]
grayside: Thanks for calling this out, I avoid using the word "helper" at all times, not sure how this got in.

run/hello-broken/test/system.test.js
grant: No offense, but the current license header or previous ones do not matter.
OSPO requires an unmodified version of the header; there is no shortened version:
go/releasing/preparing#license-headers

Many PRs and projects don't fully follow the rules, but if we are trying to add a header here, we should use the full form as described in the doc above.

grayside: Fair, I am trying to adhere to whatever standard this repo prefers. Will revise files changed in my currently open PRs.

run/hello-broken/index.js
```diff
@@ -0,0 +1,58 @@
...
+  console.log('hello: received request.');
+
+  // [START run_broken_service_problem]
+  const {NAME} = process.env;
+  if (!NAME) {
+    // Plain error logs do not appear in Stackdriver Error Reporting.
+    console.error('Environment validation failed.');
+    console.error(new Error('Missing required server parameter'));
+    res.status(500).send('Internal Server Error');
+    return;
```
grant: Nit: I find it common to add `return` to the previous line like `return res.send...`.

run/hello-broken/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "hello-broken",
+  "private": true,
+  "description": "Broken Cloud Run service for troubleshooting practice",
+  "main": "index.js",
+  "scripts": {
+    "start": "node index.js",
+    "test": "echo \"Error: no test specified\" && exit 0",
+    "e2e-test": "TARGET=Cloud test/runner.sh mocha test/system.test.js --timeout=20000"
```
grant: It might be worth it to just put this script as `test` so you may [`npm test`](https://docs.npmjs.com/cli/test.html) instead of `npm run e2e-test`.

run/logging-manual/README.md
```diff
@@ -42,7 +42,7 @@ test/deploy.sh
 
 ### runner.sh
 
-The `runner.sh` script is a helper that facilitates:
+The `runner.sh` script will:
```
grant: Saying bash files _will_ do something sounds a bit weird to me.

Maybe state what the program does rather than what it will do.

---

The `runner.sh` script:
- Deploys the service...
- Sets the ... variables...

run/logging-manual/README.md
```diff
@@ -55,11 +55,13 @@ test/runner.sh sleep 20
 
 ## Environment Variables (Testing)
 
-* `BASE_URL`: Used to designate the URL of the Cloud Run service under system test. The URL is not deterministic, so the options are to record it at deploy time or use the Cloud Run API to retrieve the value.
-* `ID_TOKEN`: Used for authenticated HTTP requests to the Cloud Run service.
+* `BASE_URL`: Specifies the Cloud Run service URL for end-to-end tests.
```
grant: The wording for this section seems inconsistent (`BASE_URL` vs `REGION` descriptions for example).
All env vars specify or configure something. 

Can we just say:
- `BASE_URL`: The Cloud Run service URL...
...
- `SERVICE_NAME`: The name of the deployed service. Used in some API calls...

run/logging-manual/README.md
```diff
@@ -55,11 +55,13 @@ test/runner.sh sleep 20
...
+  The default assigned URL of a Cloud Run service must be derived after deployment.
+* `ID_TOKEN`: JWT token used to authenticate with Cloud Run's IAM-based authentication.
 * `REGION`: [`us-central1`] Optional override region for the location of the Cloud Run service.
-* `SERVICE_NAME`: Used in testing to specify the name of the service. The name may be included in API calls and test conditions.
-* `GOOGLE_CLOUD_PROJECT`: Used in production as an override to determination of the GCP Project from the GCP metadata server. Used in testing to configure the Logging client library.
+* `SERVICE_NAME`: Specify the name of the deployed service, used in some API calls and test assertions.
+* `GOOGLE_CLOUD_PROJECT`: Required for local testing, overrides use of the GCP metadata server to determine the current GCP project.
+  Required by the logging client library.
```
grant: We say `Required` twice.
Can we combine the sentences and say something like:

> Required for the logging client library for local testing. Overrides GCP metadata server config to determine the current GCP project.

run/logging-manual/README.md
grayside: Ack

run/logging-manual/README.md
grayside: Made some clarifications, and identified that the GOOGLE_CLOUD_PROJECT variable is not needed in the hello-broken sample.

run/hello-broken/package.json
grayside: Out of scope of this PR.

grayside: Actually, now that I see your point about `npm test` being a no-op, this bears some follow-up. I will file an issue.

grant: I've looked at the other samples that include script `e2e-test` in this repo as well as the cDPE test guide and squad notes.

If we consider this change out of scope, then I oppose adding this line then:

```js
"test": "echo \"Error: no test specified\" && exit 0",
```

We do specify a test in this repo. We don't specify a config for specifying a test as the error suggests.

If we're going to add this script:

```js
"e2e-test": ...
```

We should either consider using the Node way of `npm test` or fix the error message for `npm test` to be more accurate.

grayside: #1493

PR #1487, opened 2019-09-18
run: add System Package sample and align ignore files

grayside:
This sample previously underwent review by @fhinkel and @ace-n.

Moving it here as a permanent home with the other Cloud Run Node.js samples.

Changes since last review include:

* refactoring the tests use use mocha & supertest instead of repo tools & ava
* running tests outside the container


run/logging-manual/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "logging-manual",
-  "version": "1.0.0",
   "description": "Demonstrates logging on Cloud Run",
   "main": "index.js",
+  "private": true,
```
fhinkel: 👍

run/pubsub/README.md
```diff
@@ -9,4 +9,6 @@ For more details on how to work with this sample read the [Google Cloud Run Node
 ## Dependencies
 
 * **express**: Web server framework.
```
fhinkel: I'm wondering if this list is useful or just a hassle to keep up to date. We might consider deleting it in the future.

run/system-package/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
grant: This is redundant as `node_modules` is already ignored in a higher-level `.gitignore` file.
Can we remove this?

run/pubsub/README.md
grant: I would remove this section as we aren't consistently using it throughout this repo. Instead, we can comment in the files where they are used.

run/system-package/README.md
```diff
@@ -0,0 +1,13 @@
+# Cloud Run System Package sample
```
grant: Title Case the H1. In this case, `Sample` should be uppercase.

See this `README.md` for an example:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/run/README.md#google-cloud-run-nodejs-samples

run/system-package/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This sample shows how to use a CLI tool installed as a system package as part of a web service.
+
+Use it with the [Using system packages tutorial](https://cloud.google.com/run/docs/tutorials/system-packages).
+
+For more details on how to work with this sample read the [Google Cloud Run Node.js Samples README](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/run).
```
grant: > Google Cloud Run Node.js Samples README

This phrasing is awkward. Since we're already in the Node Sample repo which is under the GCP org, can we just say something simpler like "Cloud Run Samples README"?

run/system-package/app.js
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      res.status(500).send('Internal Server Error');
+    }
+  }
+});
+
+// [END run_system_package_handler]
+
+// [START run_system_package_exec]
+
```
grant: Extra \n here and above?

run/system-package/app.js
```diff
@@ -0,0 +1,70 @@
...
+// Generate a diagram based on a graphviz DOT diagram description.
+const createDiagram = dot => {
+  if (!dot || (dot.constructor === Object && Object.keys(dot).length === 0)) {
```
grant: I know this is copied from
https://github.com/GoogleCloudPlatform/cloud-run-samples/blob/master/nodejs/system_package/app.js#L41
but I think it could be more readable.

Checking if `dot` has a `constructor` property that is the `Object` type is awkward.

```js
const dotIsEmpty = !dot || (typeof dot === 'object' && !Object.keys(dot).length);
if (dotIsEmpty) {
  throw error..
}
```

run/system-package/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "graphviz-web",
+  "version": "1.0.0",
+  "description": "Demonstrates a Cloud Run service which provides a CLI tool over HTTP.",
+  "main": "index.js",
+  "private": true,
+  "repository": {
```
grant: I don't this this property is needed in this `package.json`.

run/system-package/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "type": "git",
+    "url": "https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git"
+  },
+  "scripts": {
+    "start": "node index.js",
+    "test": "mocha test/app.test.js"
+  },
+  "author": "",
```
grant: I don't this this empty property is useful. Could be deleted.

Same with `license`.
We already have this property here:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/package.json#L5

run/system-package/app.js
```diff
@@ -0,0 +1,70 @@
...
+    throw new Error('syntax: no graphviz definition provided');
+  }
+
+  const watermark = ` \
```
grant: We might want to make this an array and just get rid of the `\` like this:

```js
// Adds a watermark to the dot graphic.
const dotFlags = [
  '-Glabel="Made on Cloud Run"',
  '-Gfontsize=10',
  '-Glabeljust=right',
  '-Glabelloc=bottom',
  '-Gfontcolor=gray',
].join(' ');
```

Also, the identifier `watermark` could be clearer. We're really talking about cmd flags that create the watermark.

run/pubsub/README.md
grayside: It's consistent in the Cloud Run samples. While the really common dependencies like express may be fairly obvious, I'd argue that it is valuable to facilitate maintenance by explaining why dependencies are present, and it may be worth the hassle to expand the practice. Let's make Cloud Run a test?

run/system-package/package.json
grayside: I got an error for not having it, so added it. Maybe it's linked to having private: true?

run/system-package/package.json
grayside: If this is standard, sure. Seems like these would be explicit in every package.json, along the same reasoning that every file has a license header. Removing for now.

run/system-package/app.js
grayside: Trying to reproduce why this might be there, but I can't recall the edge case. This is probably a holdover from when the sample was POST based and pulled the dot description from the request body.

run/system-package/README.md
grayside: There is going to be a second link added soon to the main Cloud Run Samples README in another repository.

run/logging-manual/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "logging-manual",
-  "version": "1.0.0",
```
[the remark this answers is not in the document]
grayside: Done

run/system-package/.gitignore
grayside: Done

run/system-package/README.md
grayside: Done

run/system-package/app.js
grayside: Done

run/system-package/app.js
grayside: Dropped the second condition.

run/system-package/README.md
grant: OK.

grayside:
run/logging-manual failed due to outage.
functions/helloworld failed due to system test timeout exceeded.

Re-running tests.

fhinkel:
FYI, you can restart individual tests instead of running all tests again.

PR #1491, opened 2019-09-20
refactor: Remove functions_concepts_after_response

grant:
Removes `functions_concepts_after_response`.

This is not a useful GCF function, does not have a test, and does not have a strong reason to keep for N languages. Only implemented in Node right now.

[samples/596](https://devrel.corp.google.com/samples/596)
Doc: https://cloud.google.com/functions/docs/concepts/exec#function_execution_timeline

### DIFF
--- a/functions/concepts/index.js
+++ b/functions/concepts/index.js
@@ -58,22 +58,6 @@ exports.executionCount = (req, res) => {
 };
 // [END functions_concepts_stateless]
 
-// [START functions_concepts_after_response]
-/**
- * HTTP Cloud Function that may not completely
- * execute due to early HTTP response
- *
- * @param {Object} req Cloud Function request context.
- * @param {Object} res Cloud Function response context.
- */
-exports.afterResponse = (req, res) => {
-  res.end();
-
-  // This statement may not execute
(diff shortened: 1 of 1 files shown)


ace-n:
Personally I think that we should keep this (since it's useful for explaining the related concept), but I'll defer to @labtopia

labtopia:
This may not be the very best sample we could have had here, but I think the underlying concept is important, and the sample does illustrate it. Further, this doesn't seem like a case where we need a sample for every runtime. A single sample illustrating the concept is sufficient. I think the section would be a lot weaker without the sample. Therefore, I am not in favor of removing it.

grant:
I'd rather have the one place in our docs have something shorter so the reader doesn't need to read the whole sample:

```js
res.end();

// This statement may not execute
console.log('Function complete!');
```

What about moving the region tags to just show this (above)?

ace-n:
Middle ground - how about we exclude the JSDoc comment with the region tags, to clearly delineate that this code is (still) within a function?

grant:
Closing due to lack of interest in changing this sample.

PR #1495, opened 2019-09-25
Use await instead of promises

fhinkel:
Drive-by fix: Run linter for template strings.

### DIFF
--- a/functions/slack/index.js
+++ b/functions/slack/index.js
@@ -106,26 +106,23 @@ const verifyWebhook = body => {
  *
  * @param {string} query The user's search query.
  */
-const makeSearchRequest = query => {
-  return new Promise((resolve, reject) => {
-    kgsearch.entities.search(
-      {
-        auth: config.KG_API_KEY,
-        query: query,
-        limit: 1,
-      },
-      (err, response) => {
-        console.log(err);
-        if (err) {
-          reject(err);
-          return;
-        }
-
-        // Return a formatted message
-        resolve(formatSlackMessage(query, response));
+const makeSearchRequest = async query => {
+  await kgsearch.entities.search(
+    {
+      auth: config.KG_API_KEY,
+      query: query,
+      limit: 1,
+    },
+    (err, response) => {
+      console.log(err);
+      if (err) {
+        throw err;
       }
-    );
-  });
+
+      // Return a formatted message
+      return formatSlackMessage(query, response);
+    }
+  );
 };
 // [END functions_slack_request]
 
@@ -161,12 +158,10 @@ exports.kgSearch = async (req, res) => {
 
     // Send the formatted message back to Slack
     res.json(response);
-
-    return Promise.resolve();
   } catch (err) {
     console.error(err);
     res.status(err.code || 500).send(err);
-    return Promise.reject(err);
+    throw err;
   }
 };
 // [END functions_slack_search]
--- a/functions/spanner/index.js
+++ b/functions/spanner/index.js
@@ -50,9 +50,7 @@ exports.get = async (req, res) => {
     const rows = results[0].map(row => row.toJSON());
     rows.forEach(row => {
       res.write(
-        `SingerId: ${row.SingerId}, ` +
-          `AlbumId: ${row.AlbumId}, ` +
-          `AlbumTitle: ${row.AlbumTitle}\n`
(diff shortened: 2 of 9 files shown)


ace-n:
This _may_ have been obviated by #1489 🙂

fhinkel:
Haha, yes, most is obviated. But a few last changes are left.

functions/slack/index.js
```diff
@@ -106,26 +106,23 @@ const verifyWebhook = body => {
...
-        console.log(err);
-        if (err) {
-          reject(err);
-          return;
-        }
-
-        // Return a formatted message
-        resolve(formatSlackMessage(query, response));
+const makeSearchRequest = async query => {
+  await kgsearch.entities.search(
```
ace-n: 1) Do we know if this _actually_ returns a promise? (either via running it ourselves, or reviewing Slack's docs)
2) If so, we should get rid of the callback below. 🙂

functions/slack/index.js
```diff
@@ -161,12 +158,10 @@ exports.kgSearch = async (req, res) => {
 
     // Send the formatted message back to Slack
     res.json(response);
-
-    return Promise.resolve();
   } catch (err) {
     console.error(err);
     res.status(err.code || 500).send(err);
-    return Promise.reject(err);
+    throw err;
```
ace-n: Nit (perhaps better addressed in a different PR): redundant?

functions/spanner/index.js
```diff
@@ -50,9 +50,7 @@ exports.get = async (req, res) => {
     const rows = results[0].map(row => row.toJSON());
     rows.forEach(row => {
       res.write(
-        `SingerId: ${row.SingerId}, ` +
-          `AlbumId: ${row.AlbumId}, ` +
-          `AlbumTitle: ${row.AlbumTitle}\n`
+        `SingerId: ${row.SingerId}, AlbumId: ${row.AlbumId}, AlbumTitle: ${row.AlbumTitle}\n`
```
ace-n: I believe @grant [was 👎on this](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1489#discussion_r326799551).

functions/tokenservice/functions/index.js
```diff
@@ -149,21 +149,17 @@ exports.getOAuthToken = functions.https.onCall(async (data, context) => {
...
+    const doc = await docRef.get();
     if (doc.exists && isValid(doc.data().expireTime)) {
-      //push notification
       pushNotification(
         data['deviceID'],
         doc.data().accessToken,
         doc.data().expireTime
       );
-      return doc.data();
+      await doc.data();
```
ace-n: This seems confusing - why are we using `doc.data()` presumably _before_ we've `await`ed it? (If we know this case _definitely won't happen_, we should at least leave a comment explaining that.)

cc @santhoshvaddi

functions/tokenservice/functions/index.js
```diff
@@ -149,21 +149,17 @@ exports.getOAuthToken = functions.https.onCall(async (data, context) => {
...
     } else {
       const result = await retrieveCredentials(context);
-      console.log('Print result from retrieveCredentials functions');
```
ace-n: @santhoshvaddi I assume this deletion is fine with you?

functions/tokenservice/functions/index.js
```diff
@@ -175,7 +171,7 @@ exports.getOAuthToken = functions.https.onCall(async (data, context) => {
     console.log('Error retrieving token', err);
     pushNotification(data['deviceID'], 'Error retrieving token', 'Error');
     // return 'Error retrieving token';
```
ace-n: Nit: redundant comment, remove?

functions/tokenservice/functions/index.js
santhoshvaddi: After push notification we just want to return from the function. The client will receive the token in the form of push notification. I don't think `await` is required here.

functions/tokenservice/functions/index.js
santhoshvaddi: Yes, this is fine with me.

functions/tokenservice/functions/index.js
santhoshvaddi: Yes, we can remove this.

PR #1508, opened 2019-10-05
Cloud SQL (MySQL): Add Cloud Run support and fix broken implementation

grayside:
This PR originally intended just the following:

* Adds a Dockerfile to the MySQL sample for use with Cloud Run, based on the Dockerfiles used in the Cloud Run samples.
* Adds README instructions on deployment to Cloud Run.
* Adds the sample to the list of [Node.js Cloud Run samples](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/run/README.md).

In the course of the Cloud Run work, I found the current state of the sample is broken. This PR includes fixes for those problems:

1. promise-mysql in v4 returns a promise from `mysql.createPool()`. pool instantiation in the master implementation cannot have worked since the library update. Fix: Hacky top-level async.
2. The code executed in `app.on('listening', async () => {`... is not being fired (express 4, node 10), meaning the votes table is never created and the client is left hanging while the Node process crashes. Fix: Move the table creation directly after pool creation. Note also the "votes tables" query is an important step in validating the pool configuration as part of service startup. Cloud Run will flag a revision as unhealthy if it fails before listening.

The top-level async/await is a little kludgy, happy to clean it up with additional guidance. One consideration: the potential errors from the mysql library do not seem to benefit from a stack trace, the error message itself is sufficient for troubleshooting purposes (though maybe the goal is to have automatic Stackdriver Error Reporting support?)

### DIFF
--- a/cloud-sql/mysql/mysql/server.js
+++ b/cloud-sql/mysql/mysql/server.js
@@ -43,7 +43,7 @@ const logger = winston.createLogger({
 });
 
 // [START cloud_sql_mysql_mysql_create]
-const pool = mysql.createPool({
+const connectionOptions = {
   user: process.env.DB_USER, // e.g. 'my-db-user'
   password: process.env.DB_PASS, // e.g. 'my-db-password'
   database: process.env.DB_NAME, // e.g. 'my-database'
@@ -83,18 +83,23 @@ const pool = mysql.createPool({
   // [END cloud_sql_mysql_mysql_backoff]
 
   //[END_EXCLUDE]
-});
+};
+let pool;
+(async () => {
+  try {
+    pool = await mysql.createPool(connectionOptions);
+    // Wait for tables to be created (if they don't already exist).
+    await pool.query(
+      `CREATE TABLE IF NOT EXISTS votes
+        ( vote_id SERIAL NOT NULL, time_cast timestamp NOT NULL,
+        candidate CHAR(6) NOT NULL, PRIMARY KEY (vote_id) );`
+    );
+  } catch (err) {
+    logger.err(err);
+    process.exit(1);
+  }
+})();
 // [END cloud_sql_mysql_mysql_create]
-
-// When the server starts, check for tables in the database.
-app.on('listening', async () => {
-  // Wait for tables to be created (if they don't already exist).
-  await pool.query(
-    `CREATE TABLE IF NOT EXISTS votes
-      ( vote_id SERIAL NOT NULL, time_cast timestamp NOT NULL,
-      candidate CHAR(6) NOT NULL, PRIMARY KEY (vote_id) );`
-  );
-});
 
 // Serve the index page, showing vote tallies.
 app.get('/', async (req, res) => {
--- a/cloud-sql/mysql/mysql/server.js
+++ b/cloud-sql/mysql/mysql/server.js
@@ -43,63 +43,62 @@ const logger = winston.createLogger({
 });
 
 // [START cloud_sql_mysql_mysql_create]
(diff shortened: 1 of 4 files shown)


cloud-sql/mysql/mysql/server.js
```diff
@@ -43,7 +43,7 @@ const logger = winston.createLogger({
 });
 
 // [START cloud_sql_mysql_mysql_create]
-const pool = mysql.createPool({
+const connectionOptions = {
```
kurtisvg: This changes moves the pool creation outside of the snippet, which we probably don't want to do since that's what this snippet is for.

grayside: Could you double-check the diff? I deliberately tried to keep things inside the correct region tags, and it still reads that way to me.

grayside:
> These samples have tests - is there a reason those tests didn't detect the break when mysql-promise was bumped to 4?

I took a quick look at the test for this sample, and it's a fairly shallow test with high reliance on stubbing. I expect the stubbing mechanism prevented the test from making the distinction between returning the pool and returning a promise.

I'm happy to fix things up a bit, but I don't have bandwidth to deepen the tests.

fhinkel:
lgtm. Using less stubbing in the tests is out of scope. Can you make the linter happy?

cloud-sql/mysql/mysql/server.js
kurtisvg: My mistake, I mismatched the tags. 

Downgrading this to a small nit: Can we move connectionOptions inline? It's not being used anywhere else, so I think it would be more clear inside of `mysql.createPool`.

cloud-sql/mysql/mysql/server.js
```diff
@@ -83,19 +83,24 @@ const pool = mysql.createPool({
   // [END cloud_sql_mysql_mysql_backoff]
 
   //[END_EXCLUDE]
-});
+};
+let pool;
+(async () => {
```
kurtisvg: nit: I spoke with Ben about the best way to handle top level awaits, he mentioned this might be preferable as a named function:
```
(async function startPool () {
   ...
})()
```

cloud-sql/mysql/mysql/server.js
```diff
@@ -83,19 +83,24 @@ const pool = mysql.createPool({
...
+  try {
+    pool = await mysql.createPool(connectionOptions);
+    // Wait for tables to be created (if they don't already exist).
+    await pool.query(
```
kurtisvg: Can we move this (database init) into it's own function? I don't make sense in the context of the snippet to include a query.

cloud-sql/mysql/mysql/server.js
fhinkel: Yes, please name the function. In the sample template we don't immediately invoke the function but use the function name instead 

```
async function startPool() {
   // ...
};
startPool();
```

cloud-sql/mysql/mysql/server.js
grayside: Can do. I broke it out because I like to be able to read the reasoning around a try/await/catch without scrolling.

cloud-sql/mysql/mysql/server.js
grayside: `async function startPool() {}` triggered a linting error (`Expected a function expression  func-style`).

Proceeding with the same guidance using `const createPool = async () => {}`.

grayside:
Feedback incorporated, linting fixed. Waiting for tests.

PR #1528, opened 2019-10-25
Fix iot test by using new method name

mesmacosta:
I noticed that this test started breaking in another PR that I opened.

Doing a quick analysis:

The iot/manager test failing seem to be related to this change:
googleapis/nodejs-iot@68a1d6e

Method name changed, but test from sample code was not updated.
client.registryPath -> client.deviceRegistryPath

So I can't say for sure, because I don't know the full extend of that commit, but this small change might fix it, I'd like to run it in the integrated environment to see if it will solve.
@fhinkel


gguuss:
Tests are passing for me locally, merging.

gguuss:
I will see about getting the breaking change rolled back.

gguuss:
I'll see if we can revert the breaking change

mesmacosta:
does it make sense to resolve the conflicting file? @gguuss

gguuss:
Yes, let's just go ahead and allow the breaking change in this case, I'll try and track down the other code samples that are out there requiring changes.

gguuss:
@mesmacosta Thank you so much for making the PR; we were able to get this resolved in the update to the IoT Client library version 1.3.1 so I'm going to go ahead and close this, favoring skipping the additional record of the breaking change as done in [this PR](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1539).

PR #1532, opened 2019-10-30
Updates speech-to-speech and ocr-app samples.

ricalo:
* Updates [google-cloud/translate][1] dependency to version `5.0.0`.
* Updates declaration of clients to specify the version 2 of the API instead of
the version 3, which is the default in the Translate library.

Note: The google-cloud/translate library made v3 the default in https://github.com/googleapis/nodejs-translate/pull/355

[1]: https://github.com/googleapis/nodejs-translate


ricalo:
This replaces #1531.

Issue #1550, opened 2019-11-25
Clean up IoT beta-features samples


gguuss:
These are GA now and should already be in the manager / MQTT examples.

gguuss:
Fixed in #1551

PR #1551, opened 2019-11-26
Gateway features now GA and in manager

gguuss:
These samples are no longer being used and are no longer referenced in documentation.


fhinkel:
fixes https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/1550?

PR #1555, opened 2019-11-27
Remove tools.checkCredentials()

fhinkel:
We want to remove the nodejs-repo-tools dependency. Doing so step by step.

Drive-by fix: Rename `utils` to `tools` for consistency.

### DIFF
--- a/appengine/storage/flexible/system-test/app.test.js
+++ b/appengine/storage/flexible/system-test/app.test.js
@@ -17,16 +17,23 @@ const path = require('path');
 const {Storage} = require('@google-cloud/storage');
 const storage = new Storage();
 const assert = require('assert');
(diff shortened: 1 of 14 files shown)


grant:
Isn't this introducing a lot of duplicated code?
Why can't we simply extract the method into a helper function that doesn't duplicate code or add SLOCs?

fhinkel:
The point is to get rid of helper methods. Samples need to be self contained. Black box helper methods like `checkCredentials` lead to unnecessary code execution. As a next step, we need to check for every sample if these checks are necessary.

grant:
From the PR sent, it looked like the motivation was to remove the dependency with `nodejs-repo-tools`.
I think we could consider simply moving the `checkCredentials` method to this repo. Or at least state why we aren't doing that in the PR description.

PR #1568, opened 2019-12-17
Data Catalog: add create fileset Entry sample

ricardolsmendes:
Fileset operations were added to Data Catalog's API and the client library now supports them. Samples are required to update the official docs.


ricardolsmendes:
This change depends on https://github.com/googleapis/nodejs-datacatalog/issues/74 to pass the system tests.

ricardolsmendes:
Waiting for https://github.com/googleapis/nodejs-datacatalog/issues/79. The system tests will fail without the fix.

PR #1569, opened 2019-12-17
Data Catalog: add delete fileset Entry sample

ricardolsmendes:
Fileset operations were added to Data Catalog's API and the client library now supports them. Samples are required to update the official docs.


ricardolsmendes:
This change depends on https://github.com/googleapis/nodejs-datacatalog/issues/74 to pass the system tests.

ricardolsmendes:
> @fhinkel is there a reason we'd put this here and not googleapis/nodejs-datacatalog?

Hi, @JustinBeckwith and @fhinkel,
I discussed with @bcoe a while ago and this is what we agreed so far:
1. the present change should be part of this repository because its main code includes more than one API call and is intended to be referred by Quickstart guides in Data Catalog docs;
1. the same applies to https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1568;
1. once the quickstart-related changes have been merged, we should move the samples residing at https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/datacatalog/cloud-client to **googleapis/nodejs-datacatalog**, since each of them represents a single API call and would work as smoke tests to the client library as well.

Please let me know your thoughts.

Kind regards,
Ricardo

fhinkel:
Sounds good to me!

PR #1570, opened 2019-12-19
Replace app engine js compilation with ts-node

sashamor:
Replace app engine js compilation step with running ts-node on the app engine server.

As ts-node gains popularity, the benefits of debugging and instrumenting typescript code in real-time justify skipping compilation to JS altogether.

Not sure if this has already been considered. Also note that `repo-tools test` may not support testing typescript code directly, although many popular typescript testing libraries (e.g. jest) do support this.

### DIFF
--- a/appengine/typescript/.gcloudignore
+++ b/appengine/typescript/.gcloudignore
@@ -1,5 +1,2 @@
-# Exclude compiled .js files
-*.js
-
 # Exclude dependencies
 node_modules/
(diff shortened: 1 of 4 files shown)


sashamor:
@googlebot I signed it!

fhinkel:
@sashamor Thanks for the pull request. In our docs, we want to explicitely show how to compile TypeScript sources instead of using ts-node.  Thank for understanding.

PR #1577, opened 2019-12-26
Example showing bug with gcloud and sibling directories

janakdr:
With this patch, running from `appengine/hello-world/sibling/subdir/index.js`, both `npm start` and `yarn start` succeed. However, running `gcloud app deploy` fails. If using `npm`, it seems to not find the `package.json` file in the subdirectory:

```
Step #1 - "builder": npm ERR! Could not install from "../sibling" as it does not contain a package.json file.
```

If `yarn.lock` is present, it fails with a stranger error:

```
Step #1 - "builder": error Package "sibling-dep" refers to a non-existing file '"/sibling"'.
```

### DIFF
--- a/appengine/hello-world/sibling/index.js
+++ b/appengine/hello-world/sibling/index.js
@@ -1,0 +1,2 @@
+const name = 'sibling';
+module.exports = {name};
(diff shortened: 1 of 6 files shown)


janakdr:
Sorry, I didn't actually want to add this as a PR to the original repository. Just wanted to create a patch showing the issue in order to file an internal bug. Please ignore!

PR #1588, opened 2020-01-11
Add .mocharc to create test XML results.

ace-n:
Part 2 of #1587 

~~**DO NOT MERGE** until _all_ test results are (somehow) visible in the Kokoro UI.~~


.kokoro/.mocharc.yml
```diff
@@ -0,0 +1,3 @@
+reporter: xunit
```
bcoe: will we still get human readable output in the kokoro logs as well? If output is set to a file.

ace-n: Yes, if the `sponge_log.xml` exporting is set up (as it is in this PR for `appengine` and `functions`).

⚠️ **N.B:** we should do one of the following, to avoid breaking _other_ tests:
1. add the XML export config to the top-level Kokoro config
2. only use the `.mocharc.yml` file for `functions` and `appengine` tests.

(The first option is a bit easier to implement, but it would create more UI changes in the short term.)

ace-n:
@fhinkel FYI - @bcoe requested that I apply this change to _all_ tests (as opposed to just the already-labeled tests in `functions` and `appengine`).

This PR will cause two **major** changes:
1. `mocha` runs on Kokoro will **only** output XUnit (i.e. the default human-readable `mocha` output will **not** show up)
1. Kokoro automatically breaks up XML test results into [this nice tree-view UI format](https://source.cloud.google.com/results/invocations/399e2652-eba8-4c19-9525-df9303c64a14/targets/github%2Fnodejs-docs-samples%2Ffunctions%2Fhelloworld/tests), which helps compensate for the lack of human-readable `mocha` output. (The _XUnit logs_ will still be output [as an XML artifact](https://source.cloud.google.com/results/invocations/399e2652-eba8-4c19-9525-df9303c64a14/artifacts/github%2Fnodejs-docs-samples%2Ffunctions%2Fhelloworld).)

@fhinkel  - @bcoe was fine with this for the client libraries, but I want to make sure you're OK with this as well before merging.

.kokoro/.mocharc.yml
ace-n: Choosing option 1 - asked @fhinkel for LGTM below. 🙂

ace-n:
Merging this now, as it's necessary for a demo next week.

@fhinkel if this change becomes an issue for you, **please ping me** so we can roll back and/or work out an alternative solution. 🙂

fhinkel:
Sounds good, let's see if it causes any problems (also with the build cop bot).

PR #1594, opened 2020-01-17
IoT Cloud Client Migration part 4/4

gguuss:


gguuss:
/FYI all IoT tests are green in **d1689e7**

PR #1595, opened 2020-01-17
Add --runtime argument to gcloud deploy in README.

emilyaherbert:
gcloud requires a `--runtime` argument when deploying new functions, otherwise the user sees:

`ERROR: (gcloud.functions.deploy) Missing required argument [runtime]: Flag `--runtime` is required for new functions.`

This addition is consistent with the tutorial (https://cloud.google.com/functions/docs/tutorials/http).

### DIFF
--- a/datastore/functions/README.md
+++ b/datastore/functions/README.md
@@ -25,18 +25,18 @@ Functions for your project.
 
 1. Deploy the "get" function with an HTTP trigger:
 
-        gcloud functions deploy get --runtime [YOUR_RUNTIME} --trigger-http
+        gcloud functions deploy get --runtime [YOUR_RUNTIME] --trigger-http
 
     * Replace `[YOUR_RUNTIME]` with the name of the runtime you are using. For a complete list,
     see the [gcloud reference](https://cloud.google.com/sdk/gcloud/reference/functions/deploy#--runtime).
 
 1. Deploy the "set" function with an HTTP trigger:
 
-        gcloud functions deploy set --runtime [YOUR_RUNTIME} --trigger-http
+        gcloud functions deploy set --runtime [YOUR_RUNTIME] --trigger-http
 
 1. Deploy the "del" function with an HTTP trigger:
 
-        gcloud functions deploy del --runtime [YOUR_RUNTIME} --trigger-http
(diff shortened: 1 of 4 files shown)


emilyaherbert:
@googlebot I signed it!

fhinkel:
👋 @emilyaherbert thanks for opening a pull request. @ace-n Could you have a look please. Thanks.

ace-n:
@labtopia what are your thoughts on this? (We'd have to keep this file up to date.)

@emilyaherbert thanks for the PR! If @labtopia is in favor of this change, would you mind updating the [other files listed here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/search?q=%22gcloud+functions+deploy%22&unscoped_q=%22gcloud+functions+deploy%22) as well? 🙂

cc @grant as FYI, too.

labtopia:
The README is inaccurate as is, so I'm in favor of fixing it. The problem with listing a specific runtime flag is that they can quickly go out of date. If you want to future-proof it, you could put a variable there instead of a hard-coded value. See what we did here:

https://cloud.google.com/functions/docs/deploying/filesystem#deploy_using_the_gcloud_tool

In the table below, we also link to the gcloud reference for the complete list of valid values for --runtime. You might want to do something like that.

grant:
Changing all `gcloud functions deploy` commands in this PR would be great. We should update all READMEs even if they are outdated.

emilyaherbert:
Cool beans, I'll work on this PR!

emilyaherbert:
Updated!

ace-n:
@emilyaherbert Thanks for updating the other files!

@labtopia would it be better to use explicit runtime versions here, given that GitHub docs are usually more focused on "getting things working" than a full conceptual explanation?

labtopia:
>
@labtopia would it be better to use explicit runtime versions here, given that GitHub docs are usually more focused on "getting things working" than a full conceptual explanation?
>

It's a tradeoff. You know this audience better than I do. Specifying explicit runtimes is fine; it just means the docs will go out of date more frequently, and users won't necessarily know where to look for more info when that happens.

ace-n:
OK, got it.

LGTM; @grant I assume you're cool with explicit version names?

PR #1596, opened 2020-01-30
chore: refactor tools.runAsync with execSync

sofisl:


sofisl:
This PR only addresses changes to tools.runAsync(), other commands have not yet been addressed.

PR #1597, opened 2020-01-30
chore: replace repo tools tools.runAsync function with execSync

sofisl:


healthcare/hl7v2/system-test/hl7v2_stores.test.js
```diff
@@ -45,7 +45,7 @@ before(async () => {
   // Create a Pub/Sub topic to be used for testing.
   const [topic] = await pubSubClient.createTopic(topicName);
   console.log(`Topic ${topic.name} created.`);
-  await tools.runAsync(
```
bcoe: love it!

auth/system-test/auth.test.js
```diff
@@ -16,7 +16,7 @@
 
 const path = require('path');
 const assert = require('assert');
-const tools = require('@google-cloud/nodejs-repo-tools');
+const childProcess = require('child_process');
```
fhinkel: If `execSync` is the only property we use, I'd prefer this: 

```
const execSync = require('child_process').execSync;
```

auth/system-test/auth.test.js
```diff
@@ -35,15 +35,15 @@ before(() => {
 });
 
 it('should load credentials implicitly', async () => {
-  const output = await tools.runAsync(`${cmd} auth-cloud-implicit`, cwd);
+  const output = await childProcess.execSync(`${cmd} auth-cloud-implicit`, cwd);
```
fhinkel: `execSync` does not return a promise, so we don't want/need `await`. It's blocking by default.

fhinkel:
`datacatalog` tests are failing independently of your PR. I can't tell at a first glance if the `iot` failure is related. If it keeps failing you might want to submit a PR that doesn't touch those tests. Then we know right away if the failure is related to your changes.

auth/system-test/auth.test.js
sofisl: Sounds good! I'll do that on tests that are entirely covered by execSync.

auth/system-test/auth.test.js
sofisl: perfect, thanks!

auth/system-test/auth.test.js
```diff
@@ -16,7 +16,7 @@
 
 const path = require('path');
 const assert = require('assert');
-const tools = require('@google-cloud/nodejs-repo-tools');
+const {execSync} = require('child_process');
```
fhinkel: 👍

PR #1600, opened 2020-02-04
fix: refactor stubConsole and restoreConsole methods in repo-tools with sinon dependency

sofisl:


appengine/building-an-app/update/test/server.test.js
```diff
@@ -16,15 +16,35 @@
...
   cmd: 'server',
 });
 
-beforeEach(utils.stubConsole);
-afterEach(utils.restoreConsole);
+const stubConsole = function () {
+      sinon.stub(console, `error`);
+      sinon.stub(console, `log`).callsFake((a, b) => {
+        if (
+          typeof a === `string` &&
```
fhinkel: Would the tests pass without this if clause?

sofisl: Unfortunately not (at least for this very specific one). I tried with other ones and they also failed, but definitely interested in making it simpler!

PR #1607, opened 2020-02-17
fix: reduce request usage

JustinBeckwith:
Request is now [deprecated](https://www.npmjs.com/package/request).  This moves the sample over to using `gaxios`, along with other cleanup.

### DIFF
--- a/functions/background/index.js
+++ b/functions/background/index.js
@@ -15,7 +15,7 @@
 'use strict';
 
 // [START functions_background_promise]
-const requestPromiseNative = require('request-promise-native');
+const {request} = require('gaxios');
 
 /**
  * Background Cloud Function that returns a Promise. Note that we don't pass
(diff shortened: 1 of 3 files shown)


functions/background/index.js
```diff
@@ -15,7 +15,7 @@
 'use strict';
 
 // [START functions_background_promise]
-const requestPromiseNative = require('request-promise-native');
+const {request} = require('gaxios');
```
ace-n: Nit: does the leading G in `gaxios` stand for Google here?

(If so, perhaps it should be moved to the `GoogleCloudPlatform` org instead of a personal repo.)

JustinBeckwith: This is a Google owned module :) It lives in the `googleapis` org:
https://github.com/googleapis/gaxios/

ace-n: Ah - I was going off of [the `npm` entry](https://www.npmjs.com/package/gaxios). Maybe we should update that? 🙂

JustinBeckwith: Good catch!  https://github.com/googleapis/gaxios/pull/239

PR #1621, opened 2020-03-10
fix: remove nodejs8 test for appengine analytics sample

michaelawyu:
See also CL 300189444.

The node.js 8 runtime has been deprecated in App Engine; in addition, some of the dependencies in this sample are no longer compatible with nodejs.8 (PR #1557)


michaelawyu:
This PR is ready to merge as CL 300189444 has been submitted.

PR #1672, opened 2020-03-30
OpenTelemetry: add trace example

mayurkale22:
Creates an example of OpenTelemetry exporting traces to Google Cloud Trace.


fhinkel:
Thanks! This quickstart should live in https://github.com/GoogleCloudPlatform/opentelemetry-operations-js.

PR #1712, opened 2020-04-09
fix #1710: async pooling bug in MySQL

### LINKED ISSUE #1710: Fix for MySQL Pooling Async Bug
The NodeJS MySQL server.js code has an async Pool creation call createPool() without any await. So when it is immediately followed by ensureSchema() which runs a query on the pool, it throws an error as the pool isn't yet created:

> UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'query' of undefined

This is the same example on [Google Cloud Docs](https://cloud.google.com/sql/docs/mysql/manage-connections#pools-nodejs) and I assumed them to be correct and wasted 2 days on this thinking something wrong at my end. Please fix this so even any other call can ensure that pool object is properly created before call.

Link to code:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/cloud-sql/mysql/mysql/server.js

shubha-rajan:
Fixes #1710

Ensures that the pool and table are created before serving any requests.


cloud-sql/mysql/mysql/server.js
```diff
@@ -86,18 +86,34 @@ const createPool = async () => {
...
+createPool()
+  .then(() => (schemaReady = ensureSchema()))
+  .catch((error) => console.log(error));
+
+const awaitSchema = async (req, res, next) => {
+  await schemaReady;
+  next();
+};
+
+app.use(awaitSchema);
```
kurtisvg, line 116: Why is there `awaitSchema` and `ensureSchema`? I'm not sure what `next()` is doing. Is it possible to put them in a single function?

fhinkel: 3rd parameter tells the express app to wait for the middlewear to finish rather than to start them all at once. Looks like perfectly valid use of that pattern here.

fhinkel:
I'll wait with the merge until @kurtisvg 's comments are approved

cloud-sql/mysql/mysql/server.js
kurtisvg: Thanks for the explanation!

fhinkel:
@JustinBeckwith do you still have objections for this PR?

fhinkel:
@shubha-rajan Thanks for the fix!

PR #1721, opened 2020-04-16
healthcare API: update hl7v2 samples from v1beta1 to v1

noerog:

PR #1732, opened 2020-04-20
healthcare API: update datasets and DICOM samples from v1beta1 to v1

noerog:
Similar to https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1721.


noerog:
Also had to update some tests due to https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1639.

PR #1744, opened 2020-04-24
Change for...of to for...in

Baystef:
At the busboy.on('finish') event, looping the uploads object with for...of will throw a "uploads is not iterable" error because for...of does not work on objects. The correct for-loop to use here will be for...in


Baystef:
@googlebot I signed it!

PR #1746, opened 2020-04-26
fix(cloud-sql): update unclear configuration in postgres connection sample

### LINKED ISSUE #1697: cloud-sql knex.js related example configuration is inconsistent/confusing
I'm coming from the GCloud Documentation's [Managing database connections](https://cloud.google.com/sql/docs/postgres/manage-connections#pools-nodejs) page about PostgreSQL and Node.js while trying to diagnose connection issues with a live Cloud SQL database. 

The section on how to configure the Knex Client pool left me wondering in some places, however:

```js
// 'createRetryIntervalMillis' is how long to idle after failed connection creation before trying again
knex.client.pool.createRetryIntervalMillis = 200; // 0.2 seconds
```
The explanation recommends implementing an exponential backoff, yet here we set a flat value. Does this mean that Knex cannot be configured to do an exponential backoff when acquiring connections?

Right below it, we have:
```js
// 'acquireTimeoutMillis' is the maximum number of milliseconds to wait for a connection checkout.
// Any attempt to retrieve a connection from this pool that exceeds the set limit will throw an
// SQLException.
knex.client.pool.createTimeoutMillis = 30000; // 30 seconds
```
The comment describes `acquireTimeoutMillis`, yet in the code the `createTimeoutMillis` is set. Which one is right here? Does it make a difference, and if so, what is the difference?

Below, the code then actually sets `acquireTimeoutMillis`:
```js
// 'acquireTimeoutMillis' is the maximum possible lifetime of a connection in the pool. Connections that
// live longer than this many milliseconds will be closed and reestablished between uses. This
// value should be several minutes shorter than the database's timeout value to avoid unexpected
// terminations.
knex.client.pool.acquireTimeoutMillis = 600000; // 10 minutes
```
`acquireTimeoutMillis` really does not sound like a "total lifetime limit", and indeed tarn.js (where the `Pool` implementation from knex.js is imported from) describes it like this in their [documentation](https://github.com/vincit/tarn.js/):
```js
// acquire promises are rejected after this many milliseconds
// if a resource cannot be acquired
acquireTimeoutMillis: 30000,
```
What configuration option would actually give the described "total lifetime" functionality?

The relevant code pieces are all from [cloud-sql/postgres/knex/server.js](cloud-sql/postgres/knex/server.js).
Could someone who is more familiar with the libraries and mechanisms at work here please look over the documentation and the code that goes with it?

Thank you!

shubha-rajan:
fixes #1697


fhinkel:
Please update the PR title to match our commit message convention.

cloud-sql/postgres/knex/server.js
```diff
@@ -75,26 +75,30 @@ const connect = () => {
...
   // [START cloud_sql_postgres_knex_lifetime]
-  // 'acquireTimeoutMillis' is the maximum possible lifetime of a connection in the pool. Connections that
-  // live longer than this many milliseconds will be closed and reestablished between uses. This
-  // value should be several minutes shorter than the database's timeout value to avoid unexpected
-  // terminations.
-  knex.client.pool.acquireTimeoutMillis = 600000; // 10 minutes
-  // [START cloud_sql_postgres_knex_lifetime]
+  // 'idleTimeoutMillis' is the number of milliseconds a connection must sit idle in the pool 
+  // and not be checked out before it is automatically closed.
+  knex.client.pool.idleTimeoutMillis = 600000; // 10 minutes
```
kurtisvg, line 100: Same as the other PR - similar arguments are in different places for other samples. Can we move this into connection count and see if there is a "max lifetime" parameter?

shubha-rajan: There isn't a max lifetime parameter (knex uses the same underlying pool implementation as mssql) so we'll probably have to leave this sample blank like in the mysql sample.

kurtisvg: SGTM - make sure when we update the "Connecting to Cloud SQL" page we include a similar line for SQL Server.

PR #1753, opened 2020-04-27
fix: add missing await for an async function call

rokinsky:


rokinsky:
@googlebot I signed it!

ace-n:
Maybe the better solution here is to not return anything from `blurImages` rather than `await` its result?

cc @fhinkel

rokinsky:
> Maybe the better solution here is to not return anything from `blurImages` rather than `await` its result?

Without the `return` statement in `blurImages`, the result of the function will be `undefined` each time, but we need to `await` its anyway. Let me know If I understand your question correctly.

rokinsky:
@ace-n I could change it to something like this 
```diff
diff --git a/functions/imagemagick/index.js b/functions/imagemagick/index.js
index 730ed3ef..55ddc1a7 100644
--- a/functions/imagemagick/index.js
+++ b/functions/imagemagick/index.js
@@ -49,7 +49,7 @@ exports.blurOffensiveImages = async (event) => {
       detections.violence === 'VERY_LIKELY'
     ) {
       console.log(`Detected ${file.name} as inappropriate.`);
-      return await blurImage(file, BLURRED_BUCKET_NAME);
+      await blurImage(file, BLURRED_BUCKET_NAME);
     } else {
       console.log(`Detected ${file.name} as OK.`);
     }
@@ -102,6 +102,6 @@ const blurImage = async (file, blurredBucketName) => {
 
   // Delete the temporary file.
   const unlink = promisify(fs.unlink);
-  return unlink(tempLocalPath);
+  await unlink(tempLocalPath);
 };
 // [END functions_imagemagick_blur]
```

fhinkel:
@rokinsky Thanks for spotting the missing `await` in the `try` block. Good catch. 

Merging this as is. Feel free to submit a separate PR for the unrelated changes that @ace-n suggested. Thanks!

PR #1755, opened 2020-04-28
fix: spacing for code samples

sofisl:
fixes b/152261768, which is used in a lot of samples


sofisl:
Closing PR per @ace-n's recommendation, will instead mark the bug as obsolete!

PR #1782, opened 2020-05-05
fix(#1781): Handle promise rejections in CloudSQL samples

### LINKED ISSUE #1781: cloud-sql: Tests ignore unhandled promise rejections
I found this issue when adding a new check for #1745 . 
image
Unhandled rejections are also an issue with the App Engine Cloud SQL samples: 
https://source.cloud.google.com/results/invocations/f8c1a2e6-2247-478f-a4e1-0200c5f0f80c/targets/cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fappengine%2Fnode8%2Fcloudsql_postgresql/log

It looks like the problem is that unhandled promise rejections still return a zero exit code, which is interpreted as a build success.

shubha-rajan:
fixes #1781

### DIFF
--- a/cloud-sql/postgres/knex/server.js
+++ b/cloud-sql/postgres/knex/server.js
@@ -147,8 +147,8 @@ const getVoteCount = async (knex, candidate) => {
   return await knex('votes').count('vote_id').where('candidate', candidate);
 };
 
-app.get('/', (req, res) => {
-  (async function () {
+app.get('/', async (req, res) => {
+  try {
     // Query the total count of "TABS" from the database.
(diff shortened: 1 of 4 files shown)


cloud-sql/postgres/knex/server.js
```diff
@@ -182,7 +182,14 @@ app.get('/', (req, res) => {
       voteDiff: voteDiff,
       leaderMessage: leaderMessage,
     });
-  })();
+  }
+  catch(err) {
+    res
```
kurtisvg, line 187: nit: This error isn't being logged

PR #1793, opened 2020-05-07
chore: remove Cloud Jobs v2 samples

fhinkel:
See b/151927859


JustinBeckwith:
For posterity - these samples hit the apiary npm module, and we have a GAPIC module.  We would have to completely rewrite these anyway.

PR #1799, opened 2020-05-13
feat: Create Data Catalog search sample

mesmacosta:
Currently we have inline search samples at:
https://cloud.google.com/data-catalog/docs/how-to/search#data-catalog-how-to-search-nodejs

Sending this PR so tech writer can point the samples to the nodejs-docs-samples github.

Need help at setting an environment variable with the test project Org ID at:
`.kokoro/datacatalog.cfg `
Could one of the reviewers provide the org id?


mesmacosta:
Hi @fhinkel 

Could you help providing the value for the environment variable - Organization ID for the test Project at:
.kokoro/datacatalog.cfg
?

I didn't find any other samples using the Org ID.

fhinkel:
As far as I can tell we don't have an org set up.

mesmacosta:
I will remove the org_id sample from this PR then, and keep just the sample that uses the project scope, I will try to find out internaly if there's a way to set up a test org id.

fhinkel:
Can you use the same project as this test? https://github.com/googleapis/nodejs-security-center/blob/e50061a5f834a356b0221ff01d8de4ed02089260/.kokoro/samples-test.sh#L24

mesmacosta:
Thanks! I will test it out.

mesmacosta:
@fhinkel 
Are you able to whitelist those environment variables?
"
GCLOUD_ORGANIZATION="1081635000895" (No matching key in allowed env vars)
GCLOUD_ORGANIZATION_PROJECT="project-a" (No matching key in allowed env vars)
"

mesmacosta:
ping @fhinkel

fhinkel:
Whitelisted those variables.

PR #1808, opened 2020-05-18
feat: change JS to satisfy Dataflow JS engine

sofisl:
Dataflow doesn't understand "const" or "let"


fhinkel:
lgtm but waiting for review from @leahecole

Issue #1681, opened 2020-03-30
Mocha test results not visible in sponge output


grayside:
The lack of mocha results makes it more difficult to evaluate test failures. This is presumably the result of switching to xunit output in a spong_log.xml file in #1665.

Based on some discussion in https://github.com/mochajs/mocha/pull/2184 to support multiple reporters at once, we should be able to restore console output without significant time overhead.

fhinkel:
I'm wondering why sponge doesn't have the xml file as artifact cc @tbpg

grayside:
When I look at the output, I still don't see the mocha test output. As a result I don't have an easy way to triage the test failure short of locally running the tests. I don't see any further artifacts.

Is there a reason why we can't pursue inline mocha output as part of the overall log output?

Examples include:
* https://g3c.corp.google.com/results/invocations/e5f9038d-e794-48dd-9c64-ead9524ee340/targets/cloud-devrel%2Fnodejs-docs-samples%2Frelease%2Frun%2Flogging-manual/log
* https://g3c.corp.google.com/results/invocations/e5f9038d-e794-48dd-9c64-ead9524ee340/targets/cloud-devrel%2Fnodejs-docs-samples%2Frelease%2Frun%2Flogging-manual/log

grayside:
@fhinkel do you have thoughts on the best path forward? I have a number of issues in the run/logging-manual sample that I can't reproduce locally. I can only assume general flakiness without more information.

fhinkel:
Was this not fixed with https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1711? There should be a separate log file in sponge.

grayside:
1. In poking around, I only saw one log file in the artifacts and it was the same content as the main invocation log.
2. When you say "separate", is the fix that we're going to look in two different files for the end-to-end log output?

fhinkel:
This tests has the correct output: 

https://g3c.corp.google.com/results/invocations/f1ac27e7-10cb-499b-821f-1d662112106a/targets

A lot of other CI runs show the error "XML file not found", and there's no way to see the output. 

Yes, the results from `mocha` are in a separate file (`${project}-sponge.xml`). The buildcop finds that file, otherwise it wouldn't open an issue, but Sponge doesn't find it for a lot of projects 🤔

fhinkel:
This new test also has correct output: 

[...]

PR #1835, opened 2020-05-28
run: use shorter suffix in service names

### LINKED ISSUE #1817: run/hello-broken Test Fails
The `run/hello-broken` tests fail on `master`.

This failure makes this PRs's test fail: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1809

I'm not sure what's going on and don't have access to logs. It makes changes to unrelated PRs fail.

CC: @grayside

grayside:
Fixes #1817 

Every test run that deploys a Cloud Run service should create a distinct deployment to ensure test execution does not conflict. The Kokoro Build ID as a full UUID is too long, causing Cloud Run service name to exceed max length.


grayside:
Trying the force-run one more time. I don't see any output from the build-with-run.sh script, preventing me from doing any troubleshooting. If this still doesn't provide insight, I'll start adding some more debugging code to the script and iterate it here.

.kokoro/build-with-run.sh
```diff
@@ -37,12 +37,12 @@ gcloud config set project $GOOGLE_CLOUD_PROJECT
...
 # Ensures PR-based triggers of the same branch don't collide if Kokoro attempts
-# to run them concurrently.
+# to run them concurrently. Defaults to 'latest'.
 export SAMPLE_VERSION="${KOKORO_GIT_COMMIT:-latest}"
 export SAMPLE_NAME="$(basename $(pwd))"
 
-# Builds not triggered by a PR will fall back to the commit hash then "latest".
-SUFFIX=${KOKORO_BUILD_ID}
+# Cloud Run has a max service name length, $KOKORO_BUILD_ID is too long to guarantee no conflict deploys.
+SUFFIX=$(cat /dev/urandom | LC_CTYPE=C tr -dc 'a-z0-9' | head -c 15)
```
fhinkel, line 45: Can we use `uuid=$(uuidgen)`?   But either way is fine.

averikitsch: Cloud Run has a max name length of 63 characters. This is the fix we are implementing to prevent that since commit sha and build id are usually too long.

averikitsch:
Looks like the script stops after `gcloud config set project $GOOGLE_CLOUD_PROJECT` and right before the `gcloud builds submit`. Is it an issue with kokoro's env and executing the cmd for SUFFIX?

grayside:
Yeah, that's my theory, given the change to suffix generation is the only significant change. I've added `set -x` around it to get more visibility, and will look into local reproduction using the test container next.

grayside:
I'm seeing green locally, so cannot reproduce why this is failing because of #1681.

fhinkel:
I've seen the correct log artifacts in Sponge. The run tests don't have extra artifacts because mocha never runs: 
```sh
> hello-broken@1.0.0 test /tmpfs/src/github/nodejs-docs-samples/run/hello-broken
> echo "Error: no test specified" && exit 0

Error: no test specified
```

The errors from the run script are in the logs:  
```
github/nodejs-docs-samples/.kokoro/build-with-run.sh: line 33: pushd: github/nodejs-docs-samples/run/markdown-preview/editor: No such file or directory
error: line 33 github/nodejs-docs-samples/.kokoro/build-with-run.sh
```

PR #1836, opened 2020-05-30
fix(cloud-sql): update samples to show TCP and Unix Socket connection methods

shubha-rajan:
* Updated both Postgres and MySQL samples to show connection with both TCP and Unix sockets
* Updated READMEs
* Updated tests to test whether connection is successful
* Updated build script to run proxy with both TCP and unix socket connection methods


.kokoro/build.sh
```diff
@@ -99,10 +99,13 @@ if [[ $SQL_CLIENT ]]; then
 	wget --quiet https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy
 	chmod +x cloud_sql_proxy	
 	if [[ $SQL_CLIENT == 'sqlserver' ]]; then
-		./cloud_sql_proxy -instances="${CONNECTION_NAME}"=tcp:1433 &>> cloud_sql_proxy.log &
+		./cloud_sql_proxy -instances="${INSTANCE_CONNECTION_NAME}"=tcp:1433 &>> cloud_sql_proxy.log &
+	elif [[ $SQL_CLIENT == 'mysql' ]]; then
+		export DB_HOST=127.0.0.1:3306
+		./cloud_sql_proxy -instances="${INSTANCE_CONNECTION_NAME}"=tcp:3306 &>> cloud_sql_proxy.log &
 	else
-		mkdir /cloudsql; chmod 777 /cloudsql
```
grayside, line 104: Did something change about needing to pre-create & permission this directory?

cloud-sql/postgres/knex/test/createTable.test.js
```diff
@@ -22,18 +22,28 @@ const {exec} = require('child_process');
 const cwd = path.join(__dirname, '..');
 
 const {DB_USER, DB_PASS, DB_NAME} = process.env;
-const CONNECTION_NAME = process.env.CLOUD_SQL_CONNECTION_NAME;
-
+const CONNECTION_NAME = process.env.INSTANCE_CONNECTION_NAME;
+const DB_HOST = process.env.DB_HOST;
```
grayside, line 26: nit: Why not include this in the destructuring statement above?

cloud-sql/postgres/knex/test/createTable.test.js
```diff
@@ -22,18 +22,28 @@ const {exec} = require('child_process');
...
   try {
     const knex = Knex({
       client: 'pg',
-      connection: {
-        user: DB_USER,
-        password: DB_PASS,
-        database: DB_NAME,
-        host: `/cloudsql/${CONNECTION_NAME}`,
-      },
+      connection: connection,
```
grayside, line 46: I don't recall the practice in this repo, should this use object property shorthand?

cloud-sql/postgres/knex/test/createTable.test.js
```diff
@@ -43,7 +53,7 @@ before(async () => {
 
 it('should create a table', (done) => {
   exec(
-    `node createTable.js ${DB_USER} ${DB_PASS} ${DB_NAME} ${CONNECTION_NAME}`,
+    `node createTable.js ${DB_USER} ${DB_PASS} ${DB_NAME} ${CONNECTION_NAME} ${DB_HOST}`,
```
grayside, line 56: Is there a plan to eventually move away from the CLI mechanism?

.kokoro/build.sh
shubha-rajan: I changed it to run the tests using TCP instead, so there's no need for a socket directory. I noticed that there were issues with writing to the root directory when the tests started failing once I changed them to run against an actual db connection

cloud-sql/postgres/knex/test/createTable.test.js
shubha-rajan: updated in latest commit

cloud-sql/postgres/knex/test/createTable.test.js
shubha-rajan: I updated it in the latest commit, but I'm not sure what the repo standard is.

cloud-sql/postgres/knex/test/createTable.test.js
shubha-rajan: The MySQL and SQL Server samples have an `ensureSchema` method that creates a table if it doesn't exist each time `server.js` is run. Should we switch to using that approach here?

.kokoro/build.sh
grayside: Got it, so we're not doing end-to-end tests where the code is deployed to GAE or Cloud Run?

cloud-sql/postgres/knex/test/createTable.test.js
grayside: I think that would be a good idea but it seems lower priority. Probably something for you/your team to decide.

grayside:
(Note I approved but did not test that the code continues to work)

.kokoro/build.sh
shubha-rajan: Not yet, anyway. Setting up GAE and Cloud Run e2e tests for all Cloud SQL samples is on the roadmap.

cloud-sql/postgres/knex/test/createTable.test.js
kurtisvg: +1 - probably best in a separate PR.

cloud-sql/postgres/knex/server.js
```diff
@@ -44,65 +44,93 @@ const logger = winston.createLogger({
...
-  // something like https://cloud.google.com/kms/ to help keep secrets secret.
-  const config = {
-    user: process.env.DB_USER, // e.g. 'my-user'
-    password: process.env.DB_PASS, // e.g. 'my-user-password'
-    database: process.env.DB_NAME, // e.g. 'my-database'
-  };
+// [START cloud_sql_postgres_knex_create_tcp]
+const connectWithTcp = (config) => {
+  // Extract host and port from socket address
+  const dbSocketAddr = process.env.DB_HOST.split(":")
```
kurtisvg, line 50:
```suggestion
  const dbSocketAddr = process.env.DB_HOST.split(":") // e.g. '127.0.0.1:5432'
```

shubha-rajan:
I just submitted a CL to run the tests for postgres and it looks like they're failing. I'll investigate and fix that before merging.

shubha-rajan:
Tests are passing now! Had to go change the instance name for the database.

PR #1837, opened 2020-06-01
healthcare API: clarify that server returns 200 on resource deletion …

noerog:
…regardless of whether deletion succeeds or fails


feywind:
@fhinkel Are these CI errors something we should look into, or do they happen transiently?

PR #1862, opened 2020-06-08
Add Cloud Healthcare API to Kokoro testing pipeline

noerog:
Part of b/154638640


.kokoro/healthcare/datasets.cfg
```diff
@@ -0,0 +1,13 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Set the folder in which the tests are run
+env_vars: {
+    key: "PROJECT"
+    value: "healthcare/datasets"
+}
+
+# Tell the trampoline which build file to use.
+env_vars: {
```
fhinkel, line 10: can we move this section to common.cfg because it's the same for all the healthcare samples?

fhinkel:
LGTM once the new tests show up on CI and pass.

.kokoro/healthcare/datasets.cfg
noerog: Done.

noerog:
@fhinkel I'm temporarily skipping the IAM tests which gets everything to pass, but I don't understand why the tests can't even recognize the healthcare/hl7v2 directory: https://source.cloud.google.com/results/invocations/78dbc059-289b-4e7c-aef0-4f5ef6a83f0c/targets/cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fhealthcare%2Fhl7v2/log

Do you have any idea why?

fhinkel:
```
allowed_env_vars: {
  key: "PROJECT"
  value: "([a-z]|/|-|_)+$"
}
```

Looks like the numbers in the project name are the problem. I can change the rule to allow numbers, too.

noerog:
> ```
> allowed_env_vars: {
>   key: "PROJECT"
>   value: "([a-z]|/|-|_)+$"
> }
> ```
> 
> Looks like the numbers in the project name are the problem. I can change the rule to allow numbers, too.

Yes please if you could. This has to stay as "hl7v2".

fhinkel:
Fixed, let's run the tests again.

noerog:
All healthcare/ tests are now passing.

fhinkel:
Awesome

fhinkel:
Now we just need to unskip the IAM tests.

noerog:
IAM tests are still failing. Still investigating why.

fhinkel:
> IAM tests are still failing. Still investigating why.

Thanks!

fhinkel:
`Error: Permission healthcare.datasets.setIamPolicy denied on resource projects/nodejs-docs-samples-tests/locations/us-central1/datasets/dataset_4759ee5c_43c8_4c86_a013_beba8ebf31ba (or it may be malformed or not exist)`

I don't see that dataset in console.cloud.google.com. Is it deleted by another test?

noerog:
> `Error: Permission healthcare.datasets.setIamPolicy denied on resource projects/nodejs-docs-samples-tests/locations/us-central1/datasets/dataset_4759ee5c_43c8_4c86_a013_beba8ebf31ba (or it may be malformed or not exist)`
> 
> I don't see that dataset in console.cloud.google.com. Is it deleted by another test?

Yes, the dataset is deleted in https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/healthcare/datasets/system-test/datasets.test.js#L101.

fhinkel:
Are the tests maybe executed in an order such that the dataset is deleted before the test tries to use it?

noerog:
> Are the tests maybe executed in an order such that the dataset is deleted before the test tries to use it?

The tests are executed synchronously and deletion happens at the end. I pushed a commit that had the IAM tests run earlier (right after dataset creation), and the issue still appears.

fhinkel:
@noerog , I checked the logs and changed the permission for the right service account. `healthcare/datasets` is passing now. 

The other healthcare tests are still failing: 

```
Error: Permission healthcare.fhirStores.setIamPolicy denied on resource projects/nodejs-docs-samples-tests/locations/us-central1/datasets/nodejs_docs_samples_test_8be4a9bd_81c7_41a2_b504_3ded06e75f0c/fhirStores/nodejs_docs_samples_test_fhir_storeb5946f8c_2ffa_4fad_ade8_61803a3fcc4e
```

That resource doesn't look right to me. Is is supposed to have `nodej-docs-samples_test_uuid` *twice* in the path?

PR #1867, opened 2020-06-09
Fixed region tag in run/events-pubsub/index.js

cumason123:
Region tag specified run_pubsub_server but the tag should actually specify run_events_pubsub_server


grant:
Merge when green.

PR #1870, opened 2020-06-11
Move composer code out of functions folder

leahecole:
fixes internal bug b/156988611


fhinkel:
The test configs need to be moved, too. And you need to move them in g3. 
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/.kokoro/functions/composer-storage-trigger.cfg

leahecole:
I think the force run started before I submitted the CL - I submitted it now and hopefully things pass 🤞

PR #1879, opened 2020-06-16
fix Markdown Preview flaky tests

kelsk:


run/markdown-preview/editor/render.js
```diff
@@ -32,7 +32,8 @@ const renderRequest = async (markdown) => {
       'Content-Type': 'text/plain'
     },
     body: markdown,
-    timeout: 3000
+    timeout: 3000,
+    retry: 1
```
averikitsch, line 36: Let's not add the retries to the same, only the testing.

kelsk: Done

PR #1895, opened 2020-06-22
chore: update tests to current LTS Node 12

fhinkel:


gguuss:
LGTM for IoT, still some failing tests.

fhinkel:
> My intention was to ensure the testing environment matches the Cloud Run environemt. If we don't want to relax that requirement, this requires updating all Cloud Run Dockerfiles to nodejs12 and confirming that system tests don't break.

Makes sense to keep them in sync. I won't change them. I recommend updating the Cloud Run Dockerfiles to Node 12, as 10 is in maintenance mode.

PR #1897, opened 2020-06-22
chore: update to Node 10 in datastore-functions

fhinkel:


ace-n:
LGTM, but I think @BenWhitehead owns this?

PR #1905, opened 2020-06-26
Billing code sample function change

markmirch:
Updating sample code based on Node 10 runtime environment variable changes: https://cloud.google.com/functions/docs/env-var#environment_variables_set_automatically

### DIFF
--- a/functions/billing/index.js
+++ b/functions/billing/index.js
@@ -19,7 +19,11 @@
 const {google} = require('googleapis');
 const {auth} = require('google-auth-library');
 
-const PROJECT_ID = process.env.PROJECT_ID;
+let PROJECT_ID = process.env.PROJECT_ID;
+// Attempt to use GCP_PROJECT if no PROJECT_ID was specified
+if (typeof PROJECT_ID === 'undefined') {
+  PROJECT_ID = process.env.GCP_PROJECT;
+}
 const PROJECT_NAME = `projects/${PROJECT_ID}`;
 // [END functions_billing_stop]
 // [END functions_billing_limit]
@@ -63,15 +67,7 @@ exports.stopBilling = async (pubsubEvent, context) => {
   }
 
   await _setAuthCredential();
-  let billingEnabled;
-  try {
-    billingEnabled = await _isBillingEnabled(PROJECT_NAME);
-  }
-  catch (e) {
-    billingEnabled = true;
-    console.log('Unable to determine if billing is enabled on specified project, assuming true');
-  }
-  
+  const billingEnabled = await _isBillingEnabled(PROJECT_NAME);
   if (billingEnabled) {
     return _disableBillingForProject(PROJECT_NAME);
   } else {
@@ -107,8 +103,13 @@ const _setAuthCredential = async () => {
  * @return {bool} Whether project has billing enabled or not
  */
 const _isBillingEnabled = async (projectName) => {
-  const res = await billing.getBillingInfo({name: projectName});
-  return res.data.billingEnabled;
+  try {
+    const res = await billing.getBillingInfo({name: projectName});
+    return res.data.billingEnabled;
+  } catch (e) {
+    console.log('Unable to determine if billing is enabled on specified project, assuming billing is enabled');
+    return true;
+  }
 };
 
 /**
--- a/functions/billing/index.js
+++ b/functions/billing/index.js
@@ -19,7 +19,7 @@
 const {google} = require('googleapis');
 const {auth} = require('google-auth-library');
 
-const PROJECT_ID = process.env.GCP_PROJECT;
+const PROJECT_ID = process.env.PROJECT_ID;
 const PROJECT_NAME = `projects/${PROJECT_ID}`;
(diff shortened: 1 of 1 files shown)


functions/billing/index.js
```diff
@@ -57,9 +57,22 @@ exports.stopBilling = async (pubsubEvent, context) => {
...
+    return `No project specified`;
+  }
 
   await _setAuthCredential();
-  if (await _isBillingEnabled(PROJECT_NAME)) {
+  let billingEnabled;
+  try {
+    billingEnabled = await _isBillingEnabled(PROJECT_NAME);
+  }
+  catch (e) {
```
grant, line 70: Usually the catch is on the same line in Node:

`} catch (err) {`

functions/billing/index.js
```diff
@@ -57,9 +57,22 @@ exports.stopBilling = async (pubsubEvent, context) => {
...
+    billingEnabled = true;
+    console.log('Unable to determine if billing is enabled on specified project, assuming true');
+  }
+  
+  if (billingEnabled) {
     return _disableBillingForProject(PROJECT_NAME);
   } else {
     return 'Billing already disabled';
```
grant, line 78: Logic could be simplified:

```js
try {
  if (await _isBillingEnabled(PROJECT_NAME)) {
    return _disableBillingForProject(PROJECT_NAME);
  } else {
    return 'Billing already disabled';
  }
} catch (e) {
  console.log('Unable to determine if billing is enabled on specified project, assuming true');
}
```

functions/billing/index.js
markmirch: Sounds good, will do!

functions/billing/index.js
markmirch: I believe that's slightly different logic, as I wanted to make sure _disableBillingForProject would still be called even if _isBillingEnabled threw an error. I thought about doing a try/catch/finally but figured this would be easier to read. Thoughts?

grant: Looking further, I would assume that the try/catch should be within the `_isBillingEnabled` function.

It should either return `true`/`false` and never through an error.

Depends on if the sample wants catch API errors or not.

functions/billing/index.js
```diff
@@ -57,9 +57,22 @@ exports.stopBilling = async (pubsubEvent, context) => {
...
 
   await _setAuthCredential();
-  if (await _isBillingEnabled(PROJECT_NAME)) {
+  let billingEnabled;
+  try {
+    billingEnabled = await _isBillingEnabled(PROJECT_NAME);
+  }
+  catch (e) {
+    billingEnabled = true;
+    console.log('Unable to determine if billing is enabled on specified project, assuming true');
```
ace-n, line 72: Nit: `assuming true` = awkward phrasing?

markmirch: Updating to "Unable to determine if billing is enabled on specified project, assuming billing is enabled"

functions/billing/index.js
markmirch: Makes sense, I'll move the try/catch.

functions/billing/index.js
grant: Still needs to be formatted?

functions/billing/index.js
```diff
@@ -19,7 +19,11 @@
 const {google} = require('googleapis');
 const {auth} = require('google-auth-library');
 
-const PROJECT_ID = process.env.PROJECT_ID;
+let PROJECT_ID = process.env.PROJECT_ID;
+// Attempt to use GCP_PROJECT if no PROJECT_ID was specified
+if (typeof PROJECT_ID === 'undefined') {
+  PROJECT_ID = process.env.GCP_PROJECT;
+}
```
grant, line 26: More idiomatic:

```js
const PROJECT_ID = process.env.PROJECT_ID || process.env.GCP_PROJECT;
```

However, within GCF, we can rely on just the `GCP_PROJECT` if I'm not mistaken: https://cloud.google.com/functions/docs/env-var

markmirch: I was thinking of making the logic more explicitly readable but I can change if that's preferred.

Can't depend on GCP_PROJECT for Node 10: https://cloud.google.com/functions/docs/env-var#nodejs_10_and_subsequent_runtimes

markmirch: Or maybe it makes more sense to just tell them to set GCP_PROJECT manually? I was under the impression that would throw an error but it only seems to for older runtimes.

grant: I would just use the code snippet I suggested.
It's the same logic and should be safe in case we only use one or the other env vars.

I don't know where this snippet is added in the docs, but we could check there too to see what other languages do.

markmirch: Understood, I'll simplify the code a bit and explain in the docs: https://cloud.google.com/billing/docs/how-to/notify

ace-n: We've dropped support for Node versions < 10 in our docs.

I'd drop the `GCP_PROJECT` here.

PR #1918, opened 2020-07-10
test(run): add shallow end-to-end tests to cloud run sample apps.

### LINKED ISSUE #1916: run: add minimal e2e test coverage to sample apps
Each Cloud Run sample app should have minimal e2e test coverage that establishes the following:

* Container builds
* Container deploys to Cloud Run
* The nodejs service inside the container is able to start and respond to requests

For minimal testing we are not concerned with whether the logic of the node.js service is correct, as long as a 200 or 400 response is received we can be confident it was produced by the running node.js service.

grayside:
Fixes #1916
This includes #1915, which is kept broken out to facilitate review.

The approach used in this PR involves a lot of shell script copying. A new approach is being worked on, which I hope will be used to drive simplification here.

cc @jsimonweb @dmahugh

### DIFF
--- a/run/events-pubsub/test/deploy.sh
+++ b/run/events-pubsub/test/deploy.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 Google LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail;
+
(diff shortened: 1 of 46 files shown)


run/events-pubsub/test/deploy.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 Google LLC.
```
fhinkel, line 3: I think those should all be 2020

grayside: Fixed.

grayside:
From the OP, there's an upcoming project which will allow us to delete most of the redundant code, my preference is to duplicate code and have this light e2e testing rather than building a more DRY solution only to delete.

run/pubsub/test/url.sh
```diff
@@ -0,0 +1,30 @@
...
+
+requireEnv() {
+  test "${!1}" || (echo "Environment Variable '$1' not found" && exit 1)
+}
+
+requireEnv SERVICE_NAME
+
+set -x
+gcloud run services \
+  describe "${SERVICE_NAME}" \
```
grant, line 27: nit: I'd expect the full `gcloud` method on one line, then flags can be separate lines.

```sh
gcloud run services describe ${SERVICE_NAME} \
  ...
```

PR #1928, opened 2020-07-19
fix: update run_events_pubsub to also console log

grant:
Log message in addition to returning the response, as per the design doc.

### DIFF
--- a/run/events-pubsub/app.js
+++ b/run/events-pubsub/app.js
@@ -23,8 +23,9 @@ app.post('/', (req, res) => {
     ? Buffer.from(pubSubMessage.data, 'base64').toString().trim()
     : 'World';
   
-  console.log(`Hello, ${name}! ID: ${req.get('ce-id') || ''}`);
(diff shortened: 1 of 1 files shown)


run/events-pubsub/app.js
```diff
@@ -23,6 +23,7 @@ app.post('/', (req, res) => {
     ? Buffer.from(pubSubMessage.data, 'base64').toString().trim()
     : 'World';
   
+  console.log(`Hello, ${name}! ID: ${req.get('ce-id') || ''}`);
   res.send(`Hello, ${name}! ID: ${req.get('ce-id') || ''}`);
```
grayside, line 27: Should we DRY up the response values? Should there be a comment noting the message is sent in the response for local testing?

```suggestion
  const msg = `Hello, ${name}! ID: ${req.get('ce-id') || ''}`;
  console.log(msg);
  res.send(msg);
```

averikitsch: +1

PR #1942, opened 2020-08-06
fix: standardize run events pub/sub message

grant:
Consoles out the messages in addition to returning the HTTP response.

Makes the developer experience a bit better when using Stackdriver and localhost.

---

Tested locally and with a sample event on Run.


grant:
Merging the minor change.
Failed tests are unrelated.

PR #1949, opened 2020-08-19
[Cloud SQL] Update Readme

averikitsch:
Changes:

1. Code uses `INSTANCE_CONNECTION_NAME` and not `CLOUD_SQL_CONNECTION_NAME` which is listed in the ReadMe.
2. ReadMe was missing arguments to `createTable.js`

### DIFF
--- a/cloud-sql/postgres/knex/README.md
+++ b/cloud-sql/postgres/knex/README.md
@@ -2,10 +2,10 @@
 
 ## Before you begin
 
-1. If you haven't already, set up a Node.js Development Environment by following the [Node.js setup guide](https://cloud.google.com/nodejs/docs/setup)  and 
+1. If you haven't already, set up a Node.js Development Environment by following the [Node.js setup guide](https://cloud.google.com/nodejs/docs/setup)  and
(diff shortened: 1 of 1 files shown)


kurtisvg:
I think we want to actually use `CLOUD_SQL_CONNECTION_NAME` and not `INSTANCE_CONNECTION_NAME` but @shubha-rajan can correct me which way we decided

averikitsch:
Also just saw `$CLOUD_SQL_INSTANCE_NAME` in the package.json.

shubha-rajan:
So we're not actually consistent across languages for this yet. The Python and Go samples currently use `INSTANCE_CONNECTION_NAME` while Java samples and these Node samples use `CLOUD_SQL_CONNECTION_NAME`. I think `CLOUD_SQL_CONNECTION_NAME` will make testing easier in the long run since it's more specific (in case we need env vars for instances of other things in the testing environment). I can go through and change the other samples that currently use `INSTANCE_CONNECTION_NAME`. 

@averikitsch do you mind changing the variable name in the package.json since I'm pretty sure `CLOUD_SQL_INSTANCE_NAME` is wrong both ways?

PR #1962, opened 2020-08-26
chore(functions/firebase): remove "data" property

ace-n:


ace-n:
@grant see b/162554971 internally.

(As noted there, this was likely due to a change in the runtime at some point. _Where_ that change occurred, I have no idea...)

PR #1971, opened 2020-09-10
feat: add event type conversion to run pubsub event sample

grant:
- Adds event type to Run Pub/Sub sample for IDE autocompletion.


grayside (approved):

run/events-pubsub/app.js
```diff
@@ -4,6 +4,7 @@
 
 // [START run_events_pubsub_handler]
 const express = require('express');
+const {toMessagePublishedEvent} = require('@google/events/cloud/pubsub/v1/MessagePublishedData');
```
ace-n, line 7: Nit - I don't imagine someone can do something like this?

`const {toMessagePublishedEvent} = require('@google/events/cloud/pubsub/v1').MessagePublishedData;`

(I ask because this format makes reverse engineering/"exploring" the "type definition universe" a bit easier, via Node's built-in debugging tools.)

run/events-pubsub/app.js
```diff
@@ -20,9 +21,10 @@ app.post('/', (req, res) => {
     console.log(`Bad Request: ${errorMessage}`);
     return;
   }
-  const pubSubMessage = req.body.message;
-  const name = pubSubMessage.data
-    ? Buffer.from(pubSubMessage.data, 'base64').toString().trim()
+  // Cast to MessagePublishedEvent for IDE autocompletion
+  const pubSubMessage = toMessagePublishedEvent(req.body);
```
ace-n, line 25: Nit: would `message` be too non-descript here? (It's shorter 🙂 )

ace-n (approved):

run/events-pubsub/app.js
grant: Yup, we can do that. Going to hold off for right now though.

run/events-pubsub/app.js
grant: It would be non-descript.

I'm not going to write `data.message.data`.

_(Flashbacks to [`data.data.data`](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1474))_

grant (commented):
Thanks for the reviews! Going to add the types later.

There are going to be a couple future changes to the Node types repo, and I'll look back to this PR later.

PR #1977, opened 2020-09-17
fix: Add node version in app.yaml in appengine/endpoints

dafnarosenblum:
Deploying this example outputs this error:
```
ERROR: (gcloud.beta.app.deploy) INVALID_ARGUMENT: Invalid runtime 'nodejs' specified. Accepted runtimes are: [php, php55, python27, java, java7, java8, go111, go112, go113, go114, java11, nodejs10,
nodejs12, php72, php73, php74, python37, python38, ruby25, ruby26, ruby27]
```
After this change the deploy command succeeds.

### DIFF
--- a/appengine/endpoints/app.yaml
+++ b/appengine/endpoints/app.yaml
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: nodejs
+runtime: nodejs12
 env: flex
 
 beta_settings:


dafnarosenblum:
@googlebot I signed it!

JustinBeckwith (approved):

appengine/endpoints/app.yaml
```diff
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: nodejs
+runtime: nodejs12
```
JustinBeckwith, line 14: Thanks for the fix!  I'm a little surprised this was failing with the current `app.yaml`, because this one is using `env: flex`.  The rules for valid runtimes (at least last time I checked) were different here.  

Just to double check - are you using `env: flex` in your `app.yaml`?

dafnarosenblum: Good catch, no, I was using `env: standard`.

PR #2009, opened 2020-10-16
fix: update eventarc storage output

grant:
Fixes output for events sample.


bryeung:
LGTM thanks!

ace-n:
@fhinkel seems like there are SQL-related failures - would you or @kurtisvg mind investigating?

PR #2018, opened 2020-10-22
chore(functions): remove unused signed url sample

ace-n:


grant (approved):
We're remembering to remove the tests if there are some, right?

ace-n:
The Node.js tests, yes. (I presume the Kokoro tests would fail if we didn't, anyway.)

ace-n:
@kurtisvg the Cloud SQL GAE tests are failing...is that something you can fix?

(cc @fhinkel in case we should merge as-is instead.)

kurtisvg:
> @kurtisvg the Cloud SQL GAE tests are failing...is that something you can fix?

The samples in `/appengine` are owned by the Serverless team. Our team owns the ones in `/cloudsql`, which appear to all be passing. 

If there are an issue with Cloud SQL tests going forward, please open an issue and assign it to myself or GoogleCloudPlatform/cloud-sql-dpes and we'll prioritize it against our other existing work.

ace-n:
@kurtisvg ah, thanks for clarifying!

@shubha-rajan said we could disable the offending tests for now until they're fixed. Will re-run them.

PR #2024, opened 2020-10-27
chore: use node-fetch instead of gaxios

fhinkel:


BenWhitehead (approved):
Seems fine to me, do you agree @crwilcox?

fhinkel:
See See b/171857357.

PR #2025, opened 2020-10-27
chore: use gts

fhinkel:
Fix typo in .eslintrc.json (had an extra t before).
    
Fix one usage of an invalid export. 

Convert `it.only` to `it.skip` for failing tests in iot, see https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/2026
    
Disable a bunch of rules to get this to pass. We'll need to turn them on
and fix their usages manually. Exclude `run/idp-sql/` for now from linting, too many errors.


grant (dismissed):
It's generally recommended to use parentheses even for single-parameter arrow functions.

https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

Otherwise looks fine for `functions/`

kurtisvg (approved):
LGTM for `/cloudsql`

BenWhitehead (approved):
LGTM for `datastore/`

leahecole (approved):
lgtm for composer

.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./node_modules/gts/",
+  "env": {
+    "mocha": true
+  },
+  "rules": {
+    "node/no-missing-require": ["off"], 
+    "node/no-unpublished-require": ["off"], 
+    "no-unused-vars": ["off"],
```
ace-n, line 9: Ditto - why is this disabled?

.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./node_modules/gts/",
+  "env": {
+    "mocha": true
+  },
+  "rules": {
+    "node/no-missing-require": ["off"],
```
ace-n, line 7: What's the reason for disabling this?

ace-n (changes requested):

ricalo (approved):
LGTM for speech-to-speech/

JustinBeckwith:
@grant on the parens rule - I'm onboard with changing it, but we should do it upstream in `gts` as a default.  Feel free to enable it here broadly if you like, but we should try to follow the google style guide best as we can over in gts.

JustinBeckwith (approved):

package.json
```diff
@@ -13,8 +13,8 @@
   },
   "private": true,
   "scripts": {
-    "lint": "eslint '**/*.js'",
```
JustinBeckwith, line 16: Can we also remove the direct dependencies on `eslint`, `prettier`, and any other `eslint` plugins we're using and now don't need?

.eslintrc.json
ace-n: Ah NVM, saw your comment above - but can you file a GitHub issue for tracking?

.eslintrc.json
ace-n: @fhinkel mind adding this to said (tracking) GitHub issue too?

grant:
> @grant on the parens rule - I'm onboard with changing it, but we should do it upstream in `gts` as a default.  Feel free to enable it here broadly if you like, but we should try to follow the google style guide best as we can over in gts.

Upstream in gts sounds good 👍

I don't think we'd want all of the function samples to show up in the non-paren style and then switch back at some point later, so maybe we can change the PR to add a rule or first change gts?

.eslintrc.json
fhinkel: See PR description

.eslintrc.json
fhinkel: See PR description

package.json
fhinkel: good catch, thanks.

fhinkel:
We'd be rolling out this change in 60+ libraries. I think it's better to have consistency among our samples than to discuss a particular style.

ace-n (approved):

PR #2032, opened 2020-10-29
secret manager: add example rotation cloud function

tam7t:
Add new sample for Secret Manager: https://cloud.google.com/secret-manager/docs/how-to

Demonstrate using a cloud function triggered through pub/sub from cloud scheduler to rotate a secret on a pre-defined interval.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -47,6 +47,10 @@ export FUNCTIONS_TOPIC=integration-tests-instance
 export FUNCTIONS_BUCKET=$GOOGLE_CLOUD_PROJECT
 export FUNCTIONS_DELETABLE_BUCKET=$GOOGLE_CLOUD_PROJECT-functions
 
+# functions/secret-rotation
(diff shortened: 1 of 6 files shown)


ace-n (changes requested):
Can we move this out of the `functions` folder and into its own separate folder? (e.g. `secret-manager/functions`)

_Context: the `functions` folder is reserved for samples used in `cloud.google.com/functions`._

tam7t:
> Can we move this out of the `functions` folder and into its own separate folder? (e.g. `secret-manager/functions`)
> 
> _Context: the `functions` folder is reserved for samples used in `cloud.google.com/functions`._

Sure thing, actually as I look into it more maybe this belongs in the client repo: https://github.com/googleapis/nodejs-secret-manager

ace-n:
@tam7t I'll leave that call to @bcoe and @fhinkel 🙂

fhinkel:
@tam7t Yes, this definitely belongs in the client library repo. Thanks!

Issue #2047, opened 2020-11-05
Unused (and untested) functions in run/idp-sql


fhinkel:
There are some unused functions in `run/idp-sql`. 
Some are not covered by region tags so we should be able to delete them. Those inside region tags should have test coverage and thus not show up as unused. 

```
$ npm run lint
nodejs-docs-samples/run/idp-sql/static/config.js
  2:7  warning  'config' is assigned a value but never used  no-unused-vars

nodejs-docs-samples/run/idp-sql/static/firebase.js
  22:10  warning  'signIn' is defined but never used  no-unused-vars
  40:16  warning  'vote' is defined but never used    no-unused-vars
 ```

averikitsch:
All of these variables and functions are used but organized specifically into different files for client side code. Can I opt these files out?

fhinkel:
We can opt out (which we're doing right now with `.eslintrc` in the `run/idp-sql` folder. Thanks for clarifying, I didn't realize they're used in fronttend code. Ideally, we would have tests for those functions, but then we'd need a bundler and it would probably convolute the sample too much.

PR #2050, opened 2020-11-05
chore: delete unused functions

fhinkel:
see https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/2047

`signOut` and `toggle` are not used, not exported, and not within region tags.


run/idp-sql/static/firebase.js
```diff
@@ -38,25 +38,6 @@ function signIn() {
 // [END run_end_user_firebase_sign_in]
 // [END cloudrun_end_user_firebase_sign_in]
 
-function signOut() {
```
averikitsch, line 41: These functions are used in `index.html`

averikitsch (changes requested):

run/idp-sql/static/firebase.js
fhinkel: 💡 Thanks for clarifying. Added them back in.

averikitsch (approved):

Issue #2001, opened 2020-10-09
lib request is not supported


joelviel:
Hi,

Can you update this example https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/iot/http_example with a maintained library for http requests ?

Thanks

fhinkel:
Hi @joelviel 👋 . Thanks for raising the issue. @hongalex would you have time to look into this? Thanks

hongalex:
Upon discussion with @fhinkel, we will update the sample to use `node-fetch`.

PR #2061, opened 2020-11-08
chore: use fetch instead of request

fhinkel:
Fixes: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/2001
internal b/171857357

### DIFF
--- a/iot/http_example/cloudiot_http_example.js
+++ b/iot/http_example/cloudiot_http_example.js
@@ -16,7 +16,7 @@
 // [START iot_http_includes]
 const fs = require('fs');
 const jwt = require('jsonwebtoken');
-const request = require('retry-request', {request: require('request')});
+const fetch = require('node-fetch');
 // [END iot_http_includes]
 
 // Create a Cloud IoT Core JWT for the given project ID, signed with the given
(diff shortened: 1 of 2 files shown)


iot/http_example/cloudiot_http_example.js
```diff
@@ -16,7 +16,7 @@
 // [START iot_http_includes]
 const fs = require('fs');
 const jwt = require('jsonwebtoken');
-const request = require('retry-request', {request: require('request')});
```
JustinBeckwith, line 19: I am pretty sure Gus moved these samples all over to here:
https://github.com/googleapis/nodejs-iot/blob/master/samples/http_example/cloudiot_http_example.js

Because I remember making a very similar PR 😆 Not blocking this PR, but it looks like we need to reconcile the samples left behind here with the ones in the client lib repo.

JustinBeckwith (approved):

iot/http_example/cloudiot_http_example.js
fhinkel: ohhh, let me check the region tags and delete the duplicate samples 😃

PR #2068, opened 2020-11-11
[App Engine] TS : Adding the missing node type definition 

Alutun:
The Typescript example is actually not working.
When App Engine is running the `npm run gcp-build`

It will throw erros like this one : 

`Step` #6 - "builder": index.ts(17,21): error TS2580: Cannot find name 'process'. 

The package.json is missing the node Types definitions module. 

My pull request is fixing that. 

Best


Alutun:
@googlebot I signed it!

fhinkel (approved):

michaelawyu (approved):

PR #2097, opened 2020-12-08
fix: Update test runner for Cloud Run logging-manual

### LINKED ISSUE #2096: run/logging-manual: reducing test failure noise
When this sample's first test fails, it means something is wrong with the deployment and all tests will fail. On current implementation, this means a flakey deployment will file 3-4 issues.

Instead, the following steps:

1. Run the test with the --bail option, so if that test fails we do not both checking the logs.

2. Remove 'Cloud Run Log Metadata' case, this is simply checking that Cloud Run's standard metadata is added to log entries. This is more of a product test than a sample test.

averikitsch:
Fixes #2096 #2080 #2079 #2078 #2077


fhinkel (approved):
LGTM, thank you!

PR #2099, opened 2020-12-08
chore(functions): actually support env vars in target URL

ace-n:


functions/helloworld/test/sample.integration.http.test.js
```diff
@@ -19,8 +19,8 @@ const path = require('path');
 const requestRetry = require('requestretry');
 const uuid = require('uuid');
 
-const PORT = 9010;
-const BASE_URL = `http://localhost:${PORT}`;
+const PORT = process.env.PORT || 8010;
```
grant, line 22: I prefer 8080 if we are going to change the PORT.

grant (approved):

functions/helloworld/test/sample.integration.http.test.js
ace-n: The docs and other samples use `8010` - any objection to keeping it consistent?

grant: It looks like the docs only has Node right now:

https://cloud.google.com/functions/docs/testing/test-http#integration_tests

Let's use 8080, that is the default of the FF (makes it easier to run locally).

https://github.com/GoogleCloudPlatform/functions-framework#specification-summary

And the point of the sample isn't about customizing ports.

grant: Python uses 8080, I think Java might too.

ace-n: Got it. Changed this to `8080`.

ace-n:
FYI - the environment variable multiplexing strategy I used (`sed`-ing in the correct value for the system tests) has been done before (in other DPE repos), but feels like a hack.

@fhinkel PTAL, and LMK if I should do something more fault-tolerant (maybe using different env var names for each? In the past, we wanted to point out the similarities between **integration** and **system** tests, but I don't think they're that similar anymore.)

fhinkel (approved):

Issue #2014, opened 2020-10-21
App Engine Cloud SQL tests are failing


averikitsch:
Since Oct 16th, the SQL and postgreSQL sample tests have been failing.

[Test fusion example](https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fappengine%2Fnode10%2Fcloudsql)

shubha-rajan:
It looks like only the appengine-cloudsql tests are failing. The tests for the cloud sql samples are working: 
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fcloudsql-mysql
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fcloudsql-postgres
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fcloudsql-sqlserver

@ace-n is making the failing checks non-required while we figure out what's going on.

ace-n:
I've marked the associated GitHub checks as non-required, but don't have bandwidth to investigate further this week.

@engelke is this something you can look into?

fhinkel:
@engelke @shubha-rajan Who is looking at those test failure?

engelke:
From what I can see, it seems the tests are working for Node 10 and 12. The failures for Node 8 seem to be test configuration, not the sample code itself, and have been happening since June. Somebody familiar with the Kokoro Node test configuration will have to look into this.

averikitsch:
The tests are now passing though there doesn't seem to be any changes for them breaking or passing?
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fappengine%2Fnode10%2Fcloudsql
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fnodejs-docs-samples%2Fpresubmit%2Fappengine%2Fnode10%2Fcloudsql_postgresql
[...]

Issue #1763, opened 2020-04-29
Run system tests on rennovate PR's proposing major version updates


grayside:
Issue #1759 and the cluster addressed by PR #1680 were the result of merging a dependency update carrying a major version bump, which proceeded to break the system tests on run/logging-manual.

Would it be possible to run system tests on major version updates? This is creating escalated P1 bug work instead of P2 internal cleanup work.

ace-n:
N.B: `renovate` PRs can typically have higher latency than human-written ones (in part because they're rarely, if ever, urgent) - so the attendant slowdown due to system tests is probably more palatable _even if_ we run system tests on _all_ `renovate` PRs.

fhinkel:
Yes, we should run all tests on presubmit, just need to enablet those...
Anybody have spare cycles to add those test configs?

ace-n:
To be clear - @fhinkel you're saying we should run system tests on _all_ PRs, or only _all `renovate` ones_?

(IIRC we _may_ have disabled "system tests on all PRs" in the past because they were too slow. My argument is that that's less important for `renovate` PRs than human-generated ones.)

grayside:
Python is taking the approach of running Cloud Run system tests on every PR that touches it. Cloud Run build & deploy is generally 1-2 minutes overhead.

fhinkel:
I'm in favor of running system tests on all PRs. But we have severa failing P1 system tests, so I can't turn them on on PRs. cc @JustinBeckwith

fhinkel:
> 
> (IIRC we _may_ have disabled "system tests on all PRs" in the past because they were too slow. 
I don't remember running systems tests on PRs since moving to Kokoro.

JustinBeckwith:
With nodejs client libraries we run system tests on presubmit.  If it's reasonable to do (I understand it isn't always), that would be my preference.  If we have failing system tests, shouldn't those be getting picked up by build-cop-bot?  I'm far more concerned that we have failing tests.

grayside:
Do we have a decision on how to move forward here?

fhinkel:
Oh, I see we don't have open issues about system tests, now would be a good time to turn them on for PRs. Let me work on that!

fhinkel:
[...]

PR #2105, opened 2020-12-08
build: add system test config for event arc & functions

sofisl:
After looking at #1763 and #2014 I decided to do an audit on which folders were missing system test configs.

I *think* that only eventarc is not running its system tests, but I could be wrong. 

Let me know if there are any other system tests we need to configure. If/when this approved, I'll also submit a CL for the google3 configs.


appengine/hello-world/flexible/app.js
```diff
@@ -29,6 +29,7 @@ app.listen(PORT, () => {
   console.log(`App listening on port ${PORT}`);
   console.log('Press Ctrl+C to quit.');
 });
+
```
ace-n, line 32: Nit: revert this?

ace-n (approved):

.kokoro/build.sh
```diff
@@ -119,5 +128,8 @@ if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"release"* ]]; then
 fi
 
 npm test
```
fhinkel, line 130: This seems like the wrong approach. Let's use `npm run --if-present system-test` instead of making this project specific.

fhinkel (approved):
I guess we can run on system tests one by one as we figure out the correct env variables. And once they're all turned on, change it to `npm run --if-present system-test`. 

Thanks for working on this!

sofisl:
I might recommend just having us run `test`, and converting all scripts that run `test` to run unit and system tests, as is the current convention (like this PR does). What do you think?

grant (approved):

ace-n:
Personally, I'd vote for _only_ running "sufficiently fast" (< ~30 seconds) tests with `npm test`, and putting "slower" tests under some sort of `npm run system-test` command. (This speeds up local development, IMHO.)

fhinkel:
After looking into this with Sofia, almost all samples already run something like `"test": "npm run unit-test && npm run system-test"`. We should follow that conversion for `eventarc`, too.

sofisl:
This PR is ready to go, seems like we're just snagging on a flaky test

.kokoro/build.sh
```diff
@@ -46,6 +46,7 @@ export SENDGRID_API_KEY=$(cat $KOKORO_GFILE_DIR/secrets-sendgrid-api-key.txt)
 export FUNCTIONS_TOPIC=integration-tests-instance
 export FUNCTIONS_BUCKET=$GOOGLE_CLOUD_PROJECT
 export FUNCTIONS_DELETABLE_BUCKET=$GOOGLE_CLOUD_PROJECT-functions
+export BASE_URL="https://$GCF_REGION-$GOOGLE_CLOUD_PROJECT.cloudfunctions.net/"
```
fhinkel, line 49: I believe this change is breaking the `functions/speech-to-speech tests because of this line: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/220c3f4a90bb06c434c9886f3c987148272c4462/functions/speech-to-speech/functions/test/index.test.js#L29

functions/helloworld/test/sample.system.http.test.js
```diff
@@ -16,8 +16,22 @@
 const assert = require('assert');
 const Supertest = require('supertest');
 const supertest = Supertest(process.env.BASE_URL);
+const childProcess = require('child_process');
 
 describe('system tests', () => {
+  // [END functions_http_system_test]
```
fhinkel, line 22: Did we get START and END mixed up?

sofisl: No, I'm adding an extra END line and an extra START line so that the before and after test hooks don't get included in  the sample (see line 34 as well). It's not originally included in the sample so I wanted to make sure I didn't change it too much!

fhinkel: ah ok. Thanks. At a first glance it looks like End/Start/end. I didn't see the first start.

.kokoro/build.sh
sofisl: Thank you for helping me debug this!!

PR #2112, opened 2020-12-11
chore(App Engine standard): update samples to Node 14

engelke:


fhinkel:
Should we wait until the docs are updated? https://cloud.google.com/appengine/docs/standard/nodejs/runtime

fhinkel (changes requested):
The package.json files should be updated as well: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/appengine/analytics/package.json

fhinkel (approved):
thanks!

PR #2117, opened 2020-12-16
Update index.js

AlexWick:
Update copyleft ;)


AlexWick:
@googlebot I signed it!

grant:
Sorry, we don't expect the Copyright year to be the current year or to be updated without rationale. Please file an issue if you think this is incorrect and we can re-evaluate.

PR #2120, opened 2020-12-30
feat: add cloud workflows quickstart

grant:
Inits canonical Cloud Workflows API quickstart sample.

- Adds a README for new folder, `workflows/quickstart`
- This sample...
  - Executes the Cloud Workflows quickstart (`myfirstworkflow`) using the Workflows client library
  - Awaits the results with exponential backoff
  - Includes test
    - `GOOGLE_CLOUD_PROJECT=my-project npm test`

Example test result:

```
  Cloud Workflows Quickstart Tests
Created execution: projects/392236462496/locations/us-central1/workflows/myFirstWorkflow/executions/685a01f4-3cae-4dcb-b0b4-c2965f95fc1a
Poll every second for result...
- Waiting for results...
Execution finished with state: SUCCEEDED
["Wednesday","Wednesday Night Wars","Wednesday 13","Wednesday Addams","Wednesday Campanella","Wednesdayite","Wednesday Martin","Wednesday Campanella discography","Wednesday Night Hockey (American TV program)","Wednesday Morning, 3 A.M."]
    ✓ should execute the quickstart (2196ms)


  1 passing (2s)
```

CC: @krisbraun


grant:
CI failures are unrelated.

workflows/quickstart/test/quickstart.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+const quickstart = require('../index');
+const {assert} = require('chai');
+
+const PROJECT_ID = process.env.GOOGLE_CLOUD_PROJECT;
+const LOCATION_ID = 'us-central1';
+const WORKFLOW = 'myFirstWorkflow';
+
+describe('Cloud Workflows Quickstart Tests', () => {
+  it('should execute the quickstart', async () => {
```
grayside, line 25: This test and the HTTP requests inside the workflow don't appear to have any retry, meaning there is a flakiness risk related to the availability of the Cloud Function and Wikipedia API.

If there's something we can do to improve test reliability, it would be good.

workflows/quickstart/test/quickstart.test.js
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const quickstart = require('../index');
+const {assert} = require('chai');
```
grayside, line 18: Not sure if we tend to use chai by default, but I don't see chai providing much unique value over the built-in assert library.

grant: Sounds good. Removed chai.

workflows/quickstart/test/quickstart.test.js
grant: Yes, all system calls and network calls can fail.

After inspecting similar examples, I've added a `try` / `catch`.

For the Wikipedia workflow: The workflow handles retries of the workflow execution, not the client library.

grant:
Thanks for the review @grayside, PTAL again.

grant:
`run/system-package` failure is unrelated.

workflows/quickstart/index.js
```diff
@@ -0,0 +1,86 @@
...
+    return console.error('ERROR: GOOGLE_CLOUD_PROJECT is required.');
+
+  // [START workflows_api_quickstart]
+  const {ExecutionsClient} = require('@google-cloud/workflows');
+  const client = new ExecutionsClient();
+
+  // Execute workflow
+  try {
+    const createExecutionRes = await client.createExecution({
+      parent: client.workflowPath(projectId, location, workflow),
```
averikitsch, line 48: nit: save parent as a variable

workflows/quickstart/index.js
```diff
@@ -0,0 +1,86 @@
...
+    while (!executionFinished) {
+      const [execution] = await client.getExecution({
+        name: executionName,
+      });
+      executionFinished = execution.state !== 'ACTIVE';
+
+      // If we haven't seen the result yet, wait a second.
+      if (!executionFinished) {
+        console.log('- Waiting for results...');
+        await sleep(backoffDelay);
```
averikitsch, line 66: if a user copies this from the docs, they won't have your sleep function

workflows/quickstart/index.js
grant: There's no reason to add a new variable/const, as the expression is not re-used.

workflows/quickstart/index.js
grant: Good catch. Fixed.

grant (commented):
Thanks, replied to comments with updated code.

workflows/quickstart/index.js
averikitsch: Right, but it may be more readable split apart. Preference is up to you.
```suggestion
      const parent = client.workflowPath(projectId, location, workflow);
      const createExecutionRes = await client.createExecution({parent});
```

grant: Yes, it could be argued either way. Less variables on the stack / in autocompletion results one way.

grayside: Since this is a quickstart the emphasis is not on readability to the same degree as a snippet, however the (often forgotten) rubric guideline is to avoid nesting complex initialization logic inside objects.

grayside (approved):

averikitsch (approved):

grant:
Can an admin merge this PR?

@ace-n looks like there's a flaky Functions test.

ace-n:
I'll try re-running the tests for now, and merge if they continue to be flaky.

Issue #1991, opened 2020-10-01
New Cloud Run end-to-end deployment facility


grayside:
Cloud Run's current facility to deploy end-to-end tests is a bundle of BASH scripts. They seem very prone to flakes, and have generally proven difficult to maintain. #1740 is just the latest case of a flaky test.

A new approach to deploying the samples is needed, one which de-emphasizes BASH. Step 1 I'll put together a proposal.

fhinkel:
@grayside Thanks! Do you want to keep this as P1?

grayside:
Has Cloud Run flakiness been blocking PRs? If not this can be a P2.

averikitsch:
List of samples to update:

- [ ] hello-broken
- [ ] helloworld
- [x] idp-sql
- [x] image-processing
- [ ] logging-manual
- [x] markdown-preview
  - [x] editor
  - [x] renderer
- [ ] pubsub
- [ ] system-package

averikitsch:
@grayside for the PubSub sample tests, can I mock the Pub/Sub message by sending a POST request or should I test the user journey by setting up the Pub/Sub topic and sub and checking the logs for the message? The user journey tests seems like it could be flaky due to the need to check the logs.

averikitsch:
All samples have been updated.

averikitsch:
reopening as there are test failures

averikitsch:
The last set of test errors were due to Cloud Build quotas

grayside:
In discussion with Averi, #2124 doesn't fix Cloud Build quotas, so we are seeing two different classes of flakiness. Cloud Build Quotas will be handled in a follow-up.

averikitsch:
Note: There has not been any test failures for the Run samples since the mocha timeouts extension PR was merged. Cloud build quotas have not been an issue, but could be if more Run samples are added or if multiple PRs are testing at the same time.

averikitsch:
Cloud Build flakiness has not been an issue since the record error in December. Closing this issue and will open a new one if tests continue to be flaky.

PR #2124, opened 2021-01-06
fix: extend timeout for flaky tests

averikitsch:


grayside:
Per #1991, the hypothesis was Cloud Build quotas. Does this really fix the issue?

grayside (approved):

PR #2127, opened 2021-01-07
chore(functions): Move tests to Node 14 image

ace-n:
~Part of~ Fixes **b/171596650**.


grant (dismissed):
But GCF Node 14 is not available to the public. We should not test samples against 14 if the public can't use it.

engelke (approved):
LGTM

ace-n:
Let's hold off on this for now. (Googlers - see the bug mentioned above for more details.)

fhinkel (approved):
LGTM once GCF Node.js 14 is GA.

PR #2135, opened 2021-01-20
fix dates diff in secs

cuonghuynh:


cloud-tasks/app/createTask.js
```diff
@@ -60,7 +60,7 @@ const createHttpTaskWithToken = async function (
   if (convertedDate < currentDate) {
     console.error('Scheduled date in the past.');
   } else if (convertedDate > currentDate) {
-    const date_diff_in_seconds = (convertedDate - currentDate) / 1000;
+    const date_diff_in_seconds = (convertedDate.getTime() - currentDate.getTime()) / 1000;
```
averikitsch, line 63: Thank you for your PR; however .getTime() should not be needed, since subtracting Date objects returns the number of milliseconds since the UNIX epoch.

fhinkel: I agree witih @averikitsch  and will close this.

PR #2142, opened 2021-01-28
feat: add kuberun eventing samples

eclipselu:


fhinkel:
Thanks. Looks like the test configs in `.kokoro` are missing. And you need a different prefix for the region tags.

eclipselu:
thanks! I removed region tags for now as we don't use them yet, `.kokoro` test configs are added

fhinkel:
Thanks. I don't see that the tests ran. Can you link the matching CL please. 

Please also add yourself to the Codeowners file for these samples.

eclipselu:
closing as this is not needed any more.

PR #2160, opened 2021-02-09
chore(functions): remove unused exception handler

ace-n:
The relevant Slack API client library method directly throws an exception anyway. (See b/171900608 for details.)

### DIFF
--- a/functions/slack/index.js
+++ b/functions/slack/index.js
@@ -99,8 +99,8 @@ const verifyWebhook = req => {
     body: req.rawBody,
   };
 
-  // This function throws an exception if an incoming request is invalid.
-  !verifyRequestSignature(signature);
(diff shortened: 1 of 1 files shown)


functions/slack/index.js
```diff
@@ -99,11 +99,8 @@ const verifyWebhook = req => {
     body: req.rawBody,
   };
 
-  if (!verifyRequestSignature(signature)) {
-    const error = new Error('Invalid credentials');
-    error.code = 401;
-    throw error;
-  }
+  // This function throws an exception if an incoming request is invalid.
+  !verifyRequestSignature(signature);
```
grant, line 103: Remove the bang (!)?

ace-n: Good catch, done.

fhinkel (approved):

PR #2161, opened 2021-02-09
fix(functions): add fully qualified domain name for metadata sample

iennae:
Due to differences in implementation, there can be issues if
the metadata server isn't fully qualified in code. This
minor fix resolves that issue.

Related b/154412754.

### DIFF
--- a/functions/security/index.js
+++ b/functions/security/index.js
@@ -23,7 +23,7 @@ const PROJECT_ID = 'my-project-id';
 const RECEIVING_FUNCTION = 'myFunction';
 
 // Constants for setting up metadata server request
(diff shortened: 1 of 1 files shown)


functions/security/index.js
```diff
@@ -26,7 +26,7 @@ const RECEIVING_FUNCTION = 'myFunction';
 // See https://cloud.google.com/compute/docs/instances/verifying-instance-identity#request_signature
```
grant, line 26: With this change, can you link the correct metadata server URL documentation?

https://cloud.google.com/functions/docs/securing/function-identity#identity_tokens

grant (changes requested):
One requested change. Looks good otherwise.

Was this tested manually?

grant (approved):

PR #2162, opened 2021-02-11
chore(functions): add node termination example

ace-n:
Related to b/176872447

### DIFF
--- a/functions/concepts/index.js
+++ b/functions/concepts/index.js
@@ -104,18 +104,3 @@ exports.makeRequest = async (req, res) => {
   res.sendStatus(externalRes.ok ? 200 : 500);
 };
 // [END functions_concepts_requests]
-
-exports.nodeTermination = async (event, context, callback) => {
-  // [START functions_concepts_node_termination]
-  // These will cause background tasks to stop executing immediately
-  return 1; // Returning a value
-  return (await Promise.resolve()); // Returning the result of a promise
-  return (await Promise.reject()); // Same behavior as resolved promises
-  callback(); // Invoking the callback function
-
-  // These will wait until the related background task finishes
-  return Promise.resolve(); // Returning the promise itself
-  return Promise.reject(); // Same behavior as to-be-resolved promises
-  setTimeout(callback, 10); // Invoke callback once all operations complete
-  // [END functions_concepts_node_termination]
-};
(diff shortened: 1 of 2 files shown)


functions/concepts/index.js
```diff
@@ -104,3 +104,18 @@ exports.makeRequest = async (req, res) => {
   res.sendStatus(externalRes.ok ? 200 : 500);
 };
 // [END functions_concepts_requests]
+
+exports.nodeTermination = async (event, context, callback) => {
+  // [START functions_concepts_node_termination]
```
grant, line 109: What region tag is this? Is this a new sample?

grant (changes requested):
Please add a PR description.

functions/concepts/index.js
ace-n: Yes - see b/176872447

functions/concepts/index.js
```diff
@@ -104,3 +104,18 @@ exports.makeRequest = async (req, res) => {
   res.sendStatus(externalRes.ok ? 200 : 500);
 };
 // [END functions_concepts_requests]
+
+exports.nodeTermination = async (event, context, callback) => {
```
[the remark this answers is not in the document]
ace-n: We call the `callback` function, so yes - we do need it.

functions/concepts/termination.js
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+/* eslint-disable */
+
+exports.nodeTermination = async (event, context, callback) => {
```
grant, line 19: We cannot have both `async` and `callback` in the same function. The Functions Framework won't resolve this correctly.

Either write:

```js
exports.nodeTermination = async (event, context) => {
```

or

```js
exports.nodeTermination = (event, context, callback) => {
```

grant (changes requested):

functions/concepts/termination.js
ace-n: To be clear, this function is not actually meant to be run. (It's purely there to serve as a "wrapper" for the lines inside.)

(I can add a comment to that effect, if you want.)

grant: Just remove `callback` or remove `async`. If we need both, we need 2 style of functions.

ace-n: Done.

grant (approved):

PR #2176, opened 2021-02-18
fix(run): Update Logging test with retries

averikitsch:
Fixes: #2169 #2170 #2171 #2172 #2173 #2174 #2175

### DIFF
--- a/run/logging-manual/test/system.test.js
+++ b/run/logging-manual/test/system.test.js
@@ -62,6 +62,10 @@ const dateMinutesAgo = (date, min_ago) => {
   return date.toISOString();
 };
 
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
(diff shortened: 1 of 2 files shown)


run/logging-manual/test/system.test.js
```diff
@@ -144,18 +148,26 @@ describe('Logging', () => {
...
-          requestLog = entry;
-        } else {
-          sampleLog = entry;
-        }
-      });
+      let entries;
+      let attempt = 0;
+      const maxAttempts = 5;
+      while ((!requestLog || !sampleLog) && attempt < maxAttempts) {
+        await sleep(attempt * 30000); // Linear backoff between retry attempts
```
grayside, line 155: Future: It seems like the retry logic is different in each repo. No objection there for now, but something to flag in case we notice different levels of flakiness in the future.

grayside (approved):

PR #2198, opened 2021-03-15
Fix opencensus test

yuriatgoogle:
Modifying test to use content-type, rather than response code, to confirm that the app is responding.


yuriatgoogle:
Fixes  #2196

fhinkel (approved):

PR #2210, opened 2021-03-23
chore: update CODEOWNERS for serverless

grant:
- Updates CODEOWNERs for various serverless folders
- Use updated GitHub team name: https://github.com/orgs/GoogleCloudPlatform/teams/cdpe-serverless

This prevents an admin being a required reviewer for folders.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -7,13 +7,15 @@
 
 # Kokoro
 .kokoro @fhinkel
-.kokoro/appengine @michaelawyu
+.kokoro/appengine @GoogleCloudPlatform/cdpe-cbr-containers
 .kokoro/functions @ace-n @grant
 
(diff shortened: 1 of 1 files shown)


ace-n (commented):
Is it better to use specific people, given that we do that for `functions`?

(If not, maybe we can have a GAE-specific GitHub team instead?)

**Also - do not forget to update `python-docs-samples` too.** (That's the _only_ other repo that needs to be updated, AFAICT.)

grant:
I think we want to have team ownership in general, rather than an individual.

It's harder to update individual usernames in git source than just an alias. Definitely can update this for other languages too.

I didn't touch `functions/` but probably could in this PR too.

grant:
I'm going to hold off on this change for now.

PR #2213, opened 2021-03-29
chore(functions): update node-fetch invocation

ace-n:


fhinkel (approved):

functions/tips/index.js
```diff
@@ -94,8 +94,15 @@ const httpsAgent = new https.Agent({keepAlive: true});
  */
 exports.connectionPooling = async (req, res) => {
   try {
-    const {data} = await fetch('/', {httpAgent, httpsAgent});
-    res.status(200).send(`Data: ${data}`);
+    const url = 'https://www.example.com/';
```
grant, line 97: Can we add a TODO here? A developer has to replace this URL.

functions/tips/index.js
```diff
@@ -94,8 +94,15 @@ const httpsAgent = new https.Agent({keepAlive: true});
...
+
+    // Select the appropriate agent to use based on the URL.
+    const agent = url.includes('https') ? httpsAgent : httpAgent;
```
grant, line 100: This constant could probably be removed by inlining it.

grant (approved):
LGTM. Some comments that could improve the PR.

functions/tips/index.js
ace-n: I wanted to call it out explicitly, which is why I didn't do that.

functions/tips/index.js
ace-n: Not necessarily - the URL provided will work as-is.

That said, I'll add an optional TODO.

PR #2216, opened 2021-04-05
Fix environment variable used to get GCP project

tcolloca:
The documentation says to use GCP_PROJECT, however in the code it is being used GOOGLE_CLOUD_PROJECT.


tcolloca:
Documentation at https://cloud.google.com/billing/docs/how-to/notify#selectively_control_usage should also be updated.

PR #2218, opened 2021-04-08
refactor: use node's native promisified fs

MoSattler:
[docs](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_unlink_path)


MoSattler:
@googlebot I signed it!

grant (approved):

fhinkel (approved):
good catch

PR #2220, opened 2021-04-13
Add repo-owners to code owners

fhinkel:
Add @fhinkel and @sofisl explicitely as codeowners for all folders so
that we can turn on "Require review from Code Owners".

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,31 +2,30 @@
 # for more info about CODEOWNERS file
 
 # Repo owner
-* @sofisl
-* @fhinkel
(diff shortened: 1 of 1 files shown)


ace-n (changes requested):
Nit: would it be better to have a dedicated GitHub team for this, to make future updates easier?

(Maybe something like `@GoogleCloudPlatform/nodejs-docs-samples-owners`?)

fhinkel:
@ace-n done. PTAL

ace-n (approved):

PR #2222, opened 2021-04-19
Add comment about authenticated push JWT token validation.

fayssalmartanigcp:
This snippet is used in the Cloud Pub/Sub docs (https://cloud.google.com/pubsub/docs/push#validating_tokens) and many users are not aware that signature verification of the token is not enough, the claim needs to be validated also.

### DIFF
--- a/appengine/pubsub/app.js
+++ b/appengine/pubsub/app.js
@@ -122,7 +122,7 @@ app.post('/pubsub/authenticated-push', jsonBodyParser, async (req, res) => {
     // by signature and audience verification above, including:
     //   - Ensure that `claim.email` is equal to the expected service
     //     account set up in the push subscription settings.
-    //   - Ensure that `calim.email_verified` is set to true.
+    //   - Ensure that `claim.email_verified` is set to true.
 
     claims.push(claim);
   } catch (e) {
--- a/appengine/pubsub/app.js
+++ b/appengine/pubsub/app.js
@@ -118,12 +118,10 @@ app.post('/pubsub/authenticated-push', jsonBodyParser, async (req, res) => {
 
     const claim = ticket.getPayload();
 
-    // IMPORTANT: user code must validate claim details in addition
-    // to signature verification above:
+    // IMPORTANT: you should validate claim details not covered
+    // by signature and audience verification above, including:
     //   - Ensure that `claim.email` is equal to the expected service
     //     account set up in the push subscription settings.
-    //   - Ensure that `claim.aud` is equal to the audience set up in
-    //     the push subscription settings.
     //   - Ensure that `calim.email_verified` is set to true.
 
     claims.push(claim);
--- a/appengine/pubsub/app.js
+++ b/appengine/pubsub/app.js
@@ -117,6 +117,15 @@ app.post('/pubsub/authenticated-push', jsonBodyParser, async (req, res) => {
     });
 
     const claim = ticket.getPayload();
+
+    // IMPORTANT: user code must validate claim details in addition
+    // to signature verification above:
+    //   - Ensure that `claim.email` is equal to the expected service
+    //     account set up in the push subscription settings.
+    //   - Ensure that `claim.aud` is equal to the audience set up in
+    //     the push subscription settings.
+    //   - Ensure that `calim.email_verified` is set to true.
+
     claims.push(claim);
   } catch (e) {
     res.status(400).send('Invalid token');


ace-n (commented):
I'll defer to @anguillanneuf on this, as I'm not really familiar with this sample.

(Perhaps @fhinkel knows about this too.)

fhinkel:
ideally I'd like to see a code sample of the steps the user code must do.

fayssalmartanigcp:
> ideally I'd like to see a code sample of the steps the user code must do.

I don't believe I can provide a working sample since the user needs to setup the subscription settings (service account and audience) to run these sample as stated here: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/appengine/pubsub#readme. The actual code would be either: test these values for equality between token and subscription config and/or for users running in GCP, use IAM to check for permissions on the service account.

appengine/pubsub/app.js
```diff
@@ -117,6 +117,15 @@ app.post('/pubsub/authenticated-push', jsonBodyParser, async (req, res) => {
...
 
     const claim = ticket.getPayload();
+
+    // IMPORTANT: user code must validate claim details in addition
+    // to signature verification above:
+    //   - Ensure that `claim.email` is equal to the expected service
+    //     account set up in the push subscription settings.
+    //   - Ensure that `claim.aud` is equal to the audience set up in
+    //     the push subscription settings.
+    //   - Ensure that `calim.email_verified` is set to true.
```
anguillanneuf, line 127:
```suggestion
    // IMPORTANT: you should validate claim details not covered
    // by signature and audience verification above, including:
    //   - Ensure that `claim.email` is equal to the expected service
    //     account set up in the push subscription settings.
    //   - Ensure that `calim.email_verified` is set to true.
```

anguillanneuf (changes requested):
@fayssalmartanigcp, thanks for adding the comment! 

`audience` is an [optional field](https://googleapis.dev/nodejs/google-auth-library/latest/interfaces/VerifyIdTokenOptions.html) that one can set when verifying the JWT. I think the sample correctly shows that.
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/a97dcfd13def94274adeb6ac1b25c7ab923b46e4/appengine/pubsub/app.js#L114-L117

As for the email in the decoded claims, I don't think `google-auth-library` currently checks it. See Python's [`jwt.decode()`](https://googleapis.dev/python/google-auth/1.28.1/reference/google.auth.jwt.html?highlight=jwt#google.auth.jwt.decode), [`id_token.verify_oauth2_token()`](https://googleapis.dev/python/google-auth/1.28.1/reference/google.oauth2.id_token.html#google.oauth2.id_token.verify_oauth2_token). So users indeed need to manually check the email in the decoded claims. (@busunkim96 shall we provide a convenient way for users to check claim emails in our auth libraries?)

appengine/pubsub/app.js
```diff
@@ -117,6 +117,15 @@ app.post('/pubsub/authenticated-push', jsonBodyParser, async (req, res) => {
     });
 
     const claim = ticket.getPayload();
+
+    // IMPORTANT: user code must validate claim details in addition
+    // to signature verification above:
+    //   - Ensure that `claim.email` is equal to the expected service
+    //     account set up in the push subscription settings.
+    //   - Ensure that `claim.aud` is equal to the audience set up in
+    //     the push subscription settings.
```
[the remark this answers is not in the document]
fayssalmartanigcp: Makes sense.

anguillanneuf (approved):

appengine/pubsub/app.js
engelke: Typo "calim"

fhinkel:
> > ideally I'd like to see a code sample of the steps the user code must do.
> 
> I don't believe I can provide a working sample since the user needs to setup the subscription settings (service account and audience) to run these sample as stated here: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/appengine/pubsub#readme. The actual code would be either: test these values for equality between token and subscription config and/or for users running in GCP, use IAM to check for permissions on the service account.

Okay!

fhinkel (approved):

engelke:
I think the shared secret TOKEN in the request query parameters, which is checked here, serves the same purpose. But it's useful to let people know there is more information you can take advantage of.

engelke (approved):
LGTM

fayssalmartanigcp:
> I think the shared secret TOKEN in the request query parameters, which is checked here, serves the same purpose. But it's useful to let people know there is more information you can take advantage of.

The TOKEN in the request query is not equivalent to the identity in the JWT token (`email` or `sub`). The reason is that any third party can technically set up a subscription to the endpoint URL and include a secret "token" as part of that URL (if the token is leaked for example).

For the `email` identity in the JWT token, there is an authorization check preventing malicious user from setting a service account email they don't have permission to as part of the push subscription config.

Basically the URL token method relies on secrecy of the token (can be spoofed probably) and is **_not recommended by itself_**, while the second relies on IAM checks + signature validation (the service account identity can be public and still can't be impersonated).

PR #2263, opened 2021-06-16
fix(functions/helloworld): stop using request

JustinBeckwith:
This eliminates several dependencies we didn't need, while moving from the deprecated `request` => `gaxios` for tests.


ace-n (changes requested):
This replacement seems OK to me at a high-level. The tests are failing though.

functions/helloworld/test/sample.system.storage.test.js
```diff
@@ -55,18 +54,20 @@ describe('system tests', () => {
     });
 
     // Wait for logs to become consistent
-    await promiseRetry(retry => {
+    for (let i = 0; i < 10; i++) {
```
grant, line 57: Exponential backoff for 10 cycles seems pretty arbitrary. Could we have a finally case in we keep slipping?

grant (approved):
Like Ace mentioned, fine in theory. Approved!

functions/helloworld/test/sample.system.storage.test.js
JustinBeckwith: I used 10 because that's what `promise-retry` defaulted to 🤷 anyhow, I decided this was changing too many things at once so I backed that part of the CL out and just focused on getting rid of request.

ace-n: Nit: if we're using `gaxios`, could we move the retry logic into that library (or perhaps a `gaxios-retry` library "shim") directly?

JustinBeckwith:
aiight @ace-n this was a fun adventure.  I think I figured out why we were seeing some of the flakiness in this test suite.  This change now includes:
- Explicitly waiting for functions-framework to start serving before making requests using `wait-port`
- Explicitly ending the functions-framework processes instead of relying on the bash timeout trick
- Eliminating retries, because as far as I can tell they were in there to deal with the flakiness introduced by the timeout and sleep approaches used above

While figuring this out, I decided to drop down to raw node.js (not using a child process wrapper) so I could eliminate variables while debugging.  We can discuss putting something like that or execa back if needed, but I kind of like the clarity of the pure node.js usage (though there is a little repetition).

functions/helloworld/test/sample.integration.pubsub.test.js
```diff
@@ -14,76 +14,100 @@
...
+      'helloPubSub',
+      '--signature-type',
+      'event',
+      '--port',
+      PORT,
+    ]);
+    const ffProcHandler = new Promise((resolve, reject) => {
+      let stdout = '';
+      let stderr = '';
+      ffProc.stdout.on('data', data => (stdout += data));
```
ace-n, line 50: Aside: @grant I'm curious what you'd think about adding a "shim" feature to the FFs.

(This would start other functions without requiring a separate process.)

If this sounds interesting to you, LMK and I'll look into drafting a proposal. 👍

ace-n (approved):
PR itself LGTM, but maybe we should consider formally supporting this workflow in the FFs themselves.

PR #2270, opened 2021-06-20
fix(functions/billing): remove usage of request module

JustinBeckwith:
The `request` module is fully deprecated, and we shouldn't be using it.  This transitions us to `gaxios` for tests, which we actively support.


grant (approved):

ace-n (approved):
LGTM once tests pass.

PR #2282, opened 2021-07-05
fix(functions/imagemagick): drop dependency on request

JustinBeckwith:


ace-n (changes requested):
Two concerns I have:
1) Tests are failing. 🙂
2) ~While I'm not opposed to refactors like this ^, we'll have to either a) **apply them to all `functions` samples** or b) **file issues as reminders**.~ Looks like #2263 is related to this.

~_^: this is more about the `wait-port` refactor than the `request` --> `gaxios` one._~

JustinBeckwith:
@ace-n at a glance, it looks like the failing tests are largely unrelated to the code I touched.  I assume these are flakes, since the functions/imagemagick tests are passing?

JustinBeckwith:
Oh and on the other concern - I am slowly removing `request` from all the samples since it's deprecated.  The `wait-port` changes are coming at the same time, because as far as I can tell the retry logic used in retry-request was largely there to deal with ambiguity on if functions framework was ready yet.  The `wait-port` pattern is more clear, and negates the need for retries in most of these cases.

ace-n:
Yeah, looks like they were flakes. (It may have been because the system test deployments don't use UUIDs, and collided with each other. This is somewhat hard to fix, per b/191301443 - but maybe it's worth doing since this isn't the first time it happened?)

JustinBeckwith:
@ace-n I think the real answer here is to only run tests for samples that changed during a PR.  I have a little experiment rolling here with Cloud Build (see the healthcare dicom test), but it's going to take a bit of time to work out the details.  In the interim...  is there anything else you want me to do with this PR?

ace-n (approved):
If `functions/imagemagick` is passing, I think we can merge this. :+1:

PR #2292, opened 2021-07-14
feat: add functions/esm sample

grant:
## Functions ESM Sample

Adds a sample showing how to use ES modules within Google Cloud Functions for the Node.js 12+ runtime.

### Features

- `README.md`
  - Sample title, summary, prereqs
  - Running instructions
  - Testing instructions
- `const.js` file showing how to use `export` keyword.
- `index.js` file showing how to import a file and export a user function.
- `package.json` showing how to use `"type": "module"`
- Test

This sample would be a section of the Node.js Runtime doc showing how to use the popular Node feature of ES modules.

### DIFF
--- a/functions/esm/index.js
+++ b/functions/esm/index.js
@@ -1,0 +1,23 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START functions_esm]
+import {MY_CONST} from './consts.js';
+
+export const esm = (req, res) => {
+  res.send(`Hello, ${MY_CONST}!`);
+};
+// [END functions_esm]
--- a/functions/esm/.eslintrc.yml
+++ b/functions/esm/.eslintrc.yml
@@ -1,4 +1,7 @@
 ---
 parserOptions:
   sourceType: module
-
+rules:
+  no-unused-vars: 0
+  node/no-unsupported-features/es-syntax: 0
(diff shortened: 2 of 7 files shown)


bcoe (approved):
This is slick 😄

ace-n (approved):
LGTM once tests pass.

I assume this will go on the [Node.js Runtime page](https://cloud.google.com/functions/docs/concepts/nodejs-runtime)?

ace-n (changes requested):
FYI - @iennae is looking for a firm owner for this.

While I think the sample itself is worth adding, we need to confirm (as a team) that we have a plan to support it.

functions/esm/index.js
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START functions_esm]
```
grayside, line 17: @bcoe I recall you saying the Node.js libraries weren't ready to support ESM yet, is that still the case? It seems okay if functions gets out ahead of the rest of Cloud, but we'll want to make the caveats highly visible if so.

In that case, I think this would need comments with that warning, in addition to carefully highlighting in the docs.

I assume if this is working the Functions Framework is ESM-ready?

grant:
@JustinBeckwith Hi Justin, I see you added the `do not merge` label, but didn't comment as to why, which is pretty surprising. Why was it added? Can I remove the label?

JustinBeckwith:
I entirely defer this one to @ace-n

ace-n:
Hi folks,

@iennae and myself are working on figuring out an ownership/maintenance plan for this sample.

Let's hold off on merging this until we (the Serverless DPE team) have such a plan approved and in place.

iennae:
Hey, thanks everyone. Just adding the flag back so we are clear to not merge currently. 

We appreciate the sample Grant. Serverless DPE will do the same due diligence as with all external sample requests. We'll get back to you after we review with the team on next steps.

grant:
Alright, thanks. Will wait. LMK if I can do anything to help support this and our developers.

grant:
Per sync, we'll lead developers who want to learn more about ESM with Cloud Functions to the Functions Framework and docs.

PR #2299, opened 2021-07-27
feat(functions): add CloudEvents-based PubSub sample

ace-n:
@grayside LMK if I should add a second (Node.js-focused) reviewer.


JustinBeckwith (approved):
Can confirm this is JavaScript

functions/v2/.gitignore
```diff
@@ -0,0 +1 @@
+*.zip
```
grayside, line 1: question: Why is this needed?
suggestion: Can it be added to the root gitignore? Scattered ignore files can be confusing.

functions/v2/test/index.test.js
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const assert = require('assert');
+const {request} = require('gaxios');
+const {exec} = require('child_process');
+const waitPort = require('wait-port');
+
+const startFF = async (target, signature, port) => {
```
grayside, line 20: suggestion: I tend to feel system tests using the function framework locally have a high overhead relative to their advantage over unit tests. Would kicking off v2 samples be a good moment to switch to Deploy tests?

grayside (approved):
No blocking feedback. Optional suggestions to use or file for follow-up at your discretion.

Comments attempting to follow conventionalcomments.org

functions/v2/.gitignore
ace-n: Currently, the GCF V2 deployment process sometimes generates extra `.zip` files.

This is a "convenience feature" to prevent us accidentally including them in commits.

functions/v2/test/index.test.js
ace-n: I believe the Functions Frameworks are directly integrated into [all versions of] GCF, so this will give us a medium-high fidelity test.

I agree with your sentiment, but I think we need to be more organized/deliberate about how we add deployment tests. (In particular, do we want to write deployment tests in every language - or have some common "translation" layer so we only need write them in one language [e.g. PHP, which already has deployment tests]?)

PR #2307, opened 2021-08-04
Update Cloud SQL Postgres sample to create table on app start

jsimonweb:
Fixes #2306

PR #2316, opened 2021-08-18
build: enable cloud build for healthcare/dicom

JustinBeckwith:


healthcare/dicom/system-test/dicom_stores.test.js
```diff
@@ -42,21 +41,16 @@ const dcmFileName = 'IM-0002-0001-JPEG-BASELINE.dcm';
...
 before(async () => {
-  assert(
-    process.env.GOOGLE_CLOUD_PROJECT,
-    'Must set GOOGLE_CLOUD_PROJECT environment variable!'
-  );
-  assert(
-    process.env.GOOGLE_APPLICATION_CREDENTIALS,
-    'Must set GOOGLE_APPLICATION_CREDENTIALS environment variable!'
-  );
+  projectId = await storage.getProjectId();
```
noerog, line 52: Is this storage.getProjectId() just because the Storage client is being used in this file? As opposed to google.auth.getProjectId().

bcoe:
@JustinBeckwith any reason we can't merge this? seems to have run the dicom test.

We're currently blocking on this test [here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2307).

JustinBeckwith:
@bcoe this PR shouldn't be blocking anything - I am guessing that's a result of Cloud Build not re-running the trigger after the branch transition.  I am guessing that pushing to that other branch any change will kickstart the cloud build job.  That having been said, we could totally merge this now with no side effects.

healthcare/dicom/system-test/dicom_stores.test.js
JustinBeckwith: Yup, it calls the same underlying method.  I used `storage.getProjectId()` because we already had a storage client, and it just exposes the underlying auth method.  This method is available on all client libraries.

feywind (approved):

PR #2326, opened 2021-09-01
chore(functions/logging): exclude trace code from sample

ace-n:
**Partial fix** for b/183751712


JustinBeckwith (approved):

ace-n:
@grayside went with the solution we discussed; PTAL?

grayside (approved):

PR #2331, opened 2021-09-03
chore(functions/v2): move Pub/Sub sample into its own directory

ace-n:
**DO NOT MERGE until @labtopia approves.**

In case we need to do more samples for GCF v2, we want to make sure that the directory structure allows for that context instead of putting everything in v2 directly.


iennae (approved):
LGTM

grayside (approved):

ace-n:
We're not actively working on this due to our upcoming launch schedule.

PR #2333, opened 2021-09-07
Add Secret Manager integration to Cloud SQL sample apps

iainfoulds:
This PR adds the option to retrieve the Cloud SQL database password from Secret Manager.

If a secret version is defined in an environmental variable (`CLOUD_SQL_CREDENTIALS_SECRET`), the secret value is retrieved from Secret Manager and then used in the Cloud SQL connection.

The sample can continue to work as-is for testing / non-production use by defining the password in an environment variable.

An upcoming tutorial in the [Cloud Architecture Center](https://cloud.google.com/architecture) will show a complete example of creating a Secret Manager secret, deploying Cloud SQL, and then running this sample app with the proposed changes to tie it all together.

### DIFF
--- a/cloud-sql/mysql/mysql/README.md
+++ b/cloud-sql/mysql/mysql/README.md
@@ -26,8 +26,11 @@ export DB_USER='my-db-user'
 export DB_PASS='my-db-pass'
 export DB_NAME='my_db'
 ```
-Note: Saving credentials in environment variables is convenient, but not secure - consider a more
-secure solution such as [Cloud KMS](https://cloud.google.com/kms/) or [Secret Manager](https://cloud.google.com/secret-manager/) to help keep secrets safe.
+Note: Defining credentials in environment variables is convenient, but not secure. For a more secure solution, use
+[Secret Manager](https://cloud.google.com/secret-manager/) to help keep secrets safe. You can then define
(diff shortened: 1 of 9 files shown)


shubha-rajan (commented):
Hi @iainfoulds, thanks for the PR. The Secret Manager integration looks great. I think checking to see if Secret Manager is set up, and falling back to the DB_PASS environment variable if it's not set is a good approach, since it demonstrates best practices, but also still allows users to run this application without having to set up Secret Manager, if they just want something they can get up and running fast. 

Would it be possible to similarly modify the Postgres and SQL Server samples in this PR? Our team prefers to keep the samples aligned with eachother.

iainfoulds:
Thanks, @shubha-rajan. I'll look at updating PostgreSQL and SQL Server samples, too.

iainfoulds:
Hi, @shubha-rajan. Examples have been added for PostgreSQL and SQL Server, and I've tested both of these samples work with and without the use of Secret Manager. Please let me know if there's anything else to update.

derek64 (approved):

tmatsuo:
Justin's merge commit makes cla bot unhappy. I'll omit the commit and rebase it.

tmatsuo (approved):

PR #2401, opened 2021-11-02
feat(functions/v2): add GCE + CAL sample

ace-n:
Notes:
1. This sample has an E2E test _only_. This is currently disabled, as I don't think the testing project is configured for V2. (We should enable it once GCF V2 is available within that project, **provided** we're OK with a non-GA platform being tested.)
2. This uses an alpha version of the GCE Client Library. We'll need to replace that once the library goes GA.

### DIFF
--- a/functions/v2/index.js
+++ b/functions/v2/index.js
@@ -72,3 +72,46 @@ exports.helloAuditLog = cloudevent => {
   console.log('Principal:', payload.authenticationInfo.principalEmail);
 };
 // [END functions_log_cloudevent]
+
+// [START functions_label_gce_instance]
+const compute = require('@google-cloud/compute');
+const instancesClient = new compute.InstancesClient();
+
+/**
+ * CloudEvent function that labels newly-created GCE instances with the entity
+ * (person or service account) that created them.
+ *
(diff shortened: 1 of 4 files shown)


functions/v2/index.js
```diff
@@ -72,3 +72,46 @@ exports.helloAuditLog = cloudevent => {
...
+  let creator = cloudevent.data.protoPayload.authenticationInfo.principalEmail;
+  const params = cloudevent.subject.split('/');
+
+  // Format the 'creator' parameter to match GCE label validation requirements
+  creator = creator.toLowerCase().replace(/\W/g, '_');
+
+  // Get the newly-created VM instance's label fingerprint
+  // This is required by the Compute Engine API to prevent duplicate labels
+  const getInstanceRequest = {
+    project: params[2],
```
JustinBeckwith, line 98: Reaching into an array like this to snag params is a little hard to understand.  Maybe add a comment that shows the expected shape of the subject, or what you expect to have in the array?

JustinBeckwith (approved):

functions/v2/index.js
ace-n: Added an example `subject` value.

PR #2402, opened 2021-11-02
chore(functions): address comments on CAL samples

ace-n:


functions/v2/index.js
```diff
@@ -68,8 +69,14 @@ exports.helloAuditLog = cloudevent => {
...
   // Print out details from the Cloud Audit Logging entry
-  const payload = cloudevent.data.protoPayload;
-  console.log('Principal:', payload.authenticationInfo.principalEmail);
+  const metadata =
+    cloudevent.data &&
+    cloudevent.data.protoPayload &&
+    cloudevent.data.protoPayload.requestMetadata;
+  if (metadata) {
+    console.log('Caller IP:', metadata.callerIp);
+    console.log('User Agent:', metadata.callerSuppliedUserAgent);
```
averikitsch, line 78: nit: these might not be interesting either. Do we have an interesting use case? What is usually in the protoPayload?

functions/v2/test/index.test.js
```diff
@@ -18,6 +18,9 @@ const {exec} = require('child_process');
 const waitPort = require('wait-port');
 
 const startFF = async (target, signature, port) => {
+  console.log(
```
averikitsch, line 21: nit: console.log

averikitsch (approved):
just nits

functions/v2/test/index.test.js
ace-n: Good catch, thanks!

functions/v2/index.js
ace-n: Resolved over chat - we decided to show some other properties, too.

JustinBeckwith (approved):

PR #2428, opened 2021-11-11
refactor(functions/v2): update Pub/Sub sample

matthewrobertson:
Update the GCFv2 Pub/Sub sample to use the new declarative registration
API that was added in v2.0.0 of the Functions Framework.

### DIFF
--- a/functions/v2/index.js
+++ b/functions/v2/index.js
@@ -14,16 +14,13 @@
 
 'use strict';
 
+const functions = require('@google-cloud/functions-framework');
+
 // [START functions_cloudevent_pubsub]
-/**
- * CloudEvent function to be triggered by Pub/Sub.
- * This function is exported by index.js, and executed when
- * the trigger topic receives a message.
- *
- * @param {object} cloudevent A CloudEvent containing the Pub/Sub message.
- * @param {object} cloudevent.data.message The Pub/Sub message itself.
- */
-exports.helloPubSub = cloudevent => {
+// Register a CloudEvent callback with the Functions Framework that will
+// be executed when the Pub/Sub trigger topic receives a message.
+functions.cloudEvent('helloPubSub', cloudevent => {
+  // The Pub/Sub message is passed as the CloudEvent's data payload.
   const base64name = cloudevent.data.message.data;
 
   const name = base64name
@@ -31,7 +28,7 @@ exports.helloPubSub = cloudevent => {
     : 'World';
 
   console.log(`Hello, ${name}!`);
-};
+});
 // [END functions_cloudevent_pubsub]
 
 // [START functions_cloudevent_storage]
--- a/functions/v2/package.json
+++ b/functions/v2/package.json
@@ -19,7 +19,7 @@
     "@google-cloud/compute": "^3.0.0-alpha.4"
   },
   "devDependencies": {
-    "@google-cloud/functions-framework": "^1.1.1",
+    "@google-cloud/functions-framework": "^2.0.0",
     "gaxios": "^4.3.0",
     "mocha": "^9.1.3",
     "p-retry": "^4.6.1",


functions/v2/index.js
```diff
@@ -14,24 +14,21 @@
...
- * This function is exported by index.js, and executed when
- * the trigger topic receives a message.
- *
- * @param {object} cloudevent A CloudEvent containing the Pub/Sub message.
- * @param {object} cloudevent.data.message The Pub/Sub message itself.
- */
-exports.helloPubSub = cloudevent => {
+// Register a CloudEvent callback with the Functions Framework that will
+// be executed when the Pub/Sub trigger topic receives a message.
+functions.cloudEvent('helloPubSub', cloudevent => {
```
ace-n, line 22: This will break our tests. How about configuring `functions.cloudEvent` to return the second argument, and _maybe_ (if it's possible) dropping the first?

Then we could do this:
```
exports.helloPubSub = functions.cloudEvent(cloudevent => {...})
```

ace-n (changes requested):

functions/v2/index.js
matthewrobertson: The API is already finalized and released so we can't change it now. 

Which tests will break? They all passed when I ran `npm run test`

ace-n: Oh, now I realize why: these tests use the `functions-framework` itself. What this *will* break are unit tests (which typically import the `export`ed function within Node itself).

Does `functions.cloudEvent` currently return anything? If not, how hard would it be to have it return the function itself?

matthewrobertson: For additional context: the API in which the customer passes the function name to framework was a deliberate design decision. It removes the need for them to tell us the name of the function when they run the app (eg `functions-framework --target=helloPubSub`). This unblocks our ability to run multiple functions on the same port which is a [long standing feature request](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/23)

matthewrobertson: right now `functions.cloudEvent` returns `void`: 

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/function_registry.ts#L70-L75

It makes sense to have it return something eventually, but I was trying to punt on that while we work out the design for some more features we are planning to add to the FFs. In my mind we will want to return the user's function wrapped in some object that exposes a configuration API:

```
functions
  .cloudEvent('foo', (ce) => ce.data)
  .afterSend(() => doSomeExpensiveCleanup())
```

but that is a bit speculative at the moment.

ace-n: Another possible pattern (though it feels less clean to me):

```
// Declare function
const helloPubSub = (cloudevent) => { ... }

// Expose function
functions.cloudEvent('helloPubSub', helloPubSub)
exports.helloPubSub = helloPubSub
```

matthewrobertson: @ace-n I filed a bug in the functions framework to document this / explore API changes to make it easier: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/387

As it stands, this sample is an accurate representation of the existing API. Can we land it and work out the unit testing story in the FF repo?

ace-n: How do you feel about the pattern I suggested?

ace-n: I added some comments to the example above; if you can duplicate those in this PR that'd be 👍 .

matthewrobertson: I am a bit confused, do you want me to implement this sample in that way? There are no unit tests for this sample so it seems unnecessary.

ace-n: Yes - that's my preference, for consistency reasons.

matthewrobertson:
This PR is superseded by https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2457

PR #2430, opened 2021-11-12
Add Cloud Run sample

averikitsch:
I will add a bug to add E2E tests


run/jobs/index.js
```diff
@@ -0,0 +1,65 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// Retrieve Job-defined env vars
+const {TASK_NUM, ATTEMPT_NUM} = process.env;
```
JustinBeckwith, line 18: Could we use sensible defaults here if not defined in env vars?

run/jobs/index.js
```diff
@@ -0,0 +1,65 @@
...
+  // Simulate work
+  if (SLEEP_MS) {
+    await sleep(SLEEP_MS);
+  }
+  // Simulate errors
+  if (FAIL_RATE) {
+    try {
+      randomFailure(FAIL_RATE);
+    } catch (err) {
+      throw new Error(`Task #${TASK_NUM}, Attempt #${ATTEMPT_NUM} failed.`);
```
JustinBeckwith, line 34: From a pattern perspective, I like to avoid throwing a new error from inside of a catch.  This obfuscates the original callstack, and makes it very hard to figure out what actually went wrong.  Instead, try prepending your error message to the existing error, and rethrowing:

```js
err.message = `Task #${TASK_NUM}, Attempt #${ATTEMPT_NUM} failed.\n\n${err.message}`;
throw err;
```

run/jobs/test/index.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+  it('should run successfully', async () => {
+    const stdout = exec('node index');
+    assert.match(stdout, /Starting Task #/);
+    assert.match(stdout, /Completed Task #/);
+  });
+
+  it('should fail with high fail rate', async () => {
+    process.env.FAIL_RATE = 0.9999;
+    try {
```
JustinBeckwith, line 32: This is a perfect place to use `assert.throws`.  Instead of managing the try/catch like this, you explicitly check that the given function throws, and you can use a regex to test the error message failure:

https://nodejs.org/api/assert.html#assertthrowsfn-error-message

run/jobs/index.js
```diff
@@ -0,0 +1,65 @@
...
+  const randomFailure = Math.random();
+  if (randomFailure < rate) {
+    throw new Error('Task failed.');
+  }
+};
+
+// Start script
+main().catch(err => {
+  console.error(err);
+  process.exit(1); // Trigger Task retry by exiting the process
```
ace-n, line 64:
```suggestion
  process.exit(1); // Request Cloud Run Jobs to retry by exiting the process

ace-n (approved):
LGTM once @JustinBeckwith approves

JustinBeckwith (approved):

PR #2436, opened 2021-11-18
Switch JSON middleware for UrlEncoded

philBrown:
The form in `views/form.html` submits normally via the browser (`application/x-www-form-urlencoded`) so it needs the `express.urlencoded()` middleware instead of `express.json()`


bcoe:
@jmdobry does this sample ring a bell, looks like you worked on it eons ago.

engelke (approved):

PR #2439, opened 2021-11-22
build(event-arc): migrate to GitHub Actions

JustinBeckwith:


engelke:
Core team member should see if this is a configuration issue.

JustinBeckwith:
Current status of this one - I did a bunch of the work to get it converted, but this particular set of tests rely on the ability to use `gcloud auth print-identity-token` to make a request to a Cloud Run service.  With the transition to OIDC for authentication in this PR, that command has stopped working and I don't know how to fix it :) This requires following up with @sethvargo and/or someone that really understands the relationship between that gcloud command and OIDC with GCP.

.github/workflows/eventarc-audit-storage.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - cron:  '0 2 * * *'
+jobs:
+  test:
+    if: ${{ github.event.action != 'labeled' || github.event.label.name == 'actions:force-run' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+    - uses: 'google-github-actions/auth@v0.3.1'
```
sethvargo, line 23: We recommend pinning actions to `@vx` (`@v0` in this case) unless you have security concerns, in which case you should pin to a specific SHA. This applies to all workflows and google-github-actions, but I'll only leave this comment once.

.github/workflows/eventarc-audit-storage.yaml
```diff
@@ -0,0 +1,45 @@
...
+    paths:
+    - 'eventarc/audit-storage/**'
+  pull_request:
+    paths:
+    - 'eventarc/audit-storage/**'
+  pull_request_target:
+    types: [labeled]
+  schedule:
+  - cron:  '0 2 * * *'
+jobs:
```
sethvargo, line 15: I generally recommend you add the following to prevent queue backup:

```yaml
concurrency:
  group: '${{github.workflow}}-${{ github.head_ref || github.ref }}'
  cancel-in-progress: true
```

This applies to all workflows, but I'll only leave the comment once here.

.github/workflows/eventarc-audit-storage.yaml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+    - uses: 'google-github-actions/auth@v0.3.1'
+      with:
+        workload_identity_provider: 'projects/1046198160504/locations/global/workloadIdentityPools/github-actions-pool/providers/github-actions-provider'
+        service_account: 'kokoro-system-test@long-door-651.iam.gserviceaccount.com'
+        create_credentials_file: 'true'
```
sethvargo, line 27: This is the default value and can be omitted

.github/workflows/eventarc-audit-storage.yaml
```diff
@@ -0,0 +1,45 @@
...
+        access_token_lifetime: 600s
```
sethvargo, line 28: The default behavior does not create an access token (you need to use `token_format: access_token` to generate an access token or `token_format: id_token` for OIDC tokens.

eventarc/audit-storage/test/runner.sh
```diff
@@ -39,8 +46,6 @@ function cleanup {
 }
 trap cleanup EXIT
 
-# TODO: Perform authentication inside the test.
-export ID_TOKEN=$(gcloud auth print-identity-token)
```
sethvargo, line 43: You can generate an ID token from `auth` with:

```yaml
steps: 
- uses: 'google-github-actions/auth@v0'
  with:
    # ... existing values
    token_format: 'id_token'
    id_token_audience: 'TODO'
    id_token_include_email: true
```

For Cloud Run, the audience must be the _full_ URL of the Cloud Run service (e.g. `https://my-cloud-run-service.run.app/`).

eventarc/audit-storage/test/system.test.js
```diff
@@ -39,9 +24,24 @@ describe('End-to-End Tests', () => {
   }
 
   it('post(/) without request parameters is a bad request', async () => {
-    const response = await request('post', '/', BASE_URL);
+    const account = 'kokoro-system-test@long-door-651.iam.gserviceaccount.com';
+    const res = await auth.request({
+      url: `https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/${account}:generateIdToken`,
+      params: {
+        audience: account,
```
sethvargo, line 31: I don't think this is the right `aud` value?

PR #2457, opened 2021-12-07
feat(functions/v2): update to the new declarative signature APIs

matthewrobertson:
This commit updates the functions/v2 samples to use the new Functions Framework APIs for declaratively configuring a function's signature type. It also update the unit tests to use the new testing APIs added in v2.1.0 of the functions framework.


functions/v2/index.js
```diff
@@ -14,34 +14,27 @@
...
- * This function is exported by index.js, and executed when
- * the trigger topic receives a message.
- *
- * @param {object} cloudevent A CloudEvent containing the Pub/Sub message.
- * @param {object} cloudevent.data.message The Pub/Sub message itself.
- */
-exports.helloPubSub = cloudevent => {
+// Register a CloudEvent callback with the Functions Framework that will
+// be executed when the Pub/Sub trigger topic receives a message.
+functions.cloudEvent('helloPubSub', cloudevent => {
```
ace-n, line 22: Same comment as the last PR - instead of doing:
```
functions.cloudEvent('helloPubSub', ...)
```
can we do:
```
const helloPubSub = ...

// Expose the function
functions.cloudEvent('helloPubSub', helloPubSub)
exports.helloPubSub = helloPubSub
```

functions/v2/package.json
```diff
@@ -12,19 +12,18 @@
     "node": ">=12.0.0"
   },
   "scripts": {
-    "test": "mocha --timeout 30s test/index.test.js",
+    "test": "mocha test/index.test.js",
```
ace-n, line 15: Is there a reason this was deleted?

(Did everything pass consistently without it?)

functions/v2/test/index.test.js
```diff
@@ -13,94 +13,46 @@
...
-const startFF = async (target, signature, port) => {
-  const ff = exec(
-    `npx functions-framework --target=${target} --signature-type=${signature} --port=${port}`
-  );
-  await waitPort({host: 'localhost', port});
-  return ff;
-};
+beforeEach(() => {
+  // require the module that includes the functions we are testing
+  require('../index');
```
ace-n, line 22: I feel like this needs more discussion.

If we shouldn't be running the FF during the testing process, that's fine - but we **do** need to clearly articulate (to customers) what the best practices are.

functions/v2/test/index.test.js
```diff
@@ -13,94 +13,46 @@
 // limitations under the License.
 
 const assert = require('assert');
-const {request} = require('gaxios');
-const {exec} = require('child_process');
-const waitPort = require('wait-port');
+const {getTestServer} = require('@google-cloud/functions-framework/testing');
+const sinon = require('sinon');
+const supertest = require('supertest');
```
ace-n, line 18: I believe we moved away from `supertest` in the past, but don't quite remember why.

(Let's hold this thought until we figure out whether we want to use the FF as part of the testing architecture.)

cc @JustinBeckwith

ace-n (changes requested):

functions/v2/index.js
matthewrobertson: I don't think this is what we want customers to use the API. There get better implicit typing and IDE autocompleting by passing a callback function directly to the `cloudEvent` function.

This in now unit testable using the helpers we released in v2.1.0.

functions/v2/package.json
matthewrobertson: This is no longer necessary. The tests will be much faster, and less flaky because they are no longer forking child processes and waiting for ports.

functions/v2/test/index.test.js
matthewrobertson: It is articulated here: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/docs/testing-functions.md

functions/v2/test/index.test.js
matthewrobertson: supertest seems to be used [throughout this repo](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/search?q=supertest). 

I am happy to revert the unit testing changes (all existing tests passed as written), although I strongly relying on this mechanism rather than forking processes for the reasons mentioned above (faster, less flaky).

ace-n: Ah - I didn't realize the _FF itself_ was providing a test server.

As long as we test the _integration_ with the FFs themselves, I'm fine with this.

We should file a follow-up issue to apply this change in other locations, too (assuming it's supported for Gen 1's "background" functions; if not, we should make sure to do it as part of the migration to Gen 2).

functions/v2/test/index.test.js
ace-n: So I chatted with some of the other Node folks on my team, and we thought that an `exports.<method>` model (in the tested files) would be more idiomatic than `require`-ing a tested file directly.

Is there any reason we _shouldn't_ move to that model? As a reminder, it would look like this:

_index.js_
```
const someFunction = (cloudevent) => { ... }

// Expose function to external sources
functions.cloudEvent('someFunction', someFunction);
exports.someFunction = someFunction
```

functions/v2/test/index.test.js
matthewrobertson: These helpers only work with the new declarative APIs right now. We might be able to find a way to make them work with background functions though. I will file an issue in the Functions Framework repo.

matthewrobertson: Filed issue here: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/404

functions/v2/test/index.test.js
ace-n: @matthewrobertson ah, that is new information!

Can/should we have the FF itself do an `export` under the hood?

If not, can/should we have the FF constructors return the function itself? (That way, we could do `exports.someFunction = functions.cloudevent('someFunction', ...)`)

_(Hopefully we're not going in circles here - I'd be more than happy to schedule a VC sync + create an agenda if you'd like me to.)_

matthewrobertson: > Can/should we have the FF itself do an export under the hood, or (as you may have suggested elsewhere) some sort of functionsFramework.getFunction('someFunction') call?

Yes this exists now as a helper in the `/testing` module:

```
const {getFunction} = require('@google-cloud/functions-framework/testing');

describe('myFunc', () => {
  it('works, () => {
    const myFunc = getFunction('myFunc');
    // test myFunc...
  });
});
```

> I'd be more than happy to schedule a VC sync

sure happy to chat

functions/v2/test/index.test.js
```diff
@@ -110,13 +62,18 @@ describe('functions_cloudevent_storage', () => {
...
+    const server = getTestServer('helloGCS');
+    await supertest(server)
+      .post('/')
+      .send(event)
+      .set('Content-Type', 'application/json')
+      .expect(204);
+
+    const output = console.log
+      .getCalls()
+      .map(x => x.args[0])
```
ace-n, line 74: Sinon has assertions for particular calls. Can we use something like [this](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/679979ff03f1793626a12b0aacd6f15bbf4ba3fd/functions/firebase/test/index.test.js#L67-L70)?

ace-n (approved):
LGTM once nits are addressed.

functions/v2/test/index.test.js
matthewrobertson: In this case, the module still needs to be required. The helper won't be able to find the function it is has never been loaded.

functions/v2/test/index.test.js
matthewrobertson: Done

functions/v2/test/index.test.js
```diff
@@ -13,94 +13,46 @@
 // limitations under the License.
 
 const assert = require('assert');
-const {request} = require('gaxios');
-const {exec} = require('child_process');
-const waitPort = require('wait-port');
+const {getTestServer} = require('@google-cloud/functions-framework/testing');
```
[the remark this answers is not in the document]
matthewrobertson: done

functions/v2/test/index.test.js
ace-n: Ah - so the helper does not handle the `require` internally.

Ideally we would add that, but I don't think it's worth blocking this PR. 🙂

ace-n:
@matthewrobertson I think we need one of the dedicated Node folks (@GoogleCloudPlatform/nodejs-docs-samples-owners) to sign off on this before merging.

I'll ping them over chat internally.

sofisl (approved):

PR #2476, opened 2022-01-13
chore(functions/helloworld): fix incorrectly documented error behavior

### LINKED ISSUE #2317: `console.error(new Error(''))` won't be sent to Cloud Error Reporting automatically
It seems [this line](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/functions/helloworld/index.js#L98) is incorrect.

I tested in a deployed cloud function: if I do `console.error(new Error('hello world'))`, it won't show up in console -> Error Reporting. But if I use the `@google-cloud/error-reporting` library to report errors, it'll show.

If I'm right, then the [doc](https://cloud.google.com/functions/docs/monitoring/error-reporting) is mistaken?

ace-n:
Fixes #2317

### DIFF
--- a/functions/helloworld/index.js
+++ b/functions/helloworld/index.js
@@ -98,8 +98,6 @@ exports.helloGCS = (file, context) => {
 exports.helloError = (event, context, callback) => {
   // [START functions_helloworld_error]
   // These WILL be reported to Stackdriver Error Reporting
-  console.error(new Error('I failed you'));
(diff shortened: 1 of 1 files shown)


bcoe (commented):
Would it be worth adding an example of writing structured logs?

https://cloud.google.com/logging/docs/samples/logging-write-log-entry

JustinBeckwith (approved):

PR #2497, opened 2022-02-01
feat(functions/helloworld): Use declarative syntax for http sample

matthewrobertson:
This brings it into line with the rest of the v2 samples.

### DIFF
--- a/functions/helloworld/index.js
+++ b/functions/helloworld/index.js
@@ -121,7 +121,7 @@ exports.helloError3 = (event, context, callback) => {
 };
 
 // HTTP Cloud Function that returns an error.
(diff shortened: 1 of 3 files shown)


ace-n (changes requested):
This PR only updates `helloGET`, while the [Python PR](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/7214) updated **all** the samples in that file.

For consistency, {could, should} we do the latter here as well?

matthewrobertson:
@ace-n most of the samples in this file are background functions which are not supported by the new declarative API. I updated all the http functions.

ace-n (approved):

PR #2515, opened 2022-02-02
build: move functions/spanner to GH Actions

sofisl:


ace-n (approved):
LGTM if the relevant tests pass.

(I assume you'll also handle deleting the relevant **internal** Kokoro configs?)

chingor13 (approved):

PR #2531, opened 2022-02-08
fix: revert #2383 for composer code

leahecole:
I missed #2383 which breaks Composer code that runs in Dataflow. We have to use `var` because of Dataflow limitations Partial fix for b/209830375


leahecole:
JKJK it's good I think I had an old version of the primary branch on my local machine 🤦🏻

sofisl (approved):

PR #2539, opened 2022-02-11
build: remove unused opencensus config w/ no corresponding Kokoro config, add opencensus test in GH Actions

sofisl:


.github/workflows/opencensus.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - 'opencensus/**'
+  pull_request:
+    paths:
+    - 'opencensus/**'
+  pull_request_target:
+    types: [labeled]
+  schedule:
+  - cron:  '0 2 * * *'
+jobs:
+  test:
```
tmatsuo, line 16: maybe use `opencensus` instead of `test` as the job name. Then make `opencensus` a mandatory test.

tmatsuo (approved):

PR #2566, opened 2022-02-28
fix(functions/helloworld): include FF import

ace-n:
Fixes b/221561399


functions/helloworld/index.js
```diff
@@ -32,6 +31,7 @@ functions.http('helloGET', (req, res) => {
 // [END functions_helloworld_get]
```
iennae, line 31: it looks like the region tag for functions_helloworld_get is just the functions-framework declaration. is that intentional?

ace-n: It also includes a sample at (post-change) lines 28-30.

(post-change) lines 22 and 24 are different region tags.

functions/helloworld/index.js
```diff
@@ -19,10 +19,9 @@
 // [START functions_helloworld_http]
```
averikitsch, line 19: Do you think something like this would be more clear to future contributors? 
```suggestion
// [START functions_helloworld_http]
// [START functions_helloworld_get]
const functions = require('@google-cloud/functions-framework');
// [END functions_helloworld_get]
const escapeHtml = require('escape-html');
// [END functions_helloworld_http]
```

averikitsch (approved):

functions/helloworld/index.js
ace-n: Thanks! Made a similar change. 👍

sofisl (approved):

PR #2577, opened 2022-03-22
healthcare API: do not use Healthcare client libraries for FHIR search

noerog:
Part of fix for b/221324824. Our client libraries do not allow for FHIR GET/POST search requests to be set in the HTTP request path, which is required by the FHIR specification. Due to a bug, you can send search requests by specifying a resource type in the request body, but this is an anti-pattern. Until we fix the issue with the client library, we need to use the gaxios library to explicitly make GET/POST requests and not use the client library.


sofisl (approved):

noerog:
@sofisl Is there a way to skip the Cloud Run integration tests that are failing? Or do they need to pass before this PR can be merged?

PR #2580, opened 2022-03-24
build(deps):Update google-github-actions, actions/checkout, and setup-node for appengine/typescript

sofisl:


engelke (approved):
LGTM

ace-n (approved):

.github/workflows/appengine-typescript.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       pull-requests: 'write'
       id-token: 'write'
     steps:
-    - uses: 'google-github-actions/auth@v0.3.1'
+    - uses: 'google-github-actions/auth@v0.6.0'
```
tmatsuo, line 24: I think the recent version of `google-github-actions/auth` asks you to first checkout the source code then use the `auth`.

tmatsuo (changes requested):

.github/workflows/appengine-typescript.yaml
```diff
@@ -21,17 +21,17 @@ jobs:
       pull-requests: 'write'
       id-token: 'write'
     steps:
-    - uses: 'google-github-actions/auth@v0.3.1'
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{github.event.pull_request.head.ref}}
```
tmatsuo, line 26: Can you fetch the `github.event.pull_request.head.sha` instead of `ref`?

This ensures it fetches the source code at the sha when someone added the label, I think.

tmatsuo (changes requested):

.github/workflows/appengine-typescript.yaml
sofisl: Since this is a template, I created an issue to do this for all workflows at another time: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/2589

PR #2581, opened 2022-03-24
build(deps): build(deps): Update google-github-actions, actions/checkout, and setup-node for cloud-tasks/app

sofisl:


cloud-tasks/app/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "start": "node index.js",
-    "test": "mocha --timeout 10000"
+    "test": "mocha test/*.test.js --exit"
```
bcoe, line 11: What problem does `--exit` solve, also wondering if we should keep the longer timeout?

bcoe (changes requested):

bcoe (approved):

PR #2583, opened 2022-03-25
samples(healthcare v1): initial consent store CRUD method samples with new subdirectory

noerog:
New directory and initial samples for Healthcare API consent modality. Part of b/224797364.


noerog:
@GoogleCloudPlatform/nodejs-samples-reviewers Ping

sofisl:
Looks like tests are not running for this directory. Could I trouble you to set up a workflow, just like these here? https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/main/.github/workflows. Then we can make sure that tests are passing for this new directory.

noerog:
> Looks like tests are not running for this directory. Could I trouble you to set up a workflow, just like these here? https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/main/.github/workflows. Then we can make sure that tests are passing for this new directory.

Done.

noerog:
It looks like there are some permissions errors. We had the same issue when adding these samples with a new directory in python-docs-samples. Would it be possible to verify that the service account used to run tests has the correct permissions?

PR #2585, opened 2022-03-25
feat(cloudsql/postgres): update to V2 sample.

jsimonweb:


enocom (approved):

cloud-sql/postgres/knex/app.standard.yaml
```diff
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: nodejs10
+runtime: nodejs17
```
ace-n, line 14: Even-numbered Node releases are used as long-term support (LTS) candidates.

Perhaps we should use `node16` here?

cloud-sql/postgres/knex/index.js
```diff
@@ -311,7 +240,7 @@ app.get('/', async (req, res) => {
   }
 });
 
-app.post('/', async (req, res) => {
+const httpPost = app.post('*', async (req, res) => {
```
ace-n, line 243: `httpPost` looks like it points to an Express route, not the route handler function itself.

We probably need something like this:
```
const httpPost = async (req, res) => {
   ...
}

app.post('*', httpPost)
```

ace-n (approved):
LGTM once:
a) comments addressed, **and**
b) tests pass

cloud-sql/postgres/knex/connect-unix.js
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+// [START cloud_sql_postgres_knex_connect_unix]
+const Knex = require('knex');
+
+const createUnixSocketPool = async config => {
+  const dbSocketPath = process.env.INSTANCE_UNIX_SOCKET;
```
enocom, line 20: Let's add the secret manager disclaimer in these samples. See https://github.com/GoogleCloudPlatform/golang-samples/blob/main/cloudsql/postgres/database-sql/connect_tcp.go#L42-L45.

sofisl (approved):

PR #2595, opened 2022-04-05
feat: adding http structured logging sample + test

pattishin:
Structured logging http for GCFv2

**To run sample:**

1) Execute below commands in either local terminal or cloud shell.
```
npm i
gcloud beta functions deploy nodejs-http-function --gen2 --runtime nodejs16 --trigger-http --entry-point structured_logging --source . --project YOUR_PROJ --allow-unauthenticated
```

2) After successful deployment, either curl or open the generated http function url in a new browser.

3) Open up logs and verify `Hello, world!` has logged and that you are able to see something similar to below.

Screen Shot 2022-04-04 at 5 22 27 PM

4) Run unit test via `npm run test`


pattishin:
Checking out the linting/region tag errors!

functions/v2/http-logging/test/index.test.js
```diff
@@ -0,0 +1,55 @@
...
+'use strict';
+
+const assert = require('assert');
+const sinon = require('sinon');
+const { getFunction } = require('@google-cloud/functions-framework/testing');
+
+describe('structured logging: functions http', () => {
+	let req, res, spy;
+
+	before(async () => {
```
grayside, line 24: Since the goal of the sample is to take advantage of the new stdout log delivery of the library, it would be good to make this a more comprehensive test and double-check we're using that as expected by testing captured stdout instead of intercepting the library call.

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+'use strict';
+
+// [START functions_structured_logging]
+const cloudLogging = require('@google-cloud/logging');
+const functions = require('@google-cloud/functions-framework');
+
+functions.http('structured_logging', async (req, res) => {
+  const logging = new cloudLogging.Logging();
```
averikitsch, line 22: add comment about client creation

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const log = logging.logSync('Structured_logging project');
```
averikitsch, line 23: Add comment like "Create a LogSync transport, defaulting to `process.stdout`"

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const text = 'Hello, world!';
+
+  // The metadata associated with the entry
+  // See: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity
+  const metadata = {
+    severity: 'NOTICE',
+    component: 'arbitrary-property',
+    httpRequest: req,
+  };
```
averikitsch, line 32: We should add some commentary around this. Let's brainstorm.

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,39 @@
...
+  // Prepares a log entry
+  const entry = log.entry(metadata, text);
+  log.write(entry);
+
+  res.status(200).send(entry);
```
averikitsch, line 37: Instead of sending the entry as the response, let's just default something like "A log message was written"

functions/v2/http-logging/test/index.test.js
averikitsch: yes, let's remove the mocking and just test the stdout. A similar example can be found: https://github.com/googleapis/nodejs-dlp/blob/main/samples/system-test/quickstart.test.js

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START functions_structured_logging]
+const cloudLogging = require('@google-cloud/logging');
+const functions = require('@google-cloud/functions-framework');
+
+functions.http('structured_logging', async (req, res) => {
```
averikitsch, line 21: In the CLoud Logging screenshot you provided you can see the trace field has `...{{projectId}}..` which means the project id isn't set. You will need to add `await logging.setProjectId()`

averikitsch (changes requested):

averikitsch:
I need to investigate what is needed for the tests, so we may have to add another config file

averikitsch:
The region tag test looks new and broken @sofisl

functions/v2/http-logging/test/index.test.js
```diff
@@ -0,0 +1,55 @@
...
+			component: 'arbitrary-property',
+			httpRequest: req,
+		};
+
+		const { metadata, data } = spy.getCall(0).args[0];
+
+		assert.equal(metadata['severity'], expectedMetadata['severity']);
+		assert.equal(metadata['component'], expectedMetadata['component']);
+		assert.equal(metadata['httpRequest'], expectedMetadata['httpRequest']);
+		assert.equal(data, "Hello, world!");
```
averikitsch, line 53: We should also ensure that project id is found in the log

functions/v2/http-logging/index.js
pattishin: Sure! Thoughts on the update in line 31? If not, I'll send you a chat about it :)

pattishin:
Checking on linting, but PR updated! Let me know if I left anything out!

.gitignore
```diff
@@ -10,4 +10,5 @@ logs/
 .nyc_output
```
averikitsch, line 10: this file lives at the root level. Please remove

functions/v2/http-logging/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "0.0.1",
+  "description": "Simple structured logging http example",
+  "license": "Apache-2.0",
+  "author": "Google LLC",
+  "main": "index.js",
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "scripts": {
+    "test": "mocha test/*.test.js"
```
averikitsch, line 12: We should provide a "start"

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,43 @@
...
+const functions = require('@google-cloud/functions-framework');
+const pkg = require('./package.json');
+
+functions.http('structuredLogging', async (req, res) => {
+  // Initialize the logging client
+  const logging = new Logging();
+  // Create a LogSync transport, defaulting to process.stdout
+  const log = logging.logSync(pkg.name);
+  // Required to capture your project id
+  await logging.setProjectId('YOUR_PROJ');
```
averikitsch, line 28: 'YOUR_PROJ' is not needed. Let's also move this after the client is initialized

functions/v2/http-logging/test/index.test.js
```diff
@@ -0,0 +1,69 @@
...
+const { getFunction } = require('@google-cloud/functions-framework/testing');
+
+describe('structured logging: functions http', () => {
+	let mockReq, mockRes;
+
+	before(async () => {
+		// Importing our target file
+		await import("../index.js");
+		// Adding a spy to the write function in stdout
+		sinon.spy(process.stdout, 'write');
```
averikitsch, line 29: Ahha! I like this solution :)

functions/v2/http-logging/test/index.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+		mockReq = {
+			requestMethod: 'GET',
+			requestUrl: '/',
+			userAgent: 'foobar'
```
averikitsch, line 34: We also want to test the trace, so add this to the request: `'X-Cloud-Trace-Context': '105445aa7843bc8bf206b12000100000/1;o=1'`
Then test for `"logging.googleapis.com/trace":"projects/${GOOGLE_CLOUD_PROJECT}/traces/105445aa7843bc8bf206b12000100000"`

functions/v2/http-logging/package.json
averikitsch: `npx @google-cloud/functions-framework --target=structuredLogging`

functions/v2/http-logging/index.js
averikitsch: "Create a structured log entry with severity, additional component fields, and HTTP request. Appending the httpRequest enables log correlation."

averikitsch (approved):

averikitsch:
Failed tests can be ignored

functions/v2/http-logging/index.js
```diff
@@ -0,0 +1,43 @@
...
+const functions = require('@google-cloud/functions-framework');
+const pkg = require('./package.json');
+
+functions.http('structuredLogging', async (req, res) => {
+  // Initialize the logging client
+  const logging = new Logging();
+  // Required to capture your project id
+  await logging.setProjectId();
+  // Create a LogSync transport, defaulting to process.stdout
+  const log = logging.logSync(pkg.name);
```
ace-n, line 28: Nit: use a variable, instead of package name?

ace-n (approved):

functions/v2/http-logging/index.js
ace-n: Edit: disregard, see [Averi's comment here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2596#discussion_r851429858).

averikitsch:
@sofisl PTAL

averikitsch:
@sofisl PTAL

averikitsch:
@GoogleCloudPlatform/nodejs-samples-reviewers  please review

sofisl (approved):

PR #2606, opened 2022-04-15
fix(functions/firebase): update RTDB event structure

ace-n:
Partial fix for b/226142619

### DIFF
--- a/functions/firebase/index.js
+++ b/functions/firebase/index.js
@@ -17,12 +17,13 @@
  * Triggered by a change to a Firebase RTDB reference.
  *
  * @param {!Object} event The Cloud Functions event.
+ * @param {!Object} context The Cloud Functions event context.
  */
-exports.helloRTDB = event => {
-  const triggerResource = event.resource;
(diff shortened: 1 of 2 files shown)


functions/firebase/index.js
```diff
@@ -17,12 +17,13 @@
  * Triggered by a change to a Firebase RTDB reference.
  *
  * @param {!Object} event The Cloud Functions event.
+ * @param {!Object} context The Cloud Functions event context.
  */
-exports.helloRTDB = event => {
-  const triggerResource = event.resource;
+exports.helloRTDB = (event, context) => {
```
averikitsch, line 22: nit: Link to find this info like https://cloud.google.com/functions/docs/writing/background#function_parameters

averikitsch (approved):

functions/firebase/index.js
ace-n: Called out that these are `Background Functions` - that should be precise enough for users to find that page.

feywind (approved):

PR #2645, opened 2022-05-11
feat(functions/helloPubSub): use new FF testing layer

saraford:
In this PR:

1. Updated existing test to use a name "Cecil" instead of "World". If the name parameter fails, the default is Hello World, so the test will always pass. 
2. I've added a unit test example to show the FF testing layer, hence changed the test folder to "tests" and renamed the existing test to integration tests. 

❓ I noticed that I can run `mocha tests/integration-tests.js` or `mocha tests/unit-tests.js` successfully individually. But if I try to follow the style in other samples by running `npm run test` where test is `mocha tests/*-tests.js --timeout=60000`, the second test that's run will fail.


functions/v2/helloPubSub/package.json
```diff
@@ -12,14 +12,13 @@
     "node": ">=12.0.0"
   },
   "scripts": {
-    "test": "mocha test/index.test.js"
+    "test": "mocha tests/*-tests.js --timeout=60000"
```
ace-n, line 15: Nit: I believe this repo's convention is to use a `test` directory (no `s`)

ace-n (approved):
LGTM once lint errors and comments addressed.

-------

RE testing - I wonder if the FF is running a server under the hood? Trying to run two servers on the same port might cause failures.

(We can chat more about this later this afternoon, too.)

functions/v2/helloPubSub/test/unit.test.js
```diff
@@ -0,0 +1,60 @@
+// Copyright 2018 Google LLC
```
averikitsch, line 1:
```suggestion
// Copyright 2022 Google LLC
```

averikitsch (approved):

ace-n:
Failures are unrelated; admin-merging.

sofisl (approved):

ace-n:
Failures are unrelated; admin-merging.

PR #2647, opened 2022-05-12
feat(functions/v2): use FF testing layer in audit log sample

saraford:


ace-n (approved):

functions/v2/helloAuditLog/.gcloudignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
```
ace-n, line 1: Should this be done in `functions/v2/` instead?

(I'm not sure if `gcloud` checks parent directories.)

ace-n: **Update:** @saraford confirmed that the `.gcloudignore` needs to the in the _function's_ root directory.

(We can ignore this comment.)

alexander-fenster (approved):

averikitsch (approved):

PR #2654, opened 2022-05-13
fix(run/websocket): Debug Websocket test errors

averikitsch:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/commit/05b5423242f273518e8600e26e2bd35036082959#r61378446


run/websockets/package.json
```diff
@@ -12,12 +12,12 @@
   "dependencies": {
     "@socket.io/redis-adapter": "^7.0.0",
     "express": "^4.17.1",
-    "redis": "^4.0.0",
+    "redis": "^3.1.2",
```
ace-n, line 15: Is there a way to avoid downgrading this?

averikitsch: There is a breaking change in the library. To avoid rewriting the entire sample this is needed.

ace-n: Got it. Is this something we plan to fix separately?

averikitsch: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/issues/2657

ace-n (approved):

sofisl (approved):

PR #2676, opened 2022-06-13
build: update sample to use var

### LINKED ISSUE #2469: UDF for Dataflow tutorial fails
When following [this tutorial](https://cloud.google.com/composer/docs/how-to/using/using-dataflow-template-operator#storage_create_bucket-nodejs) the [UDF code supplied in this repo](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/HEAD/composer/composer_transform_csv_to_json.js) is failing in Dataflow, as Dataflow doesn't recognize ```let```, ```const```. Works if each instance is changed to ```var```.

sofisl:
Fixes #2469

### DIFF
--- a/composer/composer_transform_csv_to_json.js
+++ b/composer/composer_transform_csv_to_json.js
@@ -39,7 +39,7 @@ module.exports = function main(
       }
     }
 
-    return (JSON.stringify(weatherInCity));
+    return JSON.stringify(weatherInCity);
   }
 
   // [END composer_transform_csv_to_json]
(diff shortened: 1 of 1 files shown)


composer/composer_transform_csv_to_json.js
```diff
@@ -39,8 +39,7 @@ module.exports = function main(
       }
     }
 
-    const jsonString = JSON.stringify(weatherInCity);
-    return jsonString;
+    return (JSON.stringify(weatherInCity));
```
leahecole, line 42:
```suggestion
    return JSON.stringify(weatherInCity);
```

sofisl:
I don't, but I assume this will be fine given that we're not saving anything to a variable!

leahecole (approved):

PR #2716, opened 2022-07-18
fix(run/idp-sql): update Firebase Admin to v11 + fix tests

### LINKED ISSUE #2706: run/idp-sql tests fail when upgrading to firebase-admin v11
I did not upgrade it here because I could not get tests to pass when upgrading the dependency: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2705

ace-n:
Fixes #2706

(This required some refactoring to get tests passing. I also added error message print-outs, in case this code fails in the future.)

### DIFF
--- a/run/idp-sql/Dockerfile
+++ b/run/idp-sql/Dockerfile
@@ -2,9 +2,9 @@
 # Use of this source code is governed by the Apache 2.0
 # license that can be found in the LICENSE file.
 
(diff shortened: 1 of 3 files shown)


ace-n:
@sofisl this should be good for review. 🙂

averikitsch (approved):

sofisl (approved):

PR #2717, opened 2022-07-18
feat!: upgrade Cloud Run tests to Node 14

### LINKED ISSUE #2706: run/idp-sql tests fail when upgrading to firebase-admin v11

ace-n:
Related to #2706

### DIFF
--- a/.kokoro/run/common.cfg
+++ b/.kokoro/run/common.cfg
@@ -18,7 +18,7 @@ env_vars: {
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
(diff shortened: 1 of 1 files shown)


iennae (approved):

grayside (commented):
Does this change affect node.js version testing outside Cloud Run? If so, it seems like we need to coordinate with more than just Cloud Run sample owners.

ace-n:
@grayside no, this is Cloud Run only.

sofisl (approved):

PR #2721, opened 2022-07-28
fix(issue-2719): upping timeout ms & adding try/catch

pattishin:
Resolves issue: #2719


ace-n (approved):
LGTM, assuming a 10% increase (`10000` -> `11000`) is sufficient.

PR #2730, opened 2022-08-23
chore(concepts/termination): split samples into directories

ace-n:
Example PR for b/230368998

**Questions for reviewers:**

------------------

> If we want to switch from `functions_` to `cloudfunctions_` as a region-tag prefix, should I do that:
> 1. in this PR?
> 2. as part of a separate effort?

------------------

> These samples aren't really practical to test. I assume, therefore, the lack of tests is OK?


iennae:
separate effort as the functions -> cloudfunctions is a large sized effort and involves collaboration with TW

averikitsch (approved):

PR #2737, opened 2022-08-23
chore(functions/firebase): split samples into dirs

ace-n:


.github/workflows/functions-firebase.yaml
```diff
@@ -14,6 +14,18 @@ on:
   - cron:  '0 0 * * 0'
 jobs:
   test:
+    strategy:
+      matrix:
+        # Each package in this list will be tested independently.
+        # New packages must be manually added to this list.
+        package:
+          - 'functions/firebase'
```
averikitsch, line 22: Should we have a comment that this will be removed once the original sample is deleted?

.github/workflows/functions-firebase.yaml
```diff
@@ -36,12 +48,10 @@ jobs:
       with:
         node-version: 14
     - run: npm install
-      working-directory: functions/firebase
+      working-directory: ${{ matrix.package }}
     - run: npm test
-      working-directory: functions/firebase
+      working-directory: ${{ matrix.package }}
       env:
-        MOCHA_REPORTER_SUITENAME: functions_firebase
```
averikitsch, line 43: I think these are needed for flakybot

averikitsch (approved):

.github/workflows/functions-firebase.yaml
muncus: I'll be handling the sample deletion, so I dont think its necessary in this case.

.github/workflows/functions-firebase.yaml
muncus: I could not find any evidence these were used outside of Kokoro. (https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/.kokoro/.mocharc.js)

Is there a way to see if the http sample I already moved is properly reporting to flakybot?

muncus: I found a way to add these options to the GHA-invoked test run, without eliminating the output for a manual `npm test`. wdyt?

PR #2738, opened 2022-08-25
chore: assign scheduleinstance sample ownership to TORuS team

ace-n:
@donmccasland `askmeegs` no longer owns these, so we want to point them at our team.

**Please confirm this works for you.**


iennae:
closing this for now; we'll resubmit when we have figured out the cross-language effort here to make sure this sample is ready to be owned by our team for maintenance.

PR #2742, opened 2022-09-02
samples(healthcare v1): initial consent store CRUD method samples

noerog:
New directory and initial samples for Healthcare API consent modality. Part of b/224797364.

### DIFF
--- a/healthcare/consent/package.json
+++ b/healthcare/consent/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "nodejs-docs-samples-healthcare",
+  "version": "0.0.1",
+  "private": true,
+  "license": "Apache-2.0",
+  "author": "Google LLC",
+  "repository": "GoogleCloudPlatform/nodejs-docs-samples",
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "scripts": {
+    "test": "mocha system-test/*.test.js --timeout=60000"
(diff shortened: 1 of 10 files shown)


healthcare/consent/package.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "scripts": {
+    "test": "mocha system-test/*.test.js --timeout=60000"
+  },
+  "devDependencies": {
+    "mocha": "^9.0.0"
+  },
+  "dependencies": {
+    "@googleapis/healthcare": "^3.0.0",
+    "mocha-test": "^2.0.4",
```
alexander-fenster, line 19: why is `mocha-test` needed?

healthcare/consent/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "uuid": "^8.0.0"
```
alexander-fenster, line 20: Since `uuid` is only used in tests, it should be a dev dependency.

alexander-fenster (approved):

healthcare/consent/package.json
noerog: Done.

healthcare/consent/package.json
noerog: May have added it by mistake. Removed.

PR #2745, opened 2022-09-06
fix: Update node version to LTS

kennethye1:
Flex does not support node version >=18 so far.


kennethye1:
Setting node to <18.0 results in 
`Arrow functions are not supported until Node.js 4.0.0. The configured version range is '<18.0.0' `

kennethye1:
@engelke @iennae please take a look at this PR when you can.

jinglundong (approved):
This is a reasonable fix to the version issue. Flex couldn't support the latest version of nodejs, due to an incompatibility between OS and nodejs.

PR #2746, opened 2022-09-07
Update SSL syntax in cloud sql postgres sample app.

jsimonweb:


enocom (approved):
This seems fine but a little more context in the commit would be nice. Is this a library change? Is this a better configuration option? Is this fixing an existing mistake?

jsimonweb:
>Is this fixing an existing mistake?
Yes this fixing an existing mistake.

PR #2766, opened 2022-09-22
Update busboy sample to use on('close')

WildSunLove:
Per this update, busboy should use on('close') rather than on('finish'). https://github.com/mscdex/busboy/issues/266


iennae:
Should the version of busboy be updated?

Breathtender:
Ah nice catch, done.

iennae (approved):

PR #2781, opened 2022-10-14
chore: fix error messages

averikitsch:


iennae (approved):
I think I'm a bit confused. I'm trying to model this in my head and I thought that the goal with the console.log is to send a message back to the service that there was an error with markdown rendering but then we also already see back a 500 status so it's a failure, so maybe this clears up the goals by using console.error instead. 

This feels like it'd be helpful to document a little more somewhere but looks good in general

alexander-fenster (approved):

PR #2785, opened 2022-10-18
Update Cloud SQL for SQL Server sample encrypted connection

jsimonweb:


cloud-sql/sqlserver/mssql/connect-tcp.js
```diff
@@ -35,7 +35,20 @@ const createTcpPool = async config => {
...
-  return await mssql.connect(dbConfig);
+  // [END cloud_sql_sqlserver_mssql_connect_tcp]
+
+  // (OPTIONAL) Configure encrypted connection
+  // For deployments that connect directly to a Cloud SQL instance without
+  // using the Cloud SQL Proxy, configure encrypted connection.
+  if (process.env.PRIVATE_IP) {
+    dbConfig.options += {
+      encrypt: true,
+    };
```
ruyadorno, line 46: The [addition assignment operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Addition_assignment) in JavaScript does not support objects.

I believe in this case you should just access the property directly, e.g:

```suggestion
    dbConfig.options.encrypt = true;
```

ruyadorno (changes requested):

ruyadorno (approved):

bcoe (approved):

PR #2787, opened 2022-10-25
chore: break out httpMethods sample and add test workflow

muncus:


muncus:
actions dont run from forks, so i'll re-do this as a branch from this repo instead.

PR #2788, opened 2022-10-25
Factor out a GCF sample to its own directory with tests

muncus:
This work is part of b/230368998

- chore: break out httpMethods sample and add test workflow
- fix: preserve case, in an attempt to work with actions automation.


muncus:
replaced by #2790

PR #2789, opened 2022-10-25
Move task tutorial into a subdirectory

averikitsch:


iennae (approved):
LGTM

averikitsch:
The docs have been frozen, so this move is safe.

PR #2805, opened 2022-11-07
build: run checkout before authentication

sofisl:


kweinmeister:
Updated template extension to `.njk` for better IDE support, per discussion with @sofisl

grayside:
What is the benefit of this change?

kweinmeister:
I was experiencing errors in the `mediatranslation` PR where the tests were failing, because the GitHub Actions credential file couldn't be found. This fix resolved my issues and the tests passed. This fix prevents it from occurring for others. You can see the log with these errors [here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/actions/runs/3412568318/jobs/5678248207)

bcoe (approved):

Issue #375, opened 2017-05-04
Google Speech recognize.js listen not recording on Raspberry pi 


jmp1548:
Hi All, 

I am trying the run the streamingMicRecognize function in recognize.js.
Whenever I run this on my raspberry pi (with a usb mic (that works with arecord)) the recording stream end immediately. 

Here is my code (modified slightly) 
```
const streamingMicRecognize = () => {
  console.log("Stream");
  // console.dir(recognizeStream);

  const recognizeStream = speech.createRecognizeStream(config).on('error', console.error).on('data', data => {
    console.log('recongizeStream > data');
    // record.stop();
    playback(data);
    process.stdout.write(data.results);
  }).on('end', () => {
    console.log("end stream")
  });
  // console.dir(recognizeStream);

  record.start({
    sampleRateHertz: 16000, threshold: 0,
    // Other options, see https://www.npmjs.com/package/node-record-lpcm16#options
    verbose: true,
    recordProgram: 'rec', // Try also "arecord" or "sox"
    silence: '10.0'
  }).on('error', console.error).pipe(recognizeStream);

  console.log('Listening, press Ctrl+C to stop.');
  // [END speech_streaming_mic_recognize]
}
```

When I run this with verbose mode set to true, I get this 
```
Stream
Recording with sample rate 16000...
Listening, press Ctrl+C to stop.
Recording 5526 bytes
End Recording 91.384ms
end stream
```

The recording ends before I get a chance to speak. Not sure what could be causing that. If you need any more information, please let me know.

I should also note, that this works perfectly on my mac. Seems to be specific to the pi.

jmp1548:
UPDATE: I had to manually set the device in the record settings. That fixed the issue for my pi. Now I have it working on both mac and pi.

KaKi87:
Hi, how did you find out that the proper device is `plughw:0` ?

Thanks

PR #255, opened 2016-11-16
Update Translate samples, with some minor tweaks to Language samples.

jmdobry:

PR #2813, opened 2022-11-09
samples(talent): migrate code from googleapis/nodejs-talent

telpirion:
- samples(talent): add Job Search v4 samples (#255)
- fix: flaky sample test two ways (#375)


grayside (changes requested):
Missing: package.json
Missing: test integration, but possibly there are no tests?

kweinmeister:
+1, we should include tests (see: https://github.com/googleapis/nodejs-talent/blob/main/samples/test/talent.test.js)

Also, need to run the script to create a test workflow (see CONTRIBUTING.md) and include that here, demonstrating that the tests run properly in this new location.

telpirion:
It looks like the tests are just that one test file. I'll add it into this PR.

telpirion:
Nevermind. I'm going to close this PR and start over.

Issue #2796, opened 2022-11-02
testing: dependency updates trigger global test running


grayside:
## Problem
When a particular sample has a package.json update, all sample tests are run, example: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2786.

We prefer to only run the minimal set of tests. We don't even need to see neutral test results from a workflow if it's not relevant.

## Initial Investigation

Looking at the GitHub Workflow for https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/.github/workflows/appengine-analytics.yaml

The pull_request_target does not have path filters in place, meaning the workflow will likely be triggered by any label.

```yaml
  pull_request_target:
    types: [labeled]
```

I wonder if the the conditional around whether this is a label action is correct?

```
if: ${{ github.event.action != 'labeled' || github.event.label.name == 'actions:force-run' }}
```

PR #2828, opened 2022-11-10
ci: update workflow template

kweinmeister:
This issue updates the template for #2796, #2814, and an internal issue.

This PR does not re-generate existing workflows using this template, as there are currently many outstanding PRs. It will ensure that future PRs have these changes. We can keep #2796 open until we've re-generated workflows.


NimJay (approved):
Thanks for fixing this, Karl.

bcoe (approved):

PR #2871, opened 2022-11-18
Filter by image digest rather than image tag

nataleecodes:
This example should filter by image digest (@sha256:123), similar to the samples for Python, Ruby, and others on this page: https://cloud.google.com/container-analysis/docs/automated-scanning-howto#container-analysis-discovery-python


ace-n (approved):
@nataleecodes thanks for this PR!

This LGTM, but I'd like @averikitsch to review as well. 👍

averikitsch (approved):

PR #2889, opened 2022-11-28
ci: regenerate workflows

kweinmeister:
* Updated all workflows to address an internal issue.
* Template now triggers on the workflow itself, to ensure that changes to the workflow are tested.


.github/workflows/functions-log.yaml
```diff
@@ -1,75 +0,0 @@
-name: functions-log
```
grayside, line 1: This was overridden in template re-run, and got added as standard per-function configs. Per Karl, there were also test failures related to some of the edges of the matrix config setup, such as testing directories without configured tests.

@ace-n  for awareness of this change: besides investigating this for efficiency, was there other considerations in applying the matrix?

grayside (approved):

.github/workflows/functions-log.yaml
ace-n: Discussed this internally - if a parent directory has no tests, we should remove its workflow config *and* ensure its child directories have matching config files.

.github/workflows/appengine-analytics.yaml
```diff
@@ -62,7 +64,7 @@ jobs:
               throw e;
             }
           }
-    - if: ${{ github.event_name == 'schedule' && always() }}
+    - if: ${{ github.event_name == 'schedule'}}
```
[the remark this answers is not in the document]
kweinmeister: Updated PR to add `always()` to each workflow, and to the template.

PR #2890, opened 2022-11-29
fix(functions/background-termination): fix incorrect comments

ace-n:
✅  This was tested **on my local machine**.


grayside (approved):
LGTM

Not sure what the goal is of this change, but doesn't seem significantly different.

In the future, we may want to consider if this type of code should be inline in the docs, not sure there's a benefit to being classed as a sample.

PR #2892, opened 2022-11-29
feat(batch): advanced job creation samples

Shnatsel:


batch/test/advanced_creation.js
```diff
@@ -0,0 +1,276 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const path = require('path');
+const cp = require('child_process');
+const {describe, it, before} = require('mocha');
+const {BatchServiceClient} = require('@google-cloud/batch').v1;
+const batchClient = new BatchServiceClient();
```
feywind, line 21: In other code I've seen, people frown a bit upon mixing the imports and the initializers, but this might be a YMMV thing, and it's test code, anyway.

batch/test/advanced_creation.js
```diff
@@ -0,0 +1,276 @@
...
+  });
+
+  it('create a job with instance template', async () => {
+    execSync(
+      `node create/create_script_job_with_template.js ${projectId} us-central1 test-job-js-template-${testRunId} ${templateName}`,
+      {cwd}
+    );
+  });
+
+  // waiting for jobs to succed in separate tests lets us create them all on the server and let them run in parallel,
```
feywind, line 73: Typo here in `succeed`

batch/test/advanced_creation.js
```diff
@@ -0,0 +1,276 @@
...
+  before(async () => {
+    projectId = await batchClient.getProjectId();
+    bucketName = `batch-js-test-bucket-${testRunId}`;
+    templateName = `batch-js-test-template-${testRunId}`;
+    await createBucket(bucketName);
+    await createTemplate(projectId, templateName);
+  });
+
+  it('create a job with a container payload', async () => {
+    execSync(
```
feywind, line 53: Are we catching errors from these and/or checking the sample output? We might also not want to use `async` so that exceptions flow up to mocha, since there are no `await` statements in there.

feywind: Also I'm being _incredibly_ nitpicky here :) but usually you'd word it like `it('creates a job...')` so that the mocha output flows well.

batch/test/advanced_creation.js
```diff
@@ -0,0 +1,276 @@
...
+  });
+}
+
+async function cleanupBucket(bucketName) {
+  // GCS bucket will fail to delete if it is non-empty.
+  // There is no 'force=true' argument in the Node GCS client,
+  // so we cannot use force deletion here and have to first delete all files,
+  // and only then delete the actual bucket.
+  for (let i = 0; i <= 3; i++) {
+    await deleteFileInBucket(bucketName, `output_task_${i}.txt`);
```
feywind, line 122: You might consider taking all of these promises and doing `await Promise.all()` so it doesn't block for each delete. (The deleteBucket of course would still need to be last.)

batch/test/basics.js
```diff
@@ -69,6 +61,28 @@ describe('Create, list and delete jobs', () => {
     assert(output !== null);
   });
 
+  it('get task', async () => {
+    // the server needs a bit of time to create the objects. 10 seconds is way more than enough.
+    await new Promise(resolve => setTimeout(resolve, 10000));
```
feywind, line 66: This will just delay the one test, right? In my experience tests sometimes run concurrently (well, async-concurrently) so you might have an issue with 'list tasks' below, too. I'm not super familiar with this code.

Shnatsel: I've checked the docs and the only concurrency I've found was for running different test files in parallel, in different processes. It was added in mocha v8.0.

Could you point me to an example of this async-concurrency happening? I'd love to use it if it's available, although I'd probably have to merge many of the individual tests we have now into bigger tests to support concurrent execution.

feywind: It's entirely anecdotally from working on the Node Pub/Sub library, I found that if I didn't use unique topic/subscription names for each test (guid or whatnot), they would sometimes step on each other. Admittedly that might've been an artifact of the service itself lagging on creating and deleting objects, but my understanding of it is that, when one of your tests hits `await`, it'll swap off to something else (might be another test). I'm not sure, it's worth looking into probably... like you say, it'd be useful to have explicit control over that. In general though, if you can, it's good to try to make the tests independent so that you can e.g. run one of them selectively with `-f`.

feywind (approved):
The testing comments I'll leave up to your discretion!

batch/test/basics.js
Shnatsel: I've been asked to do the exact opposite for Go samples so I'm really quite confused now :laughing:

Shnatsel: But according to both my tests and mocha documentation, tests are never run async-concurrently.

rsamborski (approved):

PR #2910, opened 2022-12-06
test: update older node versions to 12

kweinmeister:
There are several samples running on quite old Node.js versions. I believe that in one case, this is preventing upgrading dependencies (Puppeteer). This PR gradually upgrades these <12 dependencies up to 12. We can then upgrade further in additional PRs across the repo.


kweinmeister:
`node12:puppeteer` image does not look like it is configured correctly (no `gcloud` installed). Will close and try again later when time allows, as a new image may need to be created.

PR #2911, opened 2022-12-06
fix: deployment fails without resource requests

LukeSchlangen:
I was following this tutorial: https://cloud.google.com/sql/docs/postgres/connect-instance-kubernetes#node.js_1

I was getting an error saying: `Not enough CPU`. I added these resource request lines and the project started working.


ace-n:
@jsimonweb can you PTAL?

LukeSchlangen:
After running this lab a few more times, I think that Autopilot does provision the resources accordingly. The `CPU` error does show for a while, but eventually goes away. I'll close this for now. Sorry for the false alarm.

PR #2922, opened 2022-12-07
test: use dynamic import for ESM dependency

kweinmeister:
The `document-ai` samples are using a dependency `p-queue`, which has moved to a pure ESM implementation. This PR aims to enable us to continue using new versions of this dependency.

### DIFF
--- a/document-ai/package.json
+++ b/document-ai/package.json
@@ -4,7 +4,7 @@
   "license": "Apache-2.0",
   "author": "Google LLC",
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=14.0.0"
   },
   "files": [
     "*.js"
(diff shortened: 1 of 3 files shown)


document-ai/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "Apache-2.0",
   "author": "Google LLC",
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=14.0.0"
```
bcoe, line 7: I think we may want to avoid having samples that require Node 14, since our libraries are still supporting Node 12. Our best option may be avoiding updates to some of these dependencies, unless there's a forcing function (such as a security vulnerability).

I'm all for dropping Node 12 for client libraries, but we need a plan for how we announce this to customers.

kweinmeister:
Closing this experimental PR.

PR #2932, opened 2022-12-14
test PR do not merge [branch]

ace-n:
The `pull_request` clause does **not** support forks, so let's not run it for them.

Context:
> Code from forks can be insecure, which is why `pull_request` doesn't expose GitHub access tokens.
> The lack of an access token means any `pull_request` workflows will **automatically** fail.
>
>`pull_request_target` exposes these tokens, but requires prior contributor approval to run against forks.


ace-n:
We decided to leave this as-is in b/262461754

PR #2969, opened 2023-01-10
feat(speech): adds streaming transcribe, create recognizer v2 samples

telpirion:
Canonical samples are [here](https://github.com/googleapis/python-speech/pull/460).


speech/createRecognizer.v2.js
```diff
@@ -0,0 +1,58 @@
...
+        model: 'latest_long',
+      },
+    };
+
+    const operation = await client.createRecognizer(recognizerRequest);
+    const recognizer = operation[0].result;
+    const recognizerName = recognizer.name;
+    console.log(`Created new recognizer: ${recognizerName}`);
+
+    return recognizerName;
```
ace-n, line 51: Per the [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/#result), this sample should not `return` anything.

speech/system-test/createRecognizer.v2.test.js
```diff
@@ -0,0 +1,59 @@
...
+  before(async ()=>{
+    const client = new speech.SpeechClient();
+    projectId = await client.getProjectId();
+  })
+
+  beforeEach(stubConsole);
+  afterEach(restoreConsole);
+
+  it('should create a speech recognizer', async () => {
+    recognizerName = await createRecognizerV2(`rec-${uuid.v4()}`, projectId);
```
ace-n, line 49: Can we construct `recognizerName` directly in the test, instead of getting it from the sample?

(In other words - can we verify the sample correctly creates + formats the recognizer name?)

speech/transcribeStreaming.v2.js
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// sample-metadata:
+//   title: Transcribe a local file using streaming recognition
+//   description: Streaming transcription of a local file using Speech-to-Text v2
+
+async function main(recognizerName, filename = 'resources/brooklyn.flac') {
```
ace-n, line 21: The [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments) recommends minimizing the amount of arguments used.

Can we move `filename` into its own `const` line instead? e.g:
```
// TODO(developer): replace this file path with your own
const filename = 'resources/brooklyn.flac
```

speech/transcribeStreaming.v2.js
```diff
@@ -0,0 +1,90 @@
...
+      .on('close', () => {
+        // Config-only request should be first in stream of requests
+        stream.write(configRequest);
+        for (const chunk of chunks) {
+          stream.write(chunk);
+        }
+        stream.end();
+      });
+
+    const stream = client
```
ace-n, line 75: We should use `Promise.{resolve, reject}()` and `await` here, instead of returning `stream`.

It might look something like this:
```
await new Promise((resolve, reject) => {
  client
      ._streamingRecognize()
      .on('data', response => {
          const {results} = response;
          const {transcript} = results[0].alternatives[0];
          return Promise.resolve(transcript);
      })
      .on('error', err => {
          /* omit console.error() here - rejecting a Promise here is equivalent to throwing an exception */
          return Promise.reject(err);
      })

console.log(transcript);
```

speech/transcribeStreaming.v2.js
```diff
@@ -0,0 +1,90 @@
...
+      ._streamingRecognize()
+      .on('data', response => {
+        const {results} = response;
+        console.log(results[0].alternatives[0].transcript);
+      })
+      .on('error', err => {
+        console.error(err.message);
+      });
+    return stream;
```
ace-n, line 84: As mentioned above - avoid `return`s where possible.

ace-n (changes requested):

speech/createRecognizer.v2.js
telpirion: Done.

speech/system-test/createRecognizer.v2.test.js
telpirion: That's not quite how it works. The recognizerName (a resource URI) is returned by the API; the sample just prints it to the console.. We can (re-)compose it here in the test since we know what the recognizerId is.

I've made some changes to make the test clearer.

speech/transcribeStreaming.v2.js
telpirion: We can minimize the number of arguments, but I choose not to. I'm hewing to the "Copy-paste-runnable" goal in the [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/#introduction).

speech/transcribeStreaming.v2.js
telpirion: Done. I like this way a lot better!

speech/transcribeStreaming.v2.js
telpirion: Done.

speech/transcribeStreaming.v2.js
telpirion: Sorry, I misunderstood your comment. Done.

speech/system-test/createRecognizer.v2.test.js
```diff
@@ -0,0 +1,61 @@
...
+  before(async ()=>{
+    const client = new speech.SpeechClient();
+    projectId = await client.getProjectId();
+  })
+
+  beforeEach(stubConsole);
+  afterEach(restoreConsole);
+
+  it('should create a speech recognizer', async () => {
+    recognizerId = `rec-${uuid.v4()}`;
```
ace-n, line 49: **Nit:** `recognizerId` doesn't change here, so can we specify it near the top of the file?

e.g. remove the `recognizerId` in line 27 in favor of:
```
const recognizerId = `rec-${uuid.v4()}`;
```

speech/system-test/transcribeStreaming.v2.test.js
```diff
@@ -0,0 +1,74 @@
...
+      recognizerId: `rec-${uuid.v4()}`,
+      recognizer: {
+        languageCodes: ['en-US'],
+        model: 'latest_long',
+      },
+    };
+
+    const operation = await client.createRecognizer(recognizerRequest);
+    const recognizer = operation[0].result;
+    recognizerName = recognizer.name;
```
ace-n, line 57: Can this be calculated based on the `recognizerId` parameter?

If it can, can we move `recognizerId` into a `const` and use that instead?

> **IMHO** it's good practice to _minimize reliance on client library logic_ within tests.

speech/system-test/transcribeStreaming.v2.test.js
```diff
@@ -0,0 +1,74 @@
...
+    sinon.stub(console, 'log');
+  }
+
+  const restoreConsole = function () {
+    console.log.restore();
+    console.error.restore();
+  }
+
+  before(async ()=>{
+    const client = new speech.SpeechClient();
```
ace-n, line 43: Nit: we usually reuse the same `client` instance across multiple tests.

speech/transcribeStreaming.v2.js
```diff
@@ -0,0 +1,96 @@
...
+      .on('close', () => {
+        // Config-only request should be first in stream of requests
+        writeStream.write(configRequest);
+        for (const chunk of chunks) {
+          writeStream.write(chunk);
+        }
+        writeStream.end();
+      });
+
+    await new Promise((resolve, reject) => {
```
ace-n, line 76:
```suggestion
    const transcript = await new Promise((resolve, reject) => {
```

speech/transcribeStreaming.v2.js
```diff
@@ -0,0 +1,96 @@
...
+        .on('data', response => {
+          const {results} = response;
+          const {transcript} = results[0].alternatives[0];
+          return resolve(transcript);
+        })
+        .on('error', err => {
+          console.error(err.message);
+          return reject(err);
+        });
+    }).then(transcript => {
```
ace-n, line 88: If you do `const transcript = await ...` above, you can remove the `then` block and just do `console.log(transcript)`.

speech/system-test/createRecognizer.v2.test.js
ace-n: Got it; thanks for addressing this.

In general, reusing API data is OK _if and only if_ that API data cannot determined by the test itself:
> - If a particular property's value is _randomly chosen by the API_, then (re)using **or** loosely validating that value makes sense.
> - If that property's value can be _calculated within the test itself_, then (where possible) we should try to do that.

ace-n (approved):
LGTM once comments addressed. 🥳

PR #2974, opened 2023-01-11
feat(speech): adds v2 adaptation samples

telpirion:
Canonical samples are [here](https://github.com/googleapis/python-speech/pull/460).


yil532 (approved):
LGTM

speech/system-test/adaptationCustomClassReference.v2.test.js
```diff
@@ -0,0 +1,91 @@
...
+const {describe, it} = require('mocha');
+const sinon = require('sinon');
+const uuid = require('uuid');
+const speech = require('@google-cloud/speech').v2;
+
+const {createCustomClassV2} = require('../adaptationCustomClassReference.v2');
+
+const text = 'how old is the Brooklyn Bridge';
+
+let recognizerName, projectId, customClassName, phraseSetName;
```
ace-n, line 27: Nit: can these be defined as `const`s here?

speech/system-test/adaptationInlineCustomClass.v2.test.js
```diff
@@ -0,0 +1,70 @@
...
+const {describe, it} = require('mocha');
+const sinon = require('sinon');
+const uuid = require('uuid');
+const speech = require('@google-cloud/speech').v2;
+
+const {buildInlineCustomClassV2} = require('../adaptationInlineCustomClass.v2');
+
+const text = 'how old is the Brooklyn Bridge';
+
+let recognizerName, projectId;
```
ace-n, line 27: Nit: defined as `const`s here?

speech/system-test/adaptationInlineCustomClass.v2.test.js
```diff
@@ -0,0 +1,70 @@
...
+      recognizerId: `rec-${uuid.v4()}`,
+      recognizer: {
+        languageCodes: ['en-US'],
+        model: 'latest_short',
+      },
+    };
+
+    const operation = await client.createRecognizer(recognizerRequest);
+    const recognizer = operation[0].result;
+    recognizerName = recognizer.name;
```
ace-n, line 55: If this value is deterministic, we should compute it within the test itself.

(If this is **not** deterministic, this line is OK as-is.)

speech/system-test/adaptationInlineCustomClass.v2.test.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  beforeEach(stubConsole);
+  afterEach(restoreConsole);
+
+  it('should create an inline class reference and phrase set', async () => {
+    await buildInlineCustomClassV2(recognizerName);
+    expect(console.log.firstCall.args[0]).to.contain(text);
```
ace-n, line 63: We previously decided to move off of `expect` syntax to the [Node.js `assert` API](https://nodejs.org/api/assert.html).

_For consistency_, can we use the `assert` API here instead?

ace-n (changes requested):

speech/system-test/adaptationCustomClassReference.v2.test.js
telpirion: No, they're not deterministic as the names will depend on the projectId.

speech/system-test/adaptationInlineCustomClass.v2.test.js
telpirion: No, they're not deterministic as the names will depend on the projectId.

speech/system-test/adaptationInlineCustomClass.v2.test.js
telpirion: No, they're not deterministic as the names will depend on the projectId.

speech/system-test/adaptationInlineCustomClass.v2.test.js
telpirion: Done.

ace-n (approved):

PR #2984, opened 2023-01-13
chore(functions/concepts): Migrate backgroundTermination sample to gen2

ddegraf:
Updated backgroundTermination sample using gen2 signature.
Successfully tested on GCP (function successfully starts + same behaviour as gen1).


ace-n:
This function is not an HTTP function. As such, it does **not** need to be updated right now.

PR #2989, opened 2023-01-14
docs: add Media CDN sample for Video Stitcher; move Akamai sample to …

irataxy:
…its own file

Fixes b:248290230


media/video-stitcher/createCdnKeyAkamai.js
```diff
@@ -0,0 +1,61 @@
...
+  createCdnKeyAkamai();
+  // [END videostitcher_create_cdn_key_akamai]
+}
+
+// node createCdnKeyAkamai.js <projectId> <location> <cdnKeyId> <hostname> <akamaiTokenKey>
+process.on('unhandledRejection', err => {
+  console.error(err.message);
+  process.exitCode = 1;
+});
+main(...process.argv.slice(2));
```
ace-n, line 61: For new samples, we try to avoid adding CLIs per the [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/#no-cli).

**However** if these are existing samples migrated from elsewhere, this is OK.

media/video-stitcher/updateCdnKeyAkamai.js
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function main(projectId, location, cdnKeyId, hostname, akamaiTokenKey) {
```
ace-n, line 18: Per the [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments), we should minimize the number of arguments our samples have.

Can we hard-code _some_ of these values? (e.g. `location`, `cdnKeyId`, and `hostname`?)

media/video-stitcher/updateCdnKeyAkamai.js
```diff
@@ -0,0 +1,63 @@
...
+  updateCdnKeyAkamai();
+  // [END videostitcher_update_cdn_key_akamai]
+}
+
+// node updateCdnKeyAkamai.js <projectId> <location> <cdnKeyId> <hostname> <akamaiTokenKey>
+process.on('unhandledRejection', err => {
+  console.error(err.message);
+  process.exitCode = 1;
+});
+main(...process.argv.slice(2));
```
ace-n, line 63: As above - if this is a new sample, can we avoid using CLIs here?

ace-n (changes requested):

media/video-stitcher/createCdnKeyAkamai.js
irataxy: These samples are migrated. I'd like to keep the CLIs to match the rest of the samples.

media/video-stitcher/updateCdnKeyAkamai.js
irataxy: This code was migrated from updateCdnKey.js.

media/video-stitcher/updateCdnKeyAkamai.js
irataxy: Added some default values for most of the parameters.

media/video-stitcher/updateCdnKeyAkamai.js
ace-n: From @irataxy:
> These samples are migrated. I'd like to keep the CLIs to match the rest of the samples.

media/video-stitcher/updateCdnKeyAkamai.js
ace-n: Ah - sorry for not being specific! I was referring to declaring these _as their own values_ (i.e. `const ... = ...`).

**Rationale:**
> Default arguments are better than those without defaults, but they're still more confusing to end-users than explicitly-declared `const`s.

ace-n (changes requested):
@irataxy thanks for these changes!

**My one reservation:** we should use `const`s for parameters instead of default arguments. 🙂

media/video-stitcher/updateCdnKeyAkamai.js
irataxy: Moved some of these to explicitly-declared consts

kweinmeister (approved):

PR #2990, opened 2023-01-14
docs(speech): migrate samples

meredithslota:
Copied from: 
https://github.com/googleapis/nodejs-speech/blob/main/samples/modelAdaptation.js and https://github.com/googleapis/nodejs-speech/blob/main/samples/export-to-storage.v1p1beta1.js


speech/export-to-storage.v1p1beta1.js
```diff
@@ -0,0 +1,104 @@
...
+function main(
+  inputUri,
+  outputStorageUri,
+  encoding,
+  sampleRateHertz,
+  languageCode,
+  bucketName,
+  objectName
+) {
+  // [START speech_export_to_gcs]
```
ace-n, line 26: This sample is not (yet) used in our docs.

@meredithslota I presume you're planning to add it to `cloud.google.com` somewhere?

ace-n (changes requested):
Do these samples have existing tests?

(If not, can we file a follow-up issue to add tests later?)

speech/export-to-storage.v1p1beta1.js
kweinmeister: @ace-n actually I was able to find this sample in the docs by searching on the path rather than region tag.

+1 about tests. There are a couple matching test files we could add to this PR from https://github.com/googleapis/nodejs-speech/tree/main/samples/system-test.

kweinmeister (approved):

PR #2999, opened 2023-01-27
ci: Add header checker

kweinmeister:
Adds the [License Header bot](https://github.com/googleapis/repo-automation-bots/tree/main/packages/header-checker-lint) configuration based on [Google Cloud Node](https://github.com/googleapis/google-cloud-node/blob/main/.github/header-checker-lint.yml) and [Python Docs Samples](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/.github/header-checker-lint.yml).

Also, added a header to the workflow generation template, so new workflows are compliant. We can address adding missing headers as we go or in a separate PR.

### DIFF
--- a/.github/header-checker-lint.yml
+++ b/.github/header-checker-lint.yml
@@ -1,0 +1,33 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


grayside (approved):

grayside:
I'm assuming this check will not be marked required until we are in compliance.

iennae (approved):
LGTM

kweinmeister:
> I'm assuming this check will not be marked required until we are in compliance.

Sounds good @grayside. We can leave the check optional until it's passing consistently.

PR #3000, opened 2023-01-27
chore: Set ignore year flag in header check

kweinmeister:
We are still doing migrations, and this setting is unnecessarily flagging valid files. This setting matches the current setting in python-docs-samples.


kweinmeister:
Closing for now, as the license year is not a required check. Will reopen if needed later.

Issue #2864, opened 2022-11-17
Tests for `asset/snippets/test/` need to be refactored


ahrarmonsur:
The language repo migration for `googleapis/nodejs-asset` created the following PR:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2856

The PR was blocked due to failing tests, which are all in one file`tests/samples.test.js`. The advice from @grayside to unblock the migration PR was to:
- keep the tests in the migration
- remove the workflow for the `asset` samples
- create an issue capturing the follow-on work

The language maintainers are to:
- refactor the tests into individual files corresponding to the sample
- generate the workflow for `asset/snippets` again

This is captured in [this conversation](https://chat.google.com/room/AAAAQVhDuZk/k3xZ2L10Hrk).

PR #3005, opened 2023-01-28
fix: refactor asset tests

kweinmeister:
Asset tests were previously combined into 1 file. Per #2864, these have been refactored and a CI workflow created.

Note: 2 samples were identified and skipped that are returning an invalid argument error, related to asset relationships. These were unchanged in the refactoring. I will file a separate issue with the product team to resolve these.


Shnatsel (approved):
LGTM. Thanks!

grayside (approved):

PR #3016, opened 2023-02-01
chore: include ff google team as functions samples owners

kenneth-rosario:
Include functions-framework-google team as functions samples owner


iennae (approved):
LGTM; we probably need to segment these a little more in the directory, but this is great. Thanks!

iennae:
I've also added the team-synced team to the repository so that should resolve the error.

PR #3018, opened 2023-02-04
Re-enable dependency dashboard

kweinmeister:
Re-enable dependency dashboard (at least temporarily), to provide visibility into outstanding updates.

Also enabled setting to auto-close the dashboard when there are no open PRs.


bcoe (approved):
:shipit:

PR #3054, opened 2023-02-25
fix: update vision tests

kweinmeister:
* Update to Redis 4
* Remove a couple tests for a beta API that no longer supports those features
* Fix and un-skip several tests


yil532 (approved):
LGTM

PR #3056, opened 2023-02-26
fix: update appengine storage bucket usage

kweinmeister:


iennae (approved):
LGTM; I'm not sure why the additional dependencies added to package.json, assuming those are things you verified as working.

kweinmeister:
Thanks @iennae!

PR #3059, opened 2023-02-27
docs(recaptcha-samples): add sample for password leak assessment

Sita04:
Added Typescript code sample and test for Password Leak detection.
https://cloud.google.com/recaptcha-enterprise/docs/check-passwords

### DIFF
--- a/.github/workflows/ai-platform-snippets.yaml
+++ b/.github/workflows/ai-platform-snippets.yaml
@@ -44,7 +44,7 @@ jobs:
       run:
         working-directory: 'ai-platform/snippets'
     steps:
-    - uses: actions/checkout@v3.3.0
+    - uses: actions/checkout@v3.5.2
       with:
         ref: ${{github.event.pull_request.head.sha}}
     - uses: 'google-github-actions/auth@v1.0.0'
--- a/.github/workflows/automl.yaml
+++ b/.github/workflows/automl.yaml
@@ -44,7 +44,7 @@ jobs:
       run:
         working-directory: 'automl'
     steps:
-    - uses: actions/checkout@v3.3.0
+    - uses: actions/checkout@v3.5.2
       with:
         ref: ${{github.event.pull_request.head.sha}}
     - uses: 'google-github-actions/auth@v1.0.0'
--- a/.github/workflows/dialogflow-cx.yaml
+++ b/.github/workflows/dialogflow-cx.yaml
@@ -44,7 +44,7 @@ jobs:
       run:
         working-directory: 'dialogflow-cx'
     steps:
-    - uses: actions/checkout@v3.3.0
+    - uses: actions/checkout@v3.5.2
       with:
         ref: ${{github.event.pull_request.head.sha}}
     - uses: 'google-github-actions/auth@v1.0.0'
--- a/.github/workflows/functions-slack.yaml
+++ b/.github/workflows/functions-slack.yaml
@@ -44,7 +44,7 @@ jobs:
       run:
         working-directory: 'functions/slack'
     steps:
-    - uses: actions/checkout@v3.3.0
+    - uses: actions/checkout@v3.5.2
       with:
         ref: ${{github.event.pull_request.head.sha}}
     - uses: 'google-github-actions/auth@v1.0.0'
--- a/.github/workflows/security-center-snippets.yaml
+++ b/.github/workflows/security-center-snippets.yaml
@@ -44,7 +44,7 @@ jobs:
       run:
         working-directory: 'security-center/snippets'
     steps:
-    - uses: actions/checkout@v3.3.0
+    - uses: actions/checkout@v3.5.2
       with:
         ref: ${{github.event.pull_request.head.sha}}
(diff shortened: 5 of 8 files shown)


recaptcha_enterprise/snippets/passwordLeakAssessment.js
```diff
@@ -0,0 +1,224 @@
...
+* Detect password leaks and breached credentials to prevent account takeovers
+* (ATOs) and credential stuffing attacks.
+* For more information, see:
+* https://cloud.google.com/recaptcha-enterprise/docs/getting-started and
+* https://security.googleblog.com/2019/02/protect-your-accounts-from-data.html
+
+* Steps:
+* 1. Use the 'create' method to hash and Encrypt the hashed username and
+* password.
+* 2. Send the hash prefix (2-byte) and the encrypted credentials to create
```
serargz, line 69: The username hash prefix is 26-bit long: https://github.com/GoogleCloudPlatform/java-recaptcha-password-check-helpers/blob/main/src/main/java/com/google/cloud/recaptcha/passwordcheck/PasswordCheckVerification.java#L46

serargz: Needs to be updated here too: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/HEAD/recaptcha_enterprise/snippets/src/main/java/recaptcha/passwordleak/CreatePasswordLeakAssessment.java#L66

recaptcha_enterprise/snippets/passwordLeakAssessment.js
```diff
@@ -0,0 +1,224 @@
...
+* the assessment.(Hash prefix is used to partition the database.)
+* 3. Password leak assessment returns a database whose prefix matches the
```
serargz, line 71: Maybe more accurate: returns a list of encrypted credential hashes to be compared with the decryption of the returned re-encrypted credentials

recaptcha_enterprise/snippets/passwordLeakAssessment.js
```diff
@@ -0,0 +1,224 @@
...
+  encryptedUserCredentialsHash
+) {
+  // Build the assessment request.
+  const createAssessmentRequest = {
+    parent: `projects/${projectId}`,
+    assessment: {
+      // Set the properties of the event to be tracked.
+      event: {
+        siteKey: recaptchaSiteKey,
+        token: token,
```
serargz, line 169: nit: the reCAPTCHA token is not mandatory for password check requests (in fact, it will count as an extra assessment call)

recaptcha_enterprise/snippets/passwordLeakAssessment.ts
```diff
@@ -0,0 +1,220 @@
+/**
```
serargz, line 1: Same comments as the js file. Why are there 2 files? was the js one commited by mistake?

serargz (changes requested):
In addition to the comments, I tried to build the package unsuccessfully. Got jasmine/mocha compatibility errors, have you faced any errors when running `npm run build` (after `npm install`)?

recaptcha_enterprise/snippets/passwordLeakAssessment.ts
Sita04: No, it's WAI.
We do not have a separate repository for typescript. Hence all ts code will be compiled to Javascript and the tests will be executed.

recaptcha_enterprise/snippets/passwordLeakAssessment.js
Sita04: Thanks, will update Java as well.

recaptcha_enterprise/snippets/passwordLeakAssessment.js
Sita04: Thanks for the info. I've removed it from the sample. 
Double-checking the docs, I see that the API request doesn't include the site key or the token. But I think it should be explicitly called out in the docs as well. Since, all the other recaptcha requests include both those params, it could be mistaken as a mandatory param for all requests involving recaptcha.
Plmk your thoughts. @serargz

Sita04:
@kweinmeister The sample is ready for review. PTAL. Thanks!
Although I see many tests running, I couldn't verify which one executes the sample tests?

recaptcha_enterprise/snippets/passwordLeakAssessment.js
serargz: yeah, that's a good point. We should update the docs to make this clear.

msampathkumar (approved):
LGTM. Thank you for your work

kweinmeister:
You will need to follow the steps [here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#adding-new-samples) in the Adding New Samples section. That should generate a new workflow for this sample directory, and then we should see tests running in the PR.

Sita04:
Thanks @kweinmeister
The tests are running now.

kweinmeister (approved):

PR #3068, opened 2023-03-04
feat: add typescript to workflows

kweinmeister:


owlbot.py
```diff
@@ -15,14 +15,21 @@
...
 
 from synthtool import shell
 from synthtool.log import logger
 
+logger: Logger = logger
+
 _TOOLS_DIRECTORY = "/synthtool"
 _EXCLUDED_DIRS = [r"node_modules", r"^\."]
-_TYPELESS_EXPRESSION = "s/export {};$/\\n/"
+_TRIM_EXPRESSIONS = [
```
bcoe, line 28: I'd create a tracking ticket for each of these replacements required, assigned to @feywind as a feature request in googleapis/google-clou-node.

@feywind, I know this is probably a secondary priority currently, so don't take this as a request you need to act on immediately.

workflows/create-execution.js
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This is a generated sample, using the typeless sample bot. Please
```
bcoe, line 15: this caveat is a good addition.

workflows/quickstart/index.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Executes a Workflow and waits for the results with exponential backoff.
+ * @param {string} projectId The Google Cloud Project containing the workflow
+ * @param {string} location The workflow location
+ * @param {string} workflow The workflow name
+ */
+const main = async (projectId: string, location: string, workflow: string) => {
+  if (!projectId)
+    return console.error('ERROR: GOOGLE_CLOUD_PROJECT is required.');
+
+  const client: ExecutionsClient = new ExecutionsClient();
```
bcoe, line 28: nit: It's probably best practice to show the instantiation of the client outside of the main function, to discourage people form creating a client on every request if they're using this in a web service.

```
import {ExecutionsClient} from '@google-cloud/workflows';
const client: ExecutionsClient = new ExecutionsClient();

main() {}
```

workflows/quickstart.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+  listWorkflows();
+}
+// [END workflows_quickstart]
+
+process.on('unhandledRejection', (err: Error) => {
+  console.error(err.message);
+  process.exitCode = 1;
+});
+main(...(process.argv.slice(2) as [string, string]));
```
bcoe, line 41: I would personally pull the invocation of main into the sample itself, because I'm not confident users will understand that they need to call the method, so:


```js
const [projectId, location] = process.argv.slice(2);
// [START workflows_quickstart]
main(projectId, location);
// [END workflows_quickstart]
process.on('unhandledRejection', (err: Error) => {
  console.error(err.message);
  process.exitCode = 1;
});
```

Or, alternatively:

```js
const [projectId, location] = process.argv.slice(2);
// [START workflows_quickstart]
main(projectId, location).catch((err: Error) = {
  console.error(err.message);
  process.exitCode = 1;
});
// [END workflows_quickstart]
```

workflows/quickstart/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export default main;
+
+// Call as CLI
+// node . [projectId] [location] [workflowName]
+const projectID =
+  process.argv[2] || (process.env.GOOGLE_CLOUD_PROJECT as string);
+const location = process.argv[3] || 'us-central1';
+const workflowName = process.argv[4] || 'myFirstWorkflow';
+main(projectID, location, workflowName);
```
bcoe, line 83: same suggestion, where I would pull the execution of the method into the sample itself -- then you either want the `catch` or to handle an `uncaughtRejection`, or else this will throw an unhandledRejection error on failures (rather than the actual underlying error).

bcoe (changes requested):
Left some suggestions.

owlbot.py
kweinmeister: Filed https://github.com/googleapis/google-cloud-node/issues/4057

bcoe (approved):
Approved, with a note that I would like us to figure out some consistent standards for TypeScript along with standards for exposing Node as a script, vs., copy-pastable code in a repl.

PR #3069, opened 2023-03-06
chore: Adding self-service DLP team to labels and blunderbuss

rsamborski:


.github/auto-label.yaml
```diff
@@ -62,3 +62,4 @@ path:
     video-intelligence: "videointelligence"
     vision: "vision"
     workflows: "workflows"
+    dlp: "dlp"
```
kweinmeister, line 65: Nit: recommend keeping the list sorted alphabetically

kweinmeister (approved):

.github/blunderbuss.yml
```diff
@@ -14,6 +14,10 @@ assign_issues_by:
   - 'api: cloudfunctions'
   to:
   - GoogleCloudPlatform/aap-dpes
+- labels:
```
kweinmeister, line 17: Nit: ideally, we add license headers as we go to address header-check failures. Here's an existing YAML example: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/.github/.OwlBot.yaml

.github/auto-label.yaml
rsamborski: Done

PR #3090, opened 2023-03-17
Update Dockerfile

marwie:
fix node version


kweinmeister:
Thank you for the contribution! In #3133 , we removed the version number from the comment to avoid mismatches. We'll go and close this change for now. Please let us know if you have any questions.

PR #3100, opened 2023-04-03
docs: move region tags to fix docs

camiekim:
Per b/276737524

### DIFF
--- a/workflows/quickstart/index.js
+++ b/workflows/quickstart/index.js
@@ -14,22 +14,14 @@
 
 // This is a generated sample, using the typeless sample bot. Please
 // look for the source TypeScript sample (.ts) for modifications.
+// [START workflows_api_quickstart]
 'use strict';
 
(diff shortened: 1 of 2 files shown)


iennae (changes requested):
1. We shouldn't touch index.js; it's auto-generated

2. Removing the comment and moving the region tag  I think breaks a standard we have to let the developer uncomment these out.  I think that the problem is line 32 in the original file should probably be documenting const workflowName = 'myFirstWorkflow';

3 Do you happen to know why we have this quickstart as well as the one in the level above labeled quickstart.js ? These are different quickstarts and it confused me at first, so I think that might be distracting to others as well.

camiekim:
Thanks, Jennifer. I'll close this PR and open a new one. Regarding why there are two samples, `workflows/quickstart.js` is used in https://cloud.google.com/workflows/docs/reference/libraries and `workflows/quickstart/index.js` is used in https://cloud.google.com/workflows/docs/execute-workflow-client-libraries. I'll look into why this is and cleaning up as appropriate.

PR #3107, opened 2023-04-06
DLP: Added sample for inspect string with custom regex

dinesh-crest:
Added sample for inspect string with custom regex
Added unit test cases for the same

Reference: https://cloud.google.com/dlp/docs/samples/dlp-inspect-custom-regex

### DIFF
--- a/dlp/inspectWithCustomRegex.js
+++ b/dlp/inspectWithCustomRegex.js
@@ -17,18 +17,9 @@
 // sample-metadata:
 //  title: Inspects strings
 //  description: Inspects a string using custom regex pattern.
-//  usage: node inspectWithCustomRegex.js my-project string minLikelihood maxFindings infoTypes customInfoTypes includeQuote
+//  usage: node inspectWithCustomRegex.js my-project string customRegex
 
-function main(
-  projectId,
-  string,
-  minLikelihood,
-  maxFindings,
-  infoTypes,
-  customInfoTypes,
-  includeQuote
-) {
-  [infoTypes, customInfoTypes] = transformCLI(infoTypes, customInfoTypes);
+function main(projectId, string, customRegex) {
   // [START dlp_inspect_custom_regex]
   // Imports the Google Cloud Data Loss Prevention library
   const DLP = require('@google-cloud/dlp');
@@ -42,23 +33,8 @@ function main(
   // The string to inspect
   // const string = 'Patients MRN 444-5-22222';
 
-  // The minimum likelihood required before returning a match
-  // const minLikelihood = DLP.protos.google.privacy.dlp.v2.Likelihood.POSSIBLE;
-
-  // The maximum number of findings to report per request (0 = server maximum)
-  // const maxFindings = 0;
-
-  // The infoTypes of information to match
-  // See https://cloud.google.com/dlp/docs/concepts-infotypes for more information
-  // about supported infoTypes.
-  // const infoTypes = [{ name: 'EMAIL_ADDRESS' }];
-
-  // The customInfoTypes of information to match
-  // const customInfoTypes = [{ infoType: { name: 'DICT_TYPE' }, dictionary: { wordList: { words: ['foo', 'bar', 'baz']}}},
-  //   { infoType: { name: 'REGEX_TYPE' }, regex: {pattern: '\\(\\d{3}\\) \\d{3}-\\d{4}'}}];
-
-  // Whether to include the matching string
-  // const includeQuote = true;
+  // The regex pattern to match for
+  // const customRegex = '[1-9]{3}-[1-9]{1}-[1-9]{5}';
 
   async function inspectWithCustomRegex() {
     // Construct item to inspect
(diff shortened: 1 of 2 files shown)


realjordanna (approved):

dlp/inspectWithCustomRegex.js
```diff
@@ -0,0 +1,145 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// sample-metadata:
+//  title: Inspects strings
+//  description: Inspects a string using custom regex pattern.
+//  usage: node inspectWithCustomRegex.js my-project string minLikelihood maxFindings infoTypes customInfoTypes includeQuote
```
soumya92, line 20: Can we trim this down?

- `string`: required
- `minLikelihood`: we should be able to remove this because the likelihood for custom infotype (which is how regex is used) is controlled by the request anyway.
- `maxFindings`: does not make sense in the context of this sample. We're inspecting a small-ish string, not megabytes of file content, so there will only be so many findings. If you want to limit it anyway, just hardcode a limit of ~1000 in the sample. Users can change it if they need to.
- `infoTypes`: Can omit, since we're demonstrating regex matching only. I guess it's possible we may want to show side-by-side detection of custom and built-in infotypes, but if that's the case move this to the end and make it optional. (Also if that is the case, lets make the example string actually demonstrate that)
- `customInfoTypes`: Since the whole point of this sample is to demonstrate regex, we should ask for regex directly and construct the custom infotype in code.
- `includeQuote`: as with maxFindings, lets just set this to true for demo purposes. If users want to change it they can edit the code.

At a high level we should make the sample as easy as possible to run. Adding a lot of parameters and using obscure syntax (such as the ',' and regex/dict hybrid for customInfoTypes) will lead to confusion and frustration.

As a user of this sample, I should be able to say `node inspectWithCustomRegex.js 'this is my serial number aab-bcdd-eef' '[a-f\-]10'` and see results.

dlp/inspectWithCustomRegex.js
```diff
@@ -0,0 +1,145 @@
...
+      });
+    } else {
+      console.log('No findings.');
+    }
+  }
+  inspectWithCustomRegex();
+  // [END dlp_inspect_custom_regex]
+}
+
+main(...process.argv.slice(2));
```
soumya92, line 115: This should be after process.on to avoid missing synchronous promise rejections.

dlp/system-test/inspect.test.js
```diff
@@ -309,4 +309,38 @@ describe('inspect', () => {
     assert.notMatch(outputB, /EMAIL_ADDRESS/);
     assert.match(outputB, /PHONE_NUMBER/);
   });
+
+  // dlp_inspect_custom_regex
+  it('should inspect a string using custom regex', () => {
+    const string = 'Patients MRN 444-5-22222';
+    const custRegex = '[1-9]{3}-[1-9]{1}-[1-9]{5}';
+    const output = execSync(
+      `node inspectWithCustomRegex.js ${projectId} "${string}" UNLIKELY 0 "" ${custRegex} true`
```
soumya92, line 318: Quote the regex similar to string to avoid user confusion and test failures if the regex is something that bash could expand. (e.g. `[ab][cd]` would become `ac` if such a file existed)

soumya92 (changes requested):

dlp/inspectWithCustomRegex.js
dinesh-crest: @soumya92 I had the same thought when I started the implementation but I noticed a particular structure is being followed for all inspect samples. Couldn't figure out the exact reason but mostly it was to keep the sample code consistent. Anyway, I feel your findings look reasonable and so I have updated this sample. Also, will it be okay if I make these same changes in my other PRs?

dlp/inspectWithCustomRegex.js
```diff
@@ -0,0 +1,145 @@
...
+      },
+      item: item,
+    };
+
+    // Run request
+    const [response] = await dlp.inspectContent(request);
+    const findings = response.result.findings;
+    if (findings.length > 0) {
+      console.log('Findings:');
+      findings.forEach(finding => {
```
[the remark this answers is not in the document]
dinesh-crest: Looks, better, Updated the code. Thanks

dlp/inspectWithCustomRegex.js
dinesh-crest: Is there any use case where this can happen? I have updated the code as you mentioned but during testing, I found the same results.

dlp/inspectWithCustomRegex.js
```diff
@@ -0,0 +1,145 @@
...
+    const item = {
+      byteItem: {
+        type: DLP.protos.google.privacy.dlp.v2.ByteContentItem.BytesType
+          .TEXT_UTF8,
+        data: Buffer.from(string, 'utf-8'),
+      },
+    };
+
+    // Assigns likelihood to each match
+    customInfoTypes = customInfoTypes.map(customInfoType => {
```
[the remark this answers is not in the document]
dinesh-crest: Update code, now we are constructing the customInfoType in the code only instead of passing that as a param. Defining the likelihood property there as well.

dlp/inspectWithCustomRegex.js
soumya92: Yeah go for it! The easier we make our samples to use, the better

dlp/inspectWithCustomRegex.js
soumya92: I don't think it happens right now, but I could imagine in the future there might be synchronous validation of requests (e.g. bad enum values might throw before even making a network request).

soumya92 (approved):

kweinmeister (approved):

PR #3112, opened 2023-04-07
ci: use flakybot with matrix workflows

kweinmeister:
Refactored the 5 `functions` workflows that currently use a matrix strategy, to use a common workflow template that includes FlakyBot 1.2+.

Considerations:
* Using the matrix strategy has a slightly different syntax. It did not look possible to integrate these workflows into the standard template using the tools available in GH Actions today.
* Also, these workflows require manually specifying subfolders within the matrix. So, it is safer to handle these workflows separately when regenerating workflows.

Changes:
* **Workflow generation**
  * Creation of a new "matrix" workflow template to handle this unique structure: `ci-matrix.yaml.njk`
  * A `workflows-matrix.json` array of workflows using a matrix strategy was created. These workflows require the workflow developer to manually specify the subdirectories that need to be tested within them. So, they shouldn't be included along with all of the other workflows that can be safely automatically generated.
  * CLI option in `generate.js` to specify the matrix template.
* **Workflow updates**
  *  All existing matrix workflows were updated to use the standardized workflow.
* **Sample updates**
  * `functions/concepts` samples are small gists that do not have tests. To enable using the reusable `test.yaml`, a no-op `npm test` script was added.


package.json
```diff
@@ -31,6 +31,7 @@
     "@google-cloud/resource-manager": "^4.1.1",
     "@google-cloud/storage": "^6.7.0",
     "chai": "^4.3.7",
+    "commander": "^10.0.0",
```
bcoe, line 34: Since this is a GitHub Action, why don't you try out:

https://nodejs.org/docs/latest-v18.x/api/util.html#utilparseargsconfig

Rather than pulling in a dep, you just need to set the Node version to 18.

.github/workflows/ci-matrix.yaml.njk
```diff
@@ -0,0 +1,53 @@
...
+          - 'example/subfolder/2'
+    uses: ./.github/workflows/test.yaml
+    with:
+      name: '{{name}}'
+      path: '${% raw %}{{ matrix.path }}{% endraw %}'
+  remove_label:
+    if: github.event.action == 'labeled' && github.event.label.name == 'actions:force-run' && always()
+    uses: ./.github/workflows/remove-label.yaml
+  flakybot:
+    if: github.event_name == 'schedule' && always()
```
bcoe, line 51: Might be worth adding a comment about the use of `always()`, I had to read the docs.

.github/workflows/ci-matrix.yaml.njk
```diff
@@ -0,0 +1,53 @@
...
+    uses: ./.github/workflows/flakybot.yaml
```
bcoe, line 52: I like that this DRYs up the config so much.

.github/workflows/workflows-matrix.json
```diff
@@ -0,0 +1,7 @@
+[
+  "functions/concepts",
+  "functions/helloworld",
```
bcoe, line 3: I'd be tempted to pull this and generate.js into a utils directory, or config directory? So that it's clear that these aren't workflows themselves, they're just used to generate workflows.

bcoe (approved):
Left a few recommendations.

package.json
kweinmeister: The workflow generation process actually happens offline in the dev environment. So, I'm not sure this will work. The user might be running a lower version of Node.

kweinmeister:
> Left a few recommendations.

Thanks for the very helpful suggestions!

PR #3133, opened 2023-04-12
chore: remove nodejs version from comment to avoid mismatch

averikitsch:


kweinmeister (approved):

iennae (approved):
LGTM; thanks Averi for going the extra mile and updating all the places the wrong information is listed in the Dockerfile.

PR #3143, opened 2023-04-15
ci: regenerate workflows with updated template

kweinmeister:
Updated existing workflows to use common template from #3079 and #3112.

In the process, added multiple workflows for samples that haven't been tested. See `.github/workflows/utils/workflows.json` changes.

There are a couple blocking tests that were skipped. There are open issues to resolve these: #3121 and #3117; and new issues created: #3144 and #3153.


iam/test/deny.test.js
```diff
@@ -23,7 +23,7 @@ const iamClient = new PoliciesClient();
 
 const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
 
-describe('IAM deny samples', () => {
+describe.skip('IAM deny samples', () => {
```
rsamborski, line 26: I have updated service account permissions and this test should pass now. Please double check.

rsamborski (approved):
LGTM with a comment re: IAM.

One more clarification - do we need to regenerate workflows with every new folder added to this repository? If yes, then we should add this information in the contributors documentation and PR template to make sure people remember about it.

iam/test/deny.test.js
kweinmeister: Thanks! Retrying now.

kweinmeister:
> LGTM with a comment re: IAM.
> 
> One more clarification - do we need to regenerate workflows with every new folder added to this repository? If yes, then we should add this information in the contributors documentation and PR template to make sure people remember about it.

No, it is not necessary to regenerate all workflows. Contributors just need to add an entry to `.github/workflows/utils/workflows.json` with the new sample directory, and then run the `generate.js` script with only the new directory they added (using the `-s` option).

iam/test/deny.test.js
kweinmeister: I tried again and still received a failure message:
```
PERMISSION_DENIED: Permission iam.googleapis.com/denypolicies.create denied on resource cloudresourcemanager.googleapis.com/projects/long-door-651
```

Will revert and we can handle this separately. It should work, thanks for trying here.

PR #3168, opened 2023-04-22
chore: remove unused dependencies

kweinmeister:


node_deps.csv
rsamborski, line 1: Probably this file shouldn't be added to the repo.

find_unused_node_deps.py
rsamborski, line 1: Do we want to add this file to the repo? If yes, then probably we should add it in a subfolder with other tools (if such already exists).

rsamborski (approved):
Approved with two comments.

PR #3297, opened 2023-06-26
fix: update publishMessage argument

### LINKED ISSUE #3296: publishMessage does not work - the argument need an update
`publishMessage({dataBuffer})` will throw an error of ‘If data is undefined, at least one attribute must be present’ because the `dataBuffer` object is being passed as a property of the first argument object. The first argument object is expected to have a data property or an attributes property that is not empty. Since the first argument object has neither of these properties, the error is thrown.

Instead, using `publishMessage({data:dataBuffer})` will not throw this error because the first argument object has a data property. The data property is expected to be a Buffer instance. If it is not a Buffer instance, then another error will be thrown: ‘Data must be in the form of a Buffer.’

Instead of
```
//https://github.com/GoogleCloudPlatform/nodejs-docs-samples/functions/ocr/app/index.js#L41-L46
const publishResult = async (topicName, data) => {
  const dataBuffer = Buffer.from(JSON.stringify(data));

  const [topic] = await pubsub.topic(topicName).get({autoCreate: true});
  topic.publishMessage({dataBuffer});
};
```
the following function works:

```
const publishResult = async (topicName, data) => {
  const dataBuffer = Buffer.from(JSON.stringify(data));

  const [topic] = await pubsub.topic(topicName).get({autoCreate: true});
  topic.publishMessage({data:dataBuffer});
};
```

ahmadradi11:
## Description

Fixes #3296

Updates the argument passed to the `publishMessage` method. 
Previously, the method was called with an object containing a `dataBuffer` property. However, the `publishMessage` method expects an object with a `data` property. This change updates the argument to pass an object with a `data` property set to the value of `dataBuffer`.

This change should fix any issues caused by passing an incorrectly formatted argument to the `publishMessage` method.

### DIFF
--- a/.github/blunderbuss.yml
+++ b/.github/blunderbuss.yml
@@ -37,6 +37,11 @@ assign_issues_by:
   - 'api: dlp'
   to:
   - GoogleCloudPlatform/googleapis-dlp
+- labels:
+  - 'api: contentwarehouse'
+  - 'api: documentai'
+  to:
+  - GoogleCloudPlatform/document-ai-samples-contributors
 
 assign_prs_by:
 - labels:
@@ -58,3 +63,8 @@ assign_prs_by:
   - 'api: dlp'
   to:
   - GoogleCloudPlatform/googleapis-dlp
+- labels:
+    - 'api: contentwarehouse'
+    - 'api: documentai'
+  to:
+    - GoogleCloudPlatform/document-ai-samples-contributors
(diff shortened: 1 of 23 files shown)


pattishin (approved):
lgtm 👍🏼  Thanks @ahmadradi11 for filing the issue and creating a PR! If you could run the linter (`npm run fix` or `gts fix`) - that would be much appreciated!

ahmadradi11:
Thanks for the approval @pattishin ! I applied the linting fixes.

PR #3325, opened 2023-07-05
chore(cleanup): retire media translate samples

### LINKED ISSUE #3323: Retire Media Translation API samples
Due to the [announced] discontinue of the Media Translation API starting July 1st, 2024, the code samples for the API, located under [ /mediatranslation][folder], have to be retired.

[announced]: https://cloud.google.com/translate/media/docs/deprecations
[folder]: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/1740047420b4c20b642d8dcb8270cde776f4a8de/mediatranslation

minherz:
### Description

Adds README with the deprecation message.
Removes tests.

Fixes #3323

### DIFF
--- a/mediatranslation/test/quickstart.test.js
+++ b/mediatranslation/test/quickstart.test.js
@@ -23,8 +23,9 @@ const cmd = 'node quickstart.js';
 const filePath = path.join(__dirname, '..', 'resources/audio.raw');
 const exec = cmd => execSync(cmd, {encoding: 'utf-8'});
 
+// skip this test because testing code sample is deprecated
 describe('Quickstart', () => {
-  it('should translate from a streamed file', async () => {
+  it.skip('should translate from a streamed file', async () => {
     const stdout = exec(`${cmd} ${filePath} linear16 en-US es-ES`);
     assert.include(stdout, 'Partial translation');
   });
--- a/mediatranslation/test/translate_from_file.test.js
+++ b/mediatranslation/test/translate_from_file.test.js
@@ -23,8 +23,9 @@ const cmd = 'node translate_from_file.js';
 const filePath = path.join(__dirname, '..', 'resources/audio.raw');
 const exec = cmd => execSync(cmd, {encoding: 'utf-8'});
 
+// skip this test because testing code sample is deprecated
 describe('MediaTranslation', () => {
-  it('should translate from a streamed file', async () => {
+  it.skip('should translate from a streamed file', async () => {
     const stdout = exec(`${cmd} ${filePath} linear16 en-US es-ES`);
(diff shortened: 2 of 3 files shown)


mediatranslation/test/quickstart.test.js
```diff
@@ -23,8 +23,9 @@ const cmd = 'node quickstart.js';
 const filePath = path.join(__dirname, '..', 'resources/audio.raw');
 const exec = cmd => execSync(cmd, {encoding: 'utf-8'});
 
+// skip this test because testing code sample is deprecated
 describe('Quickstart', () => {
-  it('should translate from a streamed file', async () => {
+  xit('should translate from a streamed file', async () => {
```
pattishin, line 28:
```suggestion
it.skip('should translate from a streamed file', async () => {
```

mediatranslation/test/translate_from_file.test.js
```diff
@@ -23,8 +23,9 @@ const cmd = 'node translate_from_file.js';
 const filePath = path.join(__dirname, '..', 'resources/audio.raw');
 const exec = cmd => execSync(cmd, {encoding: 'utf-8'});
 
+// skip this test because testing code sample is deprecated
 describe('MediaTranslation', () => {
-  it('should translate from a streamed file', async () => {
+  xit('should translate from a streamed file', async () => {
```
pattishin, line 28:
```suggestion
  it.skip('should translate from a streamed file', async () => {
```

mediatranslation/test/quickstart.test.js
pattishin: If we're no longer running these tests due to future deprecation, it would be better to remove them fully.

mediatranslation/test/translate_from_file.test.js
pattishin: Same for this.

pattishin:
@minherz If future deprecation will occur, rather than skipping the tests - removing tests fully would be best so we're not carrying too much unused code.

pattishin (changes requested):
@minherz If future deprecation will occur, rather than skipping the tests - removing tests fully would be best so we're not carrying too much unused code.

minherz:
> @minherz If future deprecation will occur, rather than skipping the tests - removing tests fully would be best so we're not carrying too much unused code.

The test(s) are disabled to avoid false positives and to reduce maintenance burden. Deleting tests only is possible but it provides small benefits because we have to keep code samples until API is fully deprecated in one year (July 1, 2024). We should be able to delete the folder after that.

Does it address your concern?

pattishin (approved):

PR #3373, opened 2023-08-01
test: update GHA flakybot upload action

kweinmeister:
FlakyBot issues aren't being uploaded regularly. I believe this is because the upload action is referencing an OS environment variable with the log, and needs to access the GHA env var.


pattishin (approved):
nice, ty Karl!

PR #3382, opened 2023-08-04
chore: update blunderbuss.yml

kweinmeister:
#3379 did not get routed to an assignee. Updating team labels and GitHub group name.


pattishin (approved):
lgtm - ty Karl!

PR #3455, opened 2023-08-15
test: avoid extra PR check on forked PRs

kweinmeister:
This is a temporary PR, and will be reverted after testing a workflow scenario.


.github/workflows/secret-manager.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
     permissions:
       contents: 'read'
       id-token: 'write'
-    if: github.event.action != 'labeled' || github.event.label.name == 'actions:force-run'
+    # Test when force-run label is applied or on scheduled branch PRs
+    if: github.event.label.name == 'actions:force-run' || (github.event.action != 'labeled' && (github.head_ref != 'main' || github.event_name == 'schedule'))
```
pattishin, line 40: May we remove the `github.head_ref` condition? More details in b/296056608.

PR #3458, opened 2023-08-15
refactor(workflows): Split workflows quickstart sample to support non-nested region tags

rogerthatdev:
This PR updates the Workflows quickstart sample with the following changes to the TS and JS variants:

- splits the `executeWorkflow` function into 2 separate functions (`executeWorkflow` and `printWorkflowResult`)
- adds region tags around code snippets to surface on [documentation page](https://cloud.google.com/workflows/docs/executing-workflow#:~:text=On%20the%20Workflows%20page%2C%20select,to%20your%20workflow%20before%20execution.)
- adds ts-node to `package.json` devdependencies to allow for `npm test` command


workflows/quickstart/index.js
```diff
@@ -22,76 +22,71 @@ const workflowName = process.argv[4] || 'myFirstWorkflow';
...
-  // Execute workflow
+async function executeWorkflow(projectId, location, workflow, runtimeArgs) {
   try {
-    const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};
     const createExecutionRes = await client.createExecution({
       parent: client.workflowPath(projectId, location, workflow),
       execution: {
-        // Provide runtime arguments as a JSON string
-        argument: JSON.stringify(runtimeArgs),
+        argument: runtimeArgs,
```
pattishin, line 44: If removing line 56 `const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};` -  Would recommend having a default/ fallback similar to this, so users can safely not use that option. 
Also, this seems to need to still be in the format of a stringified json object `{searchTerm: searchTerm}` or for your changes `{searchTerm: runtimeArgs}` now.

workflows/quickstart/index.ts
```diff
@@ -19,80 +19,76 @@ const workflowName = process.argv[4] || 'myFirstWorkflow';
...
-
-  // Execute workflow
   try {
-    const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};
     const createExecutionRes = await client.createExecution({
       parent: client.workflowPath(projectId, location, workflow),
       execution: {
-        // Provide runtime arguments as a JSON string
-        argument: JSON.stringify(runtimeArgs),
+        argument: runtimeArgs,
```
pattishin, line 46: Same as above comment.

pattishin (commented):
For b/255760794 & b/215013447, we can make developers aware of the ability to customize runtime arguments via `searchTerm` on line 22 and have more verbiage highlighting it. Otherwise, we can reveal it explicitly in the function params as you've done. 
The additional refactoring work, I would recommend to move that in another PR so that doesn't block the main target of your change.

Acknowledging that the linting error if from `main` - so that won't block your PR.

workflows/quickstart/index.js
rogerthatdev: I moved the line out of the function to enclose it in its own region tag. See line 81. User should be able to run sample with or without a search term and it executes successfully

rogerthatdev:
I'll break this up into multiple pull requests

PR #3459, opened 2023-08-15
test: fix env vars in scheduled runs 

kweinmeister:
Some scheduled runs ([example](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/actions/runs/5773416880)) are showing an error because the environment variable has already been defined. Restructuring the workflow to avoid this issue.


kweinmeister:
Since this change is affecting a workflow file, CI is linting the whole repo. It looks like we have a few unfixed issues that are unrelated that we can tackle outside of this PR.

pattishin (approved):
lgtm 👍🏼

PR #3461, opened 2023-08-17
refactor(workflows): change how runtimeArgs are defined in workflows quickstart

rogerthatdev:
This PR changes the workflows quickstart in the following ways:

* Defines the runtimeArgs variable outside of the original executeWorkflow function
* adds the `runtimeArgs` parameter to the executeWorkflow function

This is a small part of the original PR #3458

### DIFF
--- a/workflows/quickstart/index.js
+++ b/workflows/quickstart/index.js
@@ -50,16 +50,16 @@ async function executeWorkflow(projectId, location, workflow, searchTerm = '') {
       setTimeout(resolve, ms);
     });
   }
-  // Runtime arguments can be passed as a JSON string
   const runtimeArgs = searchTerm
-    ? JSON.stringify({searchTerm: searchTerm})
-    : '{}';
+    ? {searchTerm: searchTerm}
+    : {};
   // Execute workflow
   try {
     const createExecutionRes = await client.createExecution({
       parent: client.workflowPath(projectId, location, workflow),
       execution: {
-        argument: runtimeArgs,
+        // Runtime arguments can be passed as a JSON string
+        argument: JSON.stringify(runtimeArgs),
       },
     });
     const executionName = createExecutionRes[0].name;
--- a/workflows/quickstart/index.js
+++ b/workflows/quickstart/index.js
@@ -38,9 +38,9 @@ const client = new ExecutionsClient();
  * @param {string} projectId The Google Cloud Project containing the workflow
  * @param {string} location The workflow location
  * @param {string} workflow The workflow name
- * @param {string} searchTerm Optional search term to pass as runtime argument to Workflow
+ * @param {string} runtimeArgs Optional search term to pass as runtime argument to Workflow
  */
-async function executeWorkflow(projectId, location, workflow) {
+async function executeWorkflow(projectId, location, workflow, runtimeArgs) {
   /**
    * Sleeps the process N number of milliseconds.
    * @param {Number} ms The number of milliseconds to sleep.
@@ -53,12 +53,11 @@ async function executeWorkflow(projectId, location, workflow) {
 
   // Execute workflow
   try {
-    const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};
     const createExecutionRes = await client.createExecution({
       parent: client.workflowPath(projectId, location, workflow),
       execution: {
         // Provide runtime arguments as a JSON string
-        argument: JSON.stringify(runtimeArgs),
+        argument: runtimeArgs,
       },
     });
     const executionName = createExecutionRes[0].name;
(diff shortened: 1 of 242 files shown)


workflows/quickstart/index.js
```diff
@@ -38,9 +38,9 @@ const client = new ExecutionsClient();
  * @param {string} projectId The Google Cloud Project containing the workflow
  * @param {string} location The workflow location
  * @param {string} workflow The workflow name
- * @param {string} searchTerm Optional search term to pass as runtime argument to Workflow
+ * @param {string} runtimeArgs Runtime arguments to pass to the Workflow as JSON string
```
pattishin, line 41: It's still optional. We should be assigning a default if no `runtimeArgs` is passed.

workflows/quickstart/index.js
```diff
@@ -53,12 +53,10 @@ async function executeWorkflow(projectId, location, workflow) {
 
   // Execute workflow
   try {
-    const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};
     const createExecutionRes = await client.createExecution({
       parent: client.workflowPath(projectId, location, workflow),
       execution: {
-        // Provide runtime arguments as a JSON string
-        argument: JSON.stringify(runtimeArgs),
+        argument: runtimeArgs,
```
pattishin, line 59: Previous [comment](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3458/files#r1297609109
) still stands. It's looking for a jsonified object with property `searchTerm`. If you want to use what you added in line 94, do move that into the function.

workflows/quickstart/index.js
```diff
@@ -90,8 +88,16 @@ async function executeWorkflow(projectId, location, workflow) {
   }
 }
 
-executeWorkflow(projectId, location, workflowName).catch(err => {
+// [START workflows_api_quickstart_runtime_args]
+// Provide runtime arguments as a JSON string
+const runtimeArgs = searchTerm
+  ? JSON.stringify({searchTerm: searchTerm})
```
pattishin, line 94: Cleaner if you keep this logic within `executeWorkflow` and just have users pass in a search term string. If you do go this route, please add to the unit test a new condition.

pattishin (changes requested):

workflows/quickstart/index.js
rogerthatdev: searchTerm is defined as `null` on line 22 

`const searchTerm = process.argv[5] || null;`

It isn't optional now but the logic captured in the region tag will replace the default `null` with `'{}'` before calling the function

workflows/quickstart/index.js
rogerthatdev: i decided to put the definition of the `runtimeArgs` variable outside of the function as to not hardcode the json string with `searchTerm` into the function.  I hope that makes sense.

workflows/quickstart/index.js
rogerthatdev: ah ha I think i got it, let me know!

workflows/quickstart/index.js
rogerthatdev: undoing that!

workflows/quickstart/index.js
```diff
@@ -50,15 +50,16 @@ async function executeWorkflow(projectId, location, workflow) {
       setTimeout(resolve, ms);
     });
   }
-
+  // Runtime arguments can be passed as a JSON string
+  const runtimeArgs = searchTerm
+    ? JSON.stringify({searchTerm: searchTerm})
+    : '{}';
```
pattishin, line 56: Instead of hard-coding the object to a string, cleaner if you keep the original format `const runtimeArgs = searchTerm ? {searchTerm: searchTerm} : {};` and allow `JSON.stringify` do that work for you.

workflows/quickstart/index.ts
```diff
@@ -35,12 +35,13 @@ const client: ExecutionsClient = new ExecutionsClient();
...
  * @param {string} workflow The workflow name
- * @param {string} searchTerm Optional search term to pass as runtime argument to Workflow
+ * @param {string} searchTerm Optional search term to pass to the Workflow as a runtime argument
  */
 async function executeWorkflow(
   projectId: string,
   location: string,
-  workflow: string
+  workflow: string,
+  searchTerm = ''
```
pattishin, line 44: When you compile this typescript, this should be throwing a typing error at you since you are haven't given a type to `searchTerm` yet.

pattishin (changes requested):

workflows/quickstart/index.ts
pattishin: Would like to highlight this again. Please take a gander at the typing for the function. ☝🏼

pattishin (changes requested):

workflows/quickstart/index.ts
rogerthatdev: I changed this but i got clobbered

pattishin: Good to know 👍🏼  - if you can add back in `:string` and set the default manually, that should not get undone by owlbot.

pattishin (approved):

PR #3463, opened 2023-08-17
chore(workflows): remove redundant variable definitions

rogerthatdev:
This PR cleans up some comments that instruct the developer to uncomment a couple of lines to define variables. Uncommenting these lines will cause the sample to fail.

This is a small part of the original PR #3458


pattishin (commented):
These can be left alone. This is a common practice to have as a comment for our samples.

PR #3467, opened 2023-08-18
chore: add region tags to workflows quickstart

rogerthatdev:
This PR adds 2 region tags to the Workflows quickstart sample for planned updates to documentation [here](https://cloud.google.com/workflows/docs/executing-workflow).


pattishin (changes requested):
This has a lot of the same work as https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3461 - do you mind just adding the region tags to that one and close this one out?

PR #3479, opened 2023-08-30
feat(compute): add sample and test for local SSD disk

Sita04:
Sample to create disk with local SSD


compute/createInstanceWithLocalSSD.js
```diff
@@ -0,0 +1,183 @@
...
+  //    boot disk of an instance.
+  //
+  //    sourceImage: source image to use when creating this disk.
+  //    You must have read access to this disk. This can be one of the publicly available images
+  //    or an image from one of your projects.
+  //    This value uses the following format: "projects/{project_name}/global/images/{image_name}"
+  //
+  //    autoDelete: boolean flag indicating whether this disk should be deleted
+  //    with the VM that uses it.
+  async function createDiskFromImage(
```
alexander-fenster, line 88: Why would this and other functions be `async`? They don't have anything asynchronous in them, so they should be synchronous and used without `await`.

Sita04: Thanks for the comment.

I've removed async from a couple of struct type functions. The other function to create an instance does require async.

pattishin:
Hi @Sita04! When you get a chance, do you mind checking out the compute test failure? Will get this merged in once that is resolved.

pattishin (approved):
Awesome, lgtm! :shipit:

PR #3500, opened 2023-09-12
chore(discoveryengine): update search sample to disable auto paginate

chenlei1216:
## Description

By default NodeJS has autoPaginate turned on, this will result in us iterate through all pages of search results, instead of just a single page restricted by pageSize.

## Checklist

### DIFF
--- a/discoveryengine/search.js
+++ b/discoveryengine/search.js
@@ -65,7 +65,14 @@ async function main(
 
     // Perform search request
     const response = await client.search(request, {
-      // Should disable autoPaginate to avoid iterate through all pages.
+      // Warning: Should always disable autoPaginate to avoid iterate through all pages.
+      //
+      // By default NodeJS SDK returns an iterable where you can iterate through all
+      // search results instead of only the limited number of results requested on
+      // pageSize, by sending multiple sequential search requests page-by-page while
+      // iterating, until it exhausts all the search results. This will be unexpected and
+      // may cause high Search API usage and long wait time, especially when the matched
+      // document numbers are huge.
       autoPaginate: false,
     });
     const results = response[IResponseParams.ISearchResponse].results;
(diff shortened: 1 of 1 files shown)


discoveryengine/search.js
```diff
@@ -57,10 +57,20 @@ async function main(
       servingConfig: name,
     };
 
+    const IResponseParams = {
+      ISearchResult: 0,
+      ISearchRequest: 1,
+      ISearchResponse: 2,
+    };
```
holtskinner, line 64: Is there a built in enum in the client library for this?

If so, use it instead.

holtskinner:
General Comment - I don't think it makes sense to change the default setting for this to not paginate, because there could be a large number of search results, causing the response to be large/time consuming. Maybe add a comment for this, or make it a parameter to the sample function to show how to change this.

holtskinner (changes requested):

discoveryengine/search.js
chenlei1216: I can't find any. Will have to go with this first and update later if I can find better options.

chenlei1216:
> General Comment - I don't think it makes sense to change the default setting for this to not paginate, because there could be a large number of search results, causing the response to be large/time consuming. Maybe add a comment for this, or make it a parameter to the sample function to show how to change this.

To clarify, we are still doing the pagination (pageSize is still being used). This is a special behavior for NodeJS that by default it's turning the API call result into an iterable of SearchResults by default (while for Python etc. it's only returning SearchResponse). This may cause several issues for our case, e.g. if users set page size as 10, and there are 100 results, NodeJS library by default will try to iterate through all 100 results by sending 10 search requests in sequence (offset=0, offset=10, offset=20...) until it exhausts the full list. If you have 1K, 10K search results this could take forever.

IMO we should always turn off the autoPaginate feature for the search use case, so we can get the single SearchResponse just for the page, in the above example it'll only return 10 documents. Hope this makes sense.

holtskinner (approved):

pattishin (approved):

PR #3505, opened 2023-09-15
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):

* demonstrate the use of Error Reporting for reporting exceptions
* use the "fixed" error message
* delegate error context processing to Error Reporting backend

### DIFF
--- a/error-reporting/quickstart.js
+++ b/error-reporting/quickstart.js
@@ -14,16 +14,25 @@
 
 'use strict';
 
-function quickstart() {
-  // [START error_reporting_quickstart]
-  // Imports the Google Cloud client library
-  const {ErrorReporting} = require('@google-cloud/error-reporting');
+// [START error_reporting_quickstart]
+// Import the Google Cloud client library
+const {ErrorReporting} = require('@google-cloud/error-reporting');
 
+function quickstart() {
+  try {
+    throw new Error('Something went wrong');
+  } catch (exception) {
+    reportError(exception);
+  }
+}
+
+function reportError(exception) {
   // Instantiates a client
   const errors = new ErrorReporting();
 
-  // Reports a simple error
-  errors.report('Something broke!');
-  // [END error_reporting_quickstart]
+  // Reports an exception
+  errors.report(exception.stack);
 }
+// [END error_reporting_quickstart]
+
 quickstart();


error-reporting/quickstart.js
```diff
@@ -14,16 +14,25 @@
...
+const {ErrorReporting} = require('@google-cloud/error-reporting');
+
 function quickstart() {
-  // [START error_reporting_quickstart]
-  // Imports the Google Cloud client library
-  const {ErrorReporting} = require('@google-cloud/error-reporting');
+  try {
+    throw new Error('Something went wrong');
+  } catch (exception) {
+    reportError(exception);
```
NimJay, line 25: **Praise:** I like the simplicity and readability of your code here.

error-reporting/quickstart.js
```diff
@@ -14,16 +14,25 @@
...
 
+function reportError(exception) {
   // Instantiates a client
   const errors = new ErrorReporting();
 
-  // Reports a simple error
-  errors.report('Something broke!');
-  // [END error_reporting_quickstart]
+  // Reports an exception
+  errors.report(exception.stack);
```
NimJay, line 34: **Question:**
In the equivalent [C# sample](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2337/files#diff-5494443eba0482e62010d9445b529d4f511c98efeeb2c01bb512d71202798356L81) and the [Java sample](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8648/files#diff-f55c582b94d740b432da5ddc2be409846bea16d36c644b65796b8db2c12be6a4R71), we create a `ReportedErrorEvent` object, and report that to Google Cloud.
Is there a reason we don't use a `ReportedErrorEvent` object here (in our JavaScript)?

NimJay (approved):
Approved.
I left a question, but it's not a blocker.

Thanks again, @minherz, do working on these adjustments!

error-reporting/quickstart.js
minherz: This lib has handwritten method that abstracts the work developers have to do when using auto-generated libraries

PR #3509, opened 2023-09-22
fix(contentwarehouse): Update tag for create-document.js

holtskinner:
- Followup to https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3507
- Tag was copied/pasted as `quickstart` when it should have been a new tag


paulinawins:
The tag for this sample code is technically correct. I was following the python sample which simply creates a doc with the content warehouse API. https://cloud.google.com/document-warehouse/docs/samples/contentwarehouse-quickstart?hl=en

The team would prefer to keep this nodeJS sample code together with the python one and not change the tag. Instead, should I rename my file and function name?

holtskinner:
> The tag for this sample code is technically correct. I was following the python sample which simply creates a doc with the content warehouse API. https://cloud.google.com/document-warehouse/docs/samples/contentwarehouse-quickstart?hl=en
> 
> The team would prefer to keep this nodeJS sample code together with the python one and not change the tag. Instead, should I rename my file and function name?

Thanks for clarifying, I'd recommend changing the file name and function name instead to match the Python sample. I can do that in this same Pull Request and revert my tag change.

paulinawins (approved):
Thanks for updating this!

PR #3510, opened 2023-09-22
feat(storagetransfer): add event driven transfer samples and tests

JesseLovelace:
Adds event driven transfer samples and tests

### DIFF
--- a/storagetransfer/test/utils/queue.js
+++ b/storagetransfer/test/utils/queue.js
@@ -22,75 +22,81 @@ const AWS = require('aws-sdk');
 const uuid = require('uuid');
 
 class QueueManager {
-    constructor() {
-        this.pubsub = new PubSub();
-        AWS.config.update({region: 'us-west-1'});
-        this.sqs = new AWS.SQS({apiVersion: '2012-11-05'});
+  constructor() {
+    this.pubsub = new PubSub();
+    AWS.config.update({region: 'us-west-1'});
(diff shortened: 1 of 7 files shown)


storagetransfer/test/utils/queue.js
```diff
@@ -0,0 +1,102 @@
...
+      const queueName = `sts-queue-name-${uuid.v4()}`;
+      this.sqs.createQueue(
+        {QueueName: queueName},
+        (error, createQueueResult) => {
+          if (error) {
+            console.error(error);
+            return reject(error);
+          }
+          this.sqsQueues.push(createQueueResult.QueueUrl);
+          this.sqs.getQueueAttributes(
```
danielbankhead, line 77: If the SQS offers a Promise API, that would be preferred over callbacks

danielbankhead (commented):
Looking great so far, a few minor things

storagetransfer/test/utils/queue.js
danielbankhead: Never mind, I see its still callback-only https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SQS.html#createQueue-property

danielbankhead: Ah, their docs are so bad. Here's how to do promises, but we can do it later: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-async-await.html

danielbankhead (approved):

PR #3616, opened 2024-01-25
feat: re-add GenAI embedding code samples

gericdong:
## Description

Fixes #3320

Re-add code samples for GenAI embeddings. These samples were published but later removed because of the launch schedule change.

- Generate image from image
- Generate image from text


pattishin (approved):
lgtm

PR #3626, opened 2024-02-15
fix(aiplatform): update Gemini model names

gericdong:
## Description

Update Gemini model names to the GA version:
- gemini-pro > gemini-1.0-pro
- gemini-pro-vision > gemini-1.0-pro-vision


pattishin (approved):
lgtm :shipit:

PR #3659, opened 2024-04-18
chore: update Dockerfiles to use modern images

iennae:
* update to use node:20-slim
* resolves warning that comes up with npm WARN
* update comment to clarify difference between using ci (clean install) over install which isn't necessarily faster (and can be in fact slower) and the context of why production.

This PR intentionally does not resolve some other
remaining Docker issues as there are remaining
issues to resolve including appropriate fonts
with Alpine image use.

This does resolve a chunk of the Dockerfile
updates. Many of these are exactly the same
and there may be room to centralize and
standardize rather than having separate files.

I did go in and build docker images to verify these worked. 

## Checklist

- [X ] Please **merge** this PR for me once it is approved

### DIFF
--- a/cloud-sql/mysql/mysql/Dockerfile
+++ b/cloud-sql/mysql/mysql/Dockerfile
@@ -4,7 +4,7 @@
 
 # Use the official lightweight Node.js image.
 # https://hub.docker.com/_/node
-FROM node:18-slim
+FROM node:20-slim
 
 # Create and change to the app directory.
 WORKDIR /app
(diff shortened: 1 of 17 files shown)


cloud-sql/mysql/mysql/Dockerfile
```diff
@@ -15,9 +15,14 @@ WORKDIR /app
...
 # Install dependencies.
-# If you add a package-lock.json speed your build by switching to 'npm ci'.
-# RUN npm ci --only=production
-RUN npm install --production
+# if you need a deterministic and repeatable build create a 
+# package-lock.json file and use npm ci:
+# RUN npm ci --omit=dev
+# if you need to include development dependencies during development
+# of your application, use:
+# RUN npm install -dev
```
glasnt, line 23: This needs to be `--dev`. Running this as is causes the following: 

```
npm WARN config dev Please use --include=dev instead.
```

This will need to be repeated in all other instances in this PR

iennae: thanks for catching

kweinmeister (approved):

PR #3686, opened 2024-05-08
feat: Add new token counting sample

arbrown:
Also move existing sample to new subfolder


gericdong (commented):
Because we're moving an existing code sample integrated in docs, we will need to notify docs team to modify the docs.

gericdong (approved):

PR #3693, opened 2024-05-10
chore: add datastore_id secret

arbrown:


arbrown:
The datastore ID is showing up as undefined in my other CI runs, I think because it needs to be added in a separate PR before it will show up in the env for the tests that need it.

PR #3729, opened 2024-07-08
feat: added samples for regional SM

vipul7499:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


secret-manager/addRegionalSecretVersion.js
```diff
@@ -0,0 +1,55 @@
...
+   */
+  // const projectId = 'my-project';
+  // const locationId = 'location';
+  // const secretId = 'my-secret';
+
+  const parent = `projects/${projectId}/locations/${locationId}/secrets/${secretId}`;
+  // Imports the Secret Manager library
+  const {SecretManagerServiceClient} = require('@google-cloud/secret-manager');
+
+  const options = {};
```
sofisl, line 30: Can you add a comment why we need the regional endpoint for grpc?

secret-manager/test/secretmanager.test.js
```diff
@@ -109,6 +150,13 @@ describe('Secret Manager samples', () => {
     assert.match(output, new RegExp('Created secret'));
   });
 
+  it('creates a regional secret', async () => {
```
sofisl, line 153: It's not good to chain test results like this - for example here, you create ${secretId}-2 and then in line 222 you delete it. Instead all resources for a test should be hermetic, i.e., should only be created and destroyed for that single test.

secret-manager/test/secretmanager.test.js
```diff
@@ -22,12 +22,21 @@ const {SecretManagerServiceClient} = require('@google-cloud/secret-manager');
 const client = new SecretManagerServiceClient();
 
 const projectId = process.env.GCLOUD_PROJECT;
+const locationId = process.env.GCLOUD_LOCATION;
```
sofisl, line 25: both of these env variables should be changed to getting from ADC, like so: https://github.com/googleapis/google-cloud-node/blob/2359c347ca89df4aee706b0c6798560d686a15cc/packages/google-cloud-secretmanager/samples/test/quickstart.test.js#L31

Makes running tests locally easier and makes debugging faster, since it would be hard to know what env variables to set without reading the code.

sofisl (changes requested):

secret-manager/addRegionalSecretVersion.js
vipul7499: Fixed

secret-manager/test/secretmanager.test.js
vipul7499: Fixed

secret-manager/test/secretmanager.test.js
vipul7499: I am not sure how will this work, because to instantiate the client itself i will need the location id, can you please elaborate?

sofisl: You're right! I think I was thinking this mostly for the project ID. For location, I think you should set a default value, i.e., `process.env.GCLOUD_LOCATION || 'us-central1'`. This will allow tests to run and not fail randomly.

sofisl (approved):

subfuzion:
Hi @vipul7499, the CI test is failing because certain workflows can't be run from a fork. Please resubmit this pull request using a branch on this repo instead of your fork. I confirmed that your pull request will pass tests once you do this. Thanks!

secret-manager/test/secretmanager.test.js
subfuzion: @sofisl @vipul7499 Hello to both of you. I'm not the expert here, but in the interests of seeing this pull request move forward, I've got a question based on the preceding comments: the purpose here isn't clear to me, but if the goal is to move a secret from one location to another, shouldn't this be demonstrated using [replication](https://cloud.google.com/secret-manager/docs/choosing-replication)?

vipul7499: @subfuzion The goal is not to move a secret from one location to another, it is to create and save a secret in a specific location and make secret manager DRZ complaint. 

I have not merged this PR because there is some concerns regarding this, the thread of which is going on Gchat and other languages PR, once that are resolved, i will be merging this.

subfuzion: @vipul7499 Ah, makes sense now. Thanks for the explanation. I'm very interested in hearing the final resolution of your discussions. I'm going to place this in draft mode for now. Feel free to let me know if there's anything I can do to help move this forward when you're ready.

PR #3793, opened 2024-08-13
feat: create_batch_custom_network

gryczj:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist
- [x] I have followed guidelines from [CONTRIBUTING.MD](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md) and [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/)
- [x] **Tests** pass:   `npm test` (see [Testing](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#run-the-tests-for-a-single-sample))
- [x] **Lint** pass:   `npm run lint` (see [Style](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#style))
- [x] This pull request is from a branch created directly off of `GoogleCloudPlatform/nodejs-docs-samples`. Not a fork.
- [x] Please **merge** this PR for me once it is approved


batch/create/create_batch_custom_network.js
```diff
@@ -0,0 +1,134 @@
...
+      jobId: jobName,
+      job,
+    };
+
+    // Run request
+    const [response] = await batchClient.createJob(request);
+    console.log(JSON.stringify(response));
+  }
+
+  callCreateBatchCustomNetwork();
```
subfuzion, line 127: Missing `await`.

batch/test/create_batch_custom_network.test.js
```diff
@@ -0,0 +1,61 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const path = require('path');
+const assert = require('assert');
```
subfuzion, line 20: Minor nit (we haven't updated guidance yet), but I recommend:

```javascript
const assert = require('node:assert/strict');
```

subfuzion (changes requested):
Minor changes requested. Thanks!

batch/create/create_batch_custom_network.js
```diff
@@ -0,0 +1,134 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+async function main() {
+  // [START batch_create_custom_network]
+  // Imports the Batch library
+  const batchLib = require('@google-cloud/batch');
+  const batch = batchLib.protos.google.cloud.batch.v1;
```
msampathkumar, line 23: Hi, please help me connect the dot. I that this code is supposed to similar to https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/batch/snippets/src/main/java/com/example/batch/CreateBatchCustomNetwork.java#L17-L149 

I am not an expert but I feel there are changes b/w this .java and .js file.

gryczj: Could you explain in more detail what changes you mean? I followed this .java file and also the documentation-https://cloud.google.com/batch/docs/job-without-external-access when I was creating this sample. 

//
I changed value for jobName to `**example-job** as it was in previous PR, because **JOB_NAME**- as in .java doesn't follow the required pattern.

batch/test/create_batch_custom_network.test.js
```diff
@@ -0,0 +1,61 @@
...
+  const jobName = 'example-job';
+  const region = 'europe-central2';
+  let projectId;
+
+  before(async () => {
+    projectId = await batchClient.getProjectId();
+  });
+
+  after(async () => {
+    await deleteJob(batchClient, projectId, region, jobName);
```
subfuzion, line 39: Hi @gryczj I know it's not part of your PR, but when I examined the code you're calling for disposing of the resource that's created, I see that the exception is swallowed. The code is simple (`await batchClient.deleteJob(request);`) so for clarity, it probably just ought to be here. In any case, the code should throw so that this after hook fails.

batch/test/create_batch_custom_network.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+const path = require('path');
+const assert = require('node:assert/strict');
+const {describe, it} = require('mocha');
+const cp = require('child_process');
+const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
+const cwd = path.join(__dirname, '..');
+const {BatchServiceClient} = require('@google-cloud/batch').v1;
+const {deleteJob} = require('./batchClient_operations');
+const batchClient = new BatchServiceClient();
```
subfuzion, line 27: The `cwd` and `batchClient` don't belong in the `require` package group. Also, not a hard rule, but typically when the group is more than a few packages, it's nice to group them by external dependencies first, then relative dependencies.

subfuzion (changes requested):
Hi @gryczj Looking good, just a few more nits. Thanks.

batch/create/create_batch_custom_network.js
subfuzion: @msampathkumar Are you ok with resolving this conversation?

batch/test/create_batch_custom_network.test.js
gryczj: I reordered imports in every file in **batch/tests** this way:
`
const path = require('path');
const assert = require('node:assert/strict');
const {describe, it} = require('mocha');
const {BatchServiceClient} = require('@google-cloud/batch').v1;
const {deleteJob} = require('./batchClient_operations');

const cp = require('child_process');
const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
const batchClient = new BatchServiceClient();

`

batch/test/create_batch_custom_network.test.js
subfuzion: No need to create a new error, you can just throw the existing one (`throw err`) to ensure the error in cleaning up gets surfaced up to the test runner. I hate to say this, but I'm not in favor of the filename either: `batchClient_operations` mixes camelcase and snakecase; `cleanup_operations` or `cleanup_helpers` might be better.

But honestly, if you just add `await batchClient.deleteJob(request);` to your `after` hooks, things would just be so much simpler. It makes it clear what you're doing and if it throws it will get handled and it will be clear that it happened in the after handler, so really no need to even catch it just to log an error string. I don't really see any particular value in breaking this out into a separate module.

subfuzion (changes requested):
Still a bit of feedback, but the rest of the PR looks good. You're almost there! Thanks! :)

batch/test/create_batch_custom_network.test.js
subfuzion: Nice, thank you!

batch/test/create_batch_custom_network.test.js
gryczj: changes applied also to other batch samples

subfuzion (approved):

PR #3853, opened 2024-09-14
chore(run/websockets): temporarily disable run/websockets on kokoro

subfuzion:
Internal: b/366619034


iennae (approved):
adding additional context in the review

Migrating the system test to system-test-fix as it won't run but leaves it in the file. new system-test when run will just echo that it's disabled. to "fix" update system-test-fix to system-test and delete the system-test.

PR #3951, opened 2025-01-15
feat: test-isolation package setup configs

davidcavazos:
## Description

Supports package setup configuration files. These allow packages to provide per-package configurations like secrets, environment variables, and the runtime version.

## Checklist

### DIFF
--- a/.github/cloud-samples-tools/pkg/config/config.go
+++ b/.github/cloud-samples-tools/pkg/config/config.go
@@ -72,7 +72,8 @@ func (c *Config) Save(file *os.File) error {
 // LoadConfig loads the config from the given path.
 func LoadConfig(path string) (*Config, error) {
 	// Read the config file.
-	config, err := ReadJsonc[Config](path, nil)
+	var config Config
+	err := ReadJsonc(path, &config)
 	if err != nil {
 		return nil, err
 	}
@@ -85,7 +86,7 @@ func LoadConfig(path string) (*Config, error) {
 		config.Match = []string{"*"}
 	}
 
-	return config, nil
+	return &config, nil
 }
 
 // Match returns true if the path matches any of the patterns.
@@ -168,14 +169,15 @@ func (c *Config) Affected(log io.Writer, diffs []string) ([]Package, error) {
 
 	packages := make([]Package, 0, len(changed))
 	for _, path := range changed {
-		pkg := Package{Path: path}
+		pkg := Package{Path: path, Setup: &c.Setup.Defaults}
 		if c.Setup.FileName != "" {
 			setup_file := filepath.Join(path, c.Setup.FileName)
-			setup, err := ReadJsonc(setup_file, &c.Setup.Defaults)
-			if err != nil {
-				return nil, err
+			if fileExists(setup_file) {
+				err := ReadJsonc(setup_file, pkg.Setup)
+				if err != nil {
+					return nil, err
+				}
 			}
-			pkg.Setup = setup
 		}
 		packages = append(packages, pkg)
 	}
--- a/.github/cloud-samples-tools/pkg/config/config.go
+++ b/.github/cloud-samples-tools/pkg/config/config.go
@@ -24,7 +24,6 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"regexp"
 	"slices"
 	"strings"
 )
@@ -32,6 +31,9 @@ import (
 type Config struct {
 	// Filename to look for the root of a package.
 	PackageFile []string `json:"package-file"`
+
+	// Setup filename, must be located in the same directory as the package file.
+	SetupFile string `json:"setup-file"`
 
 	// Pattern to match filenames or directories.
 	Match []string `json:"match"`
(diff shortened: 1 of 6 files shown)


.github/cloud-samples-tools/pkg/config/config.go
```diff
@@ -61,30 +63,22 @@ func (c *Config) Save(file *os.File) error {
...
 	if err != nil {
 		return nil, err
 	}
 
-	// Set default values if they are not set.
+	// Validate for required values.
 	if config.PackageFile == nil {
 		return nil, errors.New("package-file is required")
 	}
-	if config.Match == nil {
```
briandorsey, line 84: I *think* nil vs. empty slices are technically different in Go. [Go Playground link](https://go.dev/play/p/FyKK2w8D4U0).  And the Json parser could return either depending on what is in the underlying file. 

Consider `if len(config.Match) == 0 {`, which covers both cases. (If changes, this pattern probably needs updates in multiple places. Also might make sense to do as a followup PR. 

This is not an area where I'm an expert. Calling @telpirion for an additional look.

.github/cloud-samples-tools/cmd/main.go
```diff
@@ -29,8 +29,8 @@ import (
 var usage = `usage: tools <command> ...
 
 commands:
-  affected path/to/config.jsonc path/to/diffs.txt
-  run-all path/to/config.jsonc path/to/script.sh
+  affected path/to/config.jsonc diffs.txt paths.txt
+  setups path/to/config.jsonc paths.txt
```
telpirion, line 33: nit: `setups` sounds odd as a command in this context. Maybe `set-up`? Or something else?

.github/cloud-samples-tools/cmd/main.go
```diff
@@ -48,21 +48,31 @@ func main() {
...
+		setupsCmd(configFile, pathsFile)
 
 	default:
 		log.Fatalln("❌ unknown command: ", command, "\n", usage)
 	}
 }
 
 // affected command entry point to validate inputs.
-func affectedCmd(configFile string, diffsFile string) {
+func affectedCmd(configFile string, diffsFile string, pathsFile string) {
```
telpirion, line 75: nit: godoc style is to use the exact function name as the very first string in the comment.

Suggestion:

```go
// affectedCmd validates inputs passed into the system
```

.github/cloud-samples-tools/cmd/main.go
```diff
@@ -76,23 +86,58 @@ func affectedCmd(configFile string, diffsFile string) {
...
+	output, err := json.Marshal(paths)
 	if err != nil {
-		log.Fatalln("❌ error marshaling packages to JSON.\n", err)
+		log.Fatalln("❌ error marshaling paths to JSON.\n", err)
 	}
+	fmt.Println(string(output))
+}
 
-	fmt.Println(string(packagesJson))
+// setups command entry point to validate inputs.
```
telpirion, line 119: Same as previous: first string in comment should be the exact function name, `setupsCmd`.

.github/cloud-samples-tools/pkg/config/config.go
```diff
@@ -61,30 +63,22 @@ func (c *Config) Save(file *os.File) error {
...
+		Match: []string{"*"},
 	}
 
-	// Strip the comments and load the JSON.
-	sourceJson := multiLineCommentsRegex.ReplaceAll(sourceJsonc, []byte{})
-	sourceJson = singleLineCommentsRegex.ReplaceAll(sourceJson, []byte{})
-
-	var config Config
-	err = json.Unmarshal(sourceJson, &config)
+	// This mutates `config` so there's no need to reassign it.
```
telpirion, line 71: praise: I love the comments here that help me interpret what the code is doing.

briandorsey (approved):
Very high level review without running it myself, but overall LGTM.

.github/cloud-samples-tools/pkg/config/utils.go
```diff
@@ -17,10 +17,28 @@
 package config
 
 import (
+	"encoding/json"
 	"errors"
 	"os"
+	"regexp"
 )
 
+var multiLineCommentsRegex = regexp.MustCompile(`(?s)\s*/\*.*?\*/`)
```
telpirion, line 26: nit: Go style prefers shorter variable names, relying on the context for additional information. 

I would name this just `multiLine` and the next one `singleLine`.

telpirion (approved):
Only nits. Looks good!

.github/cloud-samples-tools/pkg/config/config.go
davidcavazos: Interesting, yet another pitfall. I did change the way I passed default values to the JSON parser, so this is no longer the case. I found a simpler way.

telpirion: That is a good catch! I agree, the `len()` function should be used here.

.github/cloud-samples-tools/cmd/main.go
davidcavazos: How about `find-setups` or `setup-files`?

.github/cloud-samples-tools/cmd/main.go
davidcavazos: Good catch, that came from a refactor and forgot to update the godoc.

PR #3986, opened 2025-02-25
ci(appengine): add metadata/standard to testing isolation

glasnt:
Migrates appengine/metadata/standard to testing isolation 


appengine/metadata/standard/ci-setup.json
```diff
@@ -0,0 +1,3 @@
+{
```
davidcavazos, line 1: Node 20 is already the default, so we can remove this ci-setup.json file altogether :)

davidcavazos (approved):

appengine/metadata/standard/ci-setup.json
glasnt: removed

PR #4037, opened 2025-03-24
fix: ensure all retry.sh failures fail exit 1

glasnt:
## Description

Identified in https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3985/files#diff-86f605e760c2507e13cf901bc344431a78deb5dbd16c6e94726a76affe9a9bb1R62

Ensures all copies of `retry.sh` use the correct exit code to correctly fail when retries exceeded. 

- [x] Please **merge** this PR for me once it is approved


iennae (approved):
noticing linting is failing for unchanged files; is that intentional behavior?

davidcavazos:
It's running lint on the entire changed packages. Looks like both `run/markdown-preview/editor` and `run/markdown-preview/renderer` were affected, so the linter ran on both of those packages. It seems like those packages had linter issues to begin with.

The previous linter ran gts lint fix first and then ran the linter, so all the automatically fixable lint issues were hidden. So a lot of packages actually have lint issues, it's just that the previous linter basically ignored them.

PR #4085, opened 2025-05-08
fix(functions): update and fix `functions_billing_stop` sample

eapl-gemugami:
## Description

Fixes b/403855924

This sample for a Cloud Run Function used in [disable-billing-with-notifications#create-cloud-run-function](https://cloud.google.com/billing/docs/how-to/disable-billing-with-notifications#create-cloud-run-function) is updated and validated to current Node 20 requirements.

As the original `index.js` file has several samples in the same file, this one will me moved to new folder, to apply the [One sample per file](https://googlecloudplatform.github.io/samples-style-guide/#one-per-file) principle.

## Checklist
- [x] I have followed guidelines from [CONTRIBUTING.MD](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md) and [Samples Style Guide](https://googlecloudplatform.github.io/samples-style-guide/)
- [x] **Tests** pass:   `npm test` (see [Testing](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#run-the-tests-for-a-single-sample))
- [x] **Lint** pass:   `npm run lint` (see [Style](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#style))
- [x] This pull request is from a branch created directly off of `GoogleCloudPlatform/nodejs-docs-samples`. Not a fork.
- [ ] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CODEOWNERS) with the codeowners for this sample <- It creates a subdirectory, I don't think this is required.
- [ ] This sample adds a new sample directory, and I created [GitHub Actions workflow](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/CONTRIBUTING.md#adding-new-samples) for this sample <- Same as above


functions/billing/stop_billing/index.js
```diff
@@ -0,0 +1,110 @@
+// Copyright 2025 Google LLC
```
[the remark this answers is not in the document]
eapl-gemugami: What's the current year ?

functions/billing/stop_billing/index.js
```diff
@@ -0,0 +1,110 @@
...
+    console.log('Billing disabled. (Simulated)');
+
+    /*
+    const [response] = await billing.updateProjectBillingInfo({
+      name: projectName,
+      resource: body, // Disable billing
+    });
+
+    console.log(`Billing disabled: ${JSON.stringify(response)}`);
+    */
```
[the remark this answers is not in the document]
eapl-gemugami: This will be fixed in the final PR. This is left to show how the sample is currently tested in the Cloud Run Environment.

iennae:
@davidcavazos if you could also take a look at this one; thanks!

eapl-gemugami:
@davidcavazos for this sample I'm planning to add a Unit Test similar to that on Python [stop_billing_test.py](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/6a8655fef0d4274b6f0174ffaeffe457981aeb4b/functions/billing/stop_billing/stop_billing_test.py)

The current one is just a copy from the test in the parent folder.

eapl-gemugami:
@davidcavazos any comments on this approach of making an Unit Test based on the Python sample? Thanks!

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+  );
+
+  if (eventObject.costAmount <= eventObject.budgetAmount) {
+    console.log('No action required. Current cost is within budget.');
+    return;
+  }
+
+  console.log(`Disabling billing for project '${projectName}'...`);
+
+  const billingEnabled = _isBillingEnabled(projectName);
```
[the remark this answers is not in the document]
eapl-gemugami: You suggested the opposite in [this comment](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/4085?notification_referrer_id=NT_kwDOA9IL-LQxNjMwMTc5MDE4Nzo2NDA5NzI3Mg#discussion_r2229288383)

As it's an async function I'll revert the `await`

functions/billing/stop-billing/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "rules": {
+    "node/no-missing-import": ["off"],
+    "node/no-missing-require": ["off"],
+    "node/no-unpublished-import": ["off"],
+    "node/no-unpublished-require": ["off"],
+    "node/no-unsupported-features/es-syntax": ["off"]
+  },
+  "parserOptions": {
+    "ecmaVersion": 2020
```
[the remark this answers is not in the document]
eapl-gemugami: 2020 is fine for compatibility

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+    eventObject = JSON.parse(eventData);
+  } catch (e) {
+    console.error('Error parsing event data:', e);
+    return;
+  }
+
+  console.log(
+    `Project ID: ${projectId} ` +
+      `Current cost: ${eventObject.costAmount} ` +
+      `Budget: ${eventObject.budgetAmount}`
```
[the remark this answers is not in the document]
eapl-gemugami: I don't see the benefit here for improved readability, while it's going against the [80 characters limit](https://google.github.io/styleguide/jsguide.html#formatting-column-limit)

functions/billing/stop-billing/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "rules": {
+    "node/no-missing-import": ["off"],
+    "node/no-missing-require": ["off"],
+    "node/no-unpublished-import": ["off"],
+    "node/no-unpublished-require": ["off"],
+    "node/no-unsupported-features/es-syntax": ["off"]
```
glasnt: These rules are covered by the top level ~/.eslintrc.json and may not need to be added here. This entire file may not be required.

functions/billing/stop-billing/.gcloudignore
```diff
@@ -0,0 +1,16 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
```
glasnt: The tutorial directs the user to use the web interface to create the Cloud Run function, so this file is not required. It was probably added during local testing.

functions/billing/stop-billing/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "0.0.1",
+  "description": "Disable billing with a budget notification.",
+  "main": "index.js",
+  "engines": {
+    "node": ">=20.0.0"
+  },
+  "scripts": {
+    "test": "c8 mocha -p -j 2 test/index.test.js --timeout=5000 --exit"
+  },
+  "author": "Emmanuel Parada <paradalicea@google.com>",
```
glasnt: Author for these samples must be "Google Inc." (see other package.json files for reference)

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+const billing = new CloudBillingClient();
+
+const projectIdEnv = process.env.GOOGLE_CLOUD_PROJECT;
+
+functions.cloudEvent('StopBillingCloudEvent', async cloudEvent => {
+  // TODO(developer): As stopping billing is a destructive action
+  // for your project, change the following constant to `false`
+  // after you validate with a test budget.
+  const simulateDeactivation = true;
```
glasnt: This is not something the Python sample does (unless there's a pending change there to match). 

Also, I would not use the term "simulateDeactivation". Instead I'd use something more forceful like "dry-run". That is a term more associated to actions that don't invoke the processing required. 

You should also remove any output later that appends "(simulated)" and make it more obvious that it's a dry-run of the function. I can imagine support requests given the seriousness of this sample, and want to make sure it's as obvious as possible what's going on.

functions/billing/stop-billing/test/index.test.js
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    it('should receive a notification exceeding the budget and simulate stopping billing', async () => {
+      const StopBillingCloudEvent = getFunction('StopBillingCloudEvent');
+      const isOverBudget = true;
+      await StopBillingCloudEvent(getCloudEventOverBudgetAlert(isOverBudget));
+
+      assert.ok(consoleOutput.includes('Getting billing info'));
+      assert.ok(consoleOutput.includes('Disabling billing for project'));
+      assert.ok(consoleOutput.includes('Billing disabled. (Simulated)'));
```
glasnt: Note to update console check here if the output is updated.

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+  let projectId = projectIdEnv;
+
+  if (projectId === undefined) {
+    console.log(
+      'Project ID not found in env variables. Getting GCP metadata...'
+    );
+    try {
+      projectId = await gcpMetadata.project('project-id');
```
glasnt: You should confirm if Cloud Run functions get the default project ID as a variable. If it is, then you don't need this extra check.

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+  // for your project, change the following constant to `false`
+  // after you validate with a test budget.
+  const simulateDeactivation = true;
+
+  let projectId = projectIdEnv;
+
+  if (projectId === undefined) {
+    console.log(
+      'Project ID not found in env variables. Getting GCP metadata...'
+    );
```
glasnt: Depending on if there's a default envvar for project, you will potentially always get this output. I would remove this extra output, and add a comment if the project is pulled from the metadata, and update the error message if the envvar wasn't set and it couldn't be pulled from metadata (this would probably indicate the command wasn't run in an authenticated environment). (Changing this may affect tests)

glasnt (dismissed):
It's unclear if this is the companion to https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/functions/billing_stop_on_notification; if so, there may be more changes required (same folder name functionality suggestions). Some other changes requested.

functions/billing/stop-billing/.eslintrc.json
eapl-gemugami: I agree about these lines, I checked again and can confirm they are not needed.

But this line is required to pass linting:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/b279adf494ac25907d7a98616d6b4384bab51ece/functions/billing/stop-billing/.eslintrc.json#L10

...to support this syntax (compatible with Node 18+)
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/b279adf494ac25907d7a98616d6b4384bab51ece/functions/billing/stop-billing/index.js#L48

I propose leaving only the `"ecmaVersion": 2020` if it is OK for you.

functions/billing/stop-billing/package.json
eapl-gemugami: Ok, thanks! I was using the previous sample as a reference:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/b279adf494ac25907d7a98616d6b4384bab51ece/functions/billing/package.json#L14

I'll use this one then:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/b279adf494ac25907d7a98616d6b4384bab51ece/package.json#L5

functions/billing/stop-billing/index.js
eapl-gemugami: Please watch my updated sample here: [stop_billing.py#L70-L73](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/functions/billing_stop_on_notification/stop_billing.py#L70-L73)

I agree with the term `dry-run`. We could update the Node sample first, and replicate the change on Python afterwards.

About "make it more obvious that it's a dry-run of the function", do you have some reference? I've seen the dry-run for Let's Encrypt, Apache, etc, but I don't know of a case for GCP on a Client Library.

I've found for example: https://cloud.google.com/resource-manager/docs/organization-policy/dry-run-policy

functions/billing/stop-billing/index.js
eapl-gemugami: I can't see how "if there's a default envvar for project, you will potentially always get this output"
```
const projectIdEnv = process.env.GOOGLE_CLOUD_PROJECT;
let projectId = projectIdEnv;

if (projectId === undefined) {
  console.log(
    'Project ID not found in env variables. Getting GCP metadata...'
  );
```

But I think this syntax could be quite simplified, at least to the sample to be shown to the developer.

To give some context, this sample is designed to be uploaded as a Cloud Run Function (please watch https://cloud.google.com/billing/docs/how-to/disable-billing-with-notifications#create-cloud-run-function) 

This validation in lines 32-42 was intended to support both local tests (supplying an env var `GOOGLE_CLOUD_PROJECT`), and as the normal Cloud Function (getting it from the Metadata server instead as in other samples)

I think getting the project from an envvar could be done outside the sample to simplify it a bit, and only show inside the sample how to get it from Metadata. What do you think?

eapl-gemugami:
@glasnt thanks a lot for your review! I'll work on those comments I have given a thumbs up to, and I have a few questions in the replies.

functions/billing/stop-billing/index.js
eapl-gemugami: I couldn't find an example in the Node repo.

Based on Let's encrypt output, I'm proposing:
```js
console.log('** DRY RUN: simulating billing deactivation');
console.log('Billing disabled.');
```

Please let me know if there's a more appropriate syntax or wording for our samples.

functions/billing/stop-billing/index.js
eapl-gemugami: Do you mean as an environment variable?

Based on my latest research and tests deploying the function, Cloud Run doesn't supply the Project ID as an envvar anymore.
https://cloud.google.com/run/docs/container-contract#services-env-vars

So it now depends on the Metadata. I found some examples [in the repo](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fnodejs-docs-samples+path%3A%2F%5Erun%5C%2F%2F+projectId&type=code) but they rely on setting the project as a hardcoded string, an env variable, getting it from the metadata server, or from the auth library.

functions/billing/stop-billing/index.js
glasnt: I think you can change your test, with comment documentation, to check for the dryrun message rather than the live message. You don't want users to see the output of "billing disabled", even with a dryrun message preceeding it. 

I'd recommend something like: 

```javascript
  // TODO(developer): As stopping billing is a destructive action
  // for your project, this code defaults to not performing the change, unless 
  // the following variable has been manually changed. 
  const dryRun = true;

...

  const _disableBillingForProject = async (projectName, dryRun) => {
  if (dryRun) {
    console.log('** This script would disable billing here, but "dryRun" has been set to True. Change "dryRun" to alter this behaviour');
    return;
  } else { // ⚠️  Important to add here

  // rest of code here
};
```
Then update the test to test for the dryrun message, with a comment saying something like "We would never test that this function actually runs, as it breaks CI. So instead, test that the dryrun worked"

functions/billing/stop-billing/index.js
glasnt: Yup, I needed to confirm this myself that the functionality has changed since the introduction of Cloud Run functions. A lot of python examples just use the equivelant metadata call rather than also use a envvar, but I get for _this_ example, you may want the script to be able to run cross-project.

eapl-gemugami: It seems it changed for 'newer' Cloud Run runtimes (Python 3.8+ and I think Node 20+): https://stackoverflow.com/a/71013818

I didn't find sound evidence in the GCP documentation, but it could help you.

glasnt (commented):
I think this PR is now okay, but I would appreciate a second reviewer/approver to confirm

functions/billing/stop-billing/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Disable billing for a project by removing its billing account
+ * @param {string} projectName The name of the project to disable billing
+ * @returns {void}
+ */
+const _disableBillingForProject = async projectName => {
+  if (dryRun) {
+    console.log(
+      '** This script would disable billing here, but "dryRun" has been set to true.' +
```
iennae: nit: ** INFO: Script running in info-only mode because "dryRun" is true. To disable billing, set "dryRun" to false.

PR #4182, opened 2025-09-12
fix(container-analysis): Update createNote.js to replace vulnerability with attestation

lvoelz:
Update createNote.js to replace references to storing vulnerabilities with storing attestations in the Artifact Analysis documentation, including code samples. The updated code sample will appear in https://cloud.google.com/artifact-analysis/docs/create-notes-occurrences and any other topics that reference it in the google cloud documentation. Additional pull requests will be sent to update the equivalent code samples for Go, Java, Python, and Ruby.

### DIFF
--- a/container-analysis/snippets/createNote.js
+++ b/container-analysis/snippets/createNote.js
@@ -46,7 +46,7 @@ async function main(
     note: {
       attestation: {
         hint: {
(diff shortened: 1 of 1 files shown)


briandorsey:
Updated the PR title to make the "conventionalcommits" check happy.

PR #4337, opened 2026-06-09
Revert "refactor(genai): migrate region tags from generative-ai to genai folder (Batch 2)"

angelcaamal:
Reverts GoogleCloudPlatform/nodejs-docs-samples#4336


iennae (approved):
I think you should be able to leave the region tag added to the files that you've verified it's the equivalent sample.

